PR #1, opened 2020-09-28
Swap Monero for Bitcoin

luckysori:


monero-harness/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
bonomat, line 1: you can move this one level up into the root of the repository

bonomat (approved):
Too big to review 😬

luckysori:
> Too big to review grimacing

It sure is :P As we keep iterating on it I think we will all notice the things that may be wrong or subpar. Going ahead with the merge as discussed!

PR #2, opened 2020-10-06
Execute Alice and Bob state machines concurrently

rishflab:
Previously we were testing the protocol by manually driving Alice and
Bob's state machines. This logic has now be moved to an async state
transition function that can take any possible state as input. The
state transition function is called in a loop until it returns the
desired state. This allows use to interrupt midway through the protocol
and perform refund and punish tests. This design was chosen over a
generator based implementation because the the generator based
implementation results in a impure state transition function that is
difficult to reason about and prone to bugs.

Test related code was extracted into the tests folder.

The 2b and 4b states were renamed to be consistent with the rest.

Macros were used to reduce code duplication when converting
child states to their parent states and vice versa.

Todos were added were neccessary.

**Sorry, this might be bit hard to review as I did not have a clean commit history and had to squash. I would focus on xmr-btc/bob.rs, tests/e2e.rs, tests/node.rs, tests/transport.rs and the `next_state(..)` functions in xmr-btc/alice.rs**


xmr-btc/src/alice.rs
```diff
@@ -2,33 +2,178 @@ use anyhow::{anyhow, Result};
...
+    transport::SendReceive,
+};
+use ecdsa_fun::nonce::Deterministic;
 use sha2::Sha256;
+use std::convert::{TryFrom, TryInto};
+
+pub mod message;
+pub use message::{Message, Message0, Message1, Message2, UnexpectedMessage};
+
+pub async fn next_state<
```
luckysori, line 19: :question: How do you see these `next_state` functions fitting in with a function that orchestrates the swap for Alice or Bob?

I see the benefit of this design as allowing you to control how far the state machine goes, which is useful for testing, but I think the end user of this library will want the swap to be executed to completion and for all the abort scenarios to be handled for him.

It would be nice if we could find a way to keep the flexibility for testing and provide such a high-level API at the same time.

luckysori (commented):
Nice work!

Apart from some minor comments I documented some of my thoughts on what I think the API of the library should be.

xmr-btc/Cargo.toml
```diff
@@ -17,6 +17,8 @@ monero = "0.9"
 rand = "0.7"
 sha2 = "0.9"
 thiserror = "1"
+tokio = { version = "0.2", default-features = false, features = ["time"] }
```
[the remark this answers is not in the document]
rishflab: Yes, this is because we have a tokio::delay in one of the blockchain calls.

xmr-btc/src/alice.rs
```diff
@@ -2,33 +2,178 @@ use anyhow::{anyhow, Result};
...
+            refund_timelock,
+            punish_timelock,
+            redeem_address,
+            punish_address,
+        ))
+    }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
rishflab: In one of the iterations I was cloning states, it is not longer needed. I am not sure whether best practice is to keep it or remove it?

xmr-btc/src/alice.rs
rishflab: In that case you can simple ask to run to the last state which will do what you are asking. This is why I kept the run_until() functions in the test lib. For production we would use a similar function except it wont take in a desired end state will always try and run until the end state.
I am not sure how the abort scenarios will tie in. I think it is probably best to write a test or design a client so we can get a better idea how the API should be.

xmr-btc/Cargo.toml
```diff
@@ -25,3 +27,5 @@ monero-harness = { path = "../monero-harness" }
 reqwest = { version = "0.10", default-features = false }
 testcontainers = "0.10"
 tokio = { version = "0.2", default-features = false, features = ["blocking", "macros", "rt-core", "time", "rt-threaded"] }
+tracing-subscriber = "0.2.12"
```
D4nte, line 30:
```suggestion
tracing-subscriber = "0.2"
```

xmr-btc/src/alice.rs
```diff
@@ -2,33 +2,178 @@ use anyhow::{anyhow, Result};
 use ecdsa_fun::adaptor::{Adaptor, EncryptedSignature};
 use rand::{CryptoRng, RngCore};
```
D4nte, line 4: Remove newline between `use` statements

xmr-btc/src/alice.rs
```diff
@@ -2,33 +2,178 @@ use anyhow::{anyhow, Result};
...
+    R: RngCore + CryptoRng,
+    B: GetRawTransaction + BroadcastSignedTransaction,
+    M: ImportOutput + Transfer,
+    T: SendReceive<Message, bob::Message>,
+>(
+    bitcoin_wallet: &B,
+    monero_wallet: &M,
+    transport: &mut T,
+    state: State,
+    rng: &mut R,
```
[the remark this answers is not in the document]
D4nte: > 🙃 We normally put dependencies at the top of argument lists, as opposed to data which goes at the bottom.

Interesting, where did you get that from?

xmr-btc/src/alice.rs
```diff
@@ -2,33 +2,178 @@ use anyhow::{anyhow, Result};
...
+    State0(State0),
+    State1(State1),
+    State2(State2),
+    State3(State3),
+    State4(State4),
+    State5(State5),
+    State6(State6),
+}
+
+// TODO: use macro or generics
```
D4nte, line 91: use a macro? :)

xmr-btc/src/alice/message.rs
```diff
@@ -0,0 +1,114 @@
+use anyhow::Result;
+use ecdsa_fun::adaptor::EncryptedSignature;
+
+use crate::{bitcoin, monero};
```
D4nte, line 4:
```suggestion
use ecdsa_fun::adaptor::EncryptedSignature;
use crate::{bitcoin, monero};
```

xmr-btc/Cargo.toml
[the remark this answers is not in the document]
tcharding: It is not obvious from the code _why_ we have those sleeps in there, can they be removed and the dependency on tokio removed?

xmr-btc/src/alice.rs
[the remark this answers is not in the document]
tcharding: I had not consciously inferred that either but would have put `rng` first. Is that based on how we used the old `Facade` in cnd?

xmr-btc/src/alice.rs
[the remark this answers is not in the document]
tcharding: Best practice is, AFAIK: if a derive is not required then do not use it.

xmr-btc/src/alice.rs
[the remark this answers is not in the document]
luckysori: My impression was that it was an implicit CoBloX convention. I blame @thomaseizinger.

xmr-btc/src/alice.rs
luckysori: > I think it is probably best to write a test or design a client so we can get a better idea how the API should be.

That's a fair point.

I guess for the time being it is sufficient if we offer the building blocks to handle all the abort scenarios, and if we develop an application that uses this library we can continue to iterate on it.

xmr-btc/src/alice.rs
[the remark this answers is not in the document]
thomaseizinger: I'll take the blame but I don't think it is a convention per se :D

xmr-btc/src/alice.rs
rishflab: the is_state functions can easily be implemented using a macro. I have delayed writing theser macros for now as these functions could be replaced I use generics in the state machine driver function

xmr-btc/tests/e2e.rs
```diff
@@ -152,12 +152,12 @@ mod tests {
         node::{run_alice_until, run_bob_until},
     };
 
-    use futures::future;
     use monero_harness::Monero;
     use rand::rngs::OsRng;
-    use std::convert::TryInto;
     use testcontainers::clients::Cli;
```
tcharding, line 158: Please do not leave whitespace between use statements unless it implies some meaning (such as separating current repository libraries from other libraries).

xmr-btc/tests/node.rs
```diff
@@ -0,0 +1,92 @@
...
+    ) -> AliceNode<'a> {
+        Self {
+            transport,
+            bitcoin_wallet,
+            monero_wallet,
+        }
+    }
+}
+
+pub async fn run_alice_until<'a, R: RngCore + CryptoRng>(
```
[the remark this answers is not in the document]
rishflab: done

xmr-btc/Cargo.toml
```diff
@@ -25,3 +27,5 @@ monero-harness = { path = "../monero-harness" }
...
+tracing = "0.1"
```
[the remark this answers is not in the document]
rishflab: Needs both

xmr-btc/src/alice.rs
```diff
@@ -1,31 +1,167 @@
...
-    pub(crate) v_a: monero::PrivateViewKey,
-    pub(crate) redeem_address: bitcoin::Address,
-    pub(crate) punish_address: bitcoin::Address,
+            let bob_message0: bob::Message0 = transport.receive_message().await?.try_into()?;
+            let state1 = state0.receive(bob_message0)?;
+            Ok(state1.into())
+        }
+        State::State1(state1) => {
+            let bob_message1: bob::Message1 = transport.receive_message().await?.try_into()?;
+            let state2 = state1.receive(bob_message1);
```
thomaseizinger, line 44: I believe `state1.receive` only expects one specific type: `bob::Message1` right? Shouldn't need the type annotation then :)

xmr-btc/src/alice.rs
```diff
@@ -250,12 +386,15 @@ pub struct State3 {
 impl State3 {
     pub async fn watch_for_lock_btc<W>(self, bitcoin_wallet: &W) -> Result<State4>
     where
-        W: bitcoin::GetRawTransaction,
+        W: bitcoin::WatchForRawTransaction,
     {
-        let _ = bitcoin_wallet
-            .get_raw_transaction(self.tx_lock.txid())
+        tracing::info!("{}", self.tx_lock.txid());
```
thomaseizinger, line 391: Might want to include some text in your log statements.

xmr-btc/src/alice/message.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+impl TryFrom<Message> for Message0 {
+    type Error = UnexpectedMessage;
+
+    fn try_from(m: Message) -> Result<Self, Self::Error> {
+        match m {
+            Message::Message0(m) => Ok(m),
+            _ => Err(UnexpectedMessage {
+                expected_type: "Create0".to_string(),
```
thomaseizinger, line 49: Stale string? Should probably say `Message0`. Could implement this as a macro and use `stringify!()` to get the name of the struct you are implementing it on.

xmr-btc/src/bob/message.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl TryFrom<Message> for Message0 {
+    type Error = UnexpectedMessage;
+
+    fn try_from(m: Message) -> Result<Self, Self::Error> {
+        match m {
+            Message::Message0(m) => Ok(m),
+            _ => Err(UnexpectedMessage {
+                expected_type: "Create0".to_string(),
```
thomaseizinger, line 53: Also looks stale.

xmr-btc/tests/e2e.rs
```diff
@@ -0,0 +1,386 @@
...
+        .await
+        .unwrap();
+    let bob_btc_wallet = wallet::bitcoin::make_wallet("bob", &bitcoind, btc_amount)
+        .await
+        .unwrap();
+
+    let (alice_transport, bob_transport) = init_alice_and_bob_transports();
+
+    let alice = AliceNode::new(alice_transport, alice_btc_wallet, alice_monero_wallet);
+
```
thomaseizinger, line 99: Could remove the newline for consistency.

```suggestion
```

xmr-btc/tests/e2e.rs
```diff
@@ -0,0 +1,386 @@
...
+        alice_state0,
+        bob_state0,
+        alice,
+        bob,
+        initial_balances,
+        swap_amounts,
+    )
+}
+
+#[cfg(test)]
```
thomaseizinger, line 145: That is redundant within the `tests/` directory. All of this code gets only compiled if the `test` cfg is set anyway.

xmr-btc/src/bob.rs
```diff
@@ -11,31 +16,122 @@ use ecdsa_fun::{
 };
 use rand::{CryptoRng, RngCore};
 use sha2::Sha256;
+use std::convert::{TryFrom, TryInto};
+
+pub mod message;
+pub use message::{Message, Message0, Message1, Message2, Message3, UnexpectedMessage};
+
+pub async fn next_state<
```
thomaseizinger, line 24: This could be a function `next` on the actual `State` itself. That would make it consistent with how the individual states are implemented.

xmr-btc/tests/e2e.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        bob_node
+            .monero_wallet
+            .0
+            .wait_for_bob_wallet_block_height()
+            .await
+            .unwrap();
+
+        let bob_final_xmr_balance = bob_node.monero_wallet.0.get_balance_bob().await.unwrap();
+
```
thomaseizinger, line 233: If possible, I would always group the asserts together in the end. That makes it easier to follow that the test is now over once you reach the assertions.

```suggestion
        let alice_final_xmr_balance = alice_node
            .monero_wallet
            .0
            .get_balance_alice()
            .await
            .unwrap();

        bob_node
            .monero_wallet
            .0
            .wait_for_bob_wallet_block_height()
            .await
            .unwrap();

        let bob_final_xmr_balance = bob_node.monero_wallet.0.get_balance_bob().await.unwrap();

        assert_eq!(
            alice_final_btc_balance,
            initial_balances.alice_btc + swap_amounts.btc
                - bitcoin::Amount::from_sat(bitcoin::TX_FEE)
        );
        assert_eq!(
            bob_final_btc_balance,
            initial_balances.bob_btc - swap_amounts.btc - lock_tx_bitcoin_fee
        );
```

xmr-btc/tests/e2e.rs
```diff
@@ -0,0 +1,386 @@
...
+            .transaction_fee(bob_state3.tx_lock_id())
+            .await
+            .unwrap();
+
+        assert_eq!(alice_final_btc_balance, initial_balances.alice_btc);
+        assert_eq!(
+            bob_final_btc_balance,
+            // The 2 * TX_FEE corresponds to tx_refund and tx_cancel.
+            initial_balances.bob_btc - Amount::from_sat(2 * TX_FEE) - lock_tx_bitcoin_fee
+        );
```
thomaseizinger, line 305: Same here, would be good to group the asserts together.

xmr-btc/src/alice.rs
```diff
@@ -1,31 +1,167 @@
...
+    State4(State4),
+    State5(State5),
+    State6(State6),
 }
 
-#[derive(Debug)]
-pub struct Message2 {
-    pub(crate) tx_lock_proof: monero::TransferProof,
+// TODO: use macro or generics
+pub fn is_state4(state: &State) -> bool {
```
thomaseizinger, line 92: As far as I could see, these are only used for the test right? Could move it to the harness test harness then and keep the library slim.

xmr-btc/tests/harness/transport.rs
```diff
@@ -0,0 +1,64 @@
...
+impl Send<alice::Message> for Transport<alice::Message, bob::Message> {
+    async fn send_message(&mut self, message: alice::Message) -> Result<()> {
+        let _ = self
+            .sender
+            .send(message)
+            .await
+            .map_err(|_| anyhow!("failed to send message"))?;
+        Ok(())
+    }
+}
```
thomaseizinger, line 28: Should be able to implement this generically no?

```suggestion
#[async_trait]
impl<A, B> Send<A> for Transport<A, B> {
    async fn send_message(&mut self, message: A) -> Result<()> {
        let _ = self
            .sender
            .send(message)
            .await
            .map_err(|_| anyhow!("failed to send message"))?;
        Ok(())
    }
}
```

thomaseizinger: That would save you two of these impl blocks.

thomaseizinger (approved):
Couple of nits but looks good in general :)

xmr-btc/src/bob.rs
rishflab: leaving this for now, I think there is going to be fair bit of churn on this state machine

bonomat: That's a good idea. Let's do this in the next iteration.

xmr-btc/tests/harness/transport.rs
rishflab: Good pick!

xmr-btc/Cargo.toml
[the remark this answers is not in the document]
luckysori: Hmmm, I just tested locally and it works without the one in `dev-dependencies`.

xmr-btc/src/alice/message.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+impl TryFrom<Message> for Message2 {
+    type Error = UnexpectedMessage;
+
+    fn try_from(m: Message) -> Result<Self, Self::Error> {
+        match m {
+            Message::Message2(m) => Ok(m),
+            _ => Err(UnexpectedMessage {
```
luckysori, line 88: :wrench: The `UnexpectedMessage` struct was just an idea that I introduced in `thor`. If you want to use it here I recommend using the constructor and take the expected type of the message as a type parameter.

It may even be better to move the type parameter from the method to the struct, so that the associated type `Error` of `TryFrom` already specifies what the expected message is.

rishflab: Implemented this conversion with a macro now so no longer need this struct

xmr-btc/src/bob.rs
```diff
@@ -279,8 +341,9 @@ pub struct State2b {
     tx_refund_encsig: EncryptedSignature,
 }
 
-impl State2b {
-    pub async fn watch_for_lock_xmr<W>(self, xmr_wallet: &W, msg: alice::Message2) -> Result<State3>
+impl State3 {
+    // todo: loop until punish? timelock has expired
```
luckysori, line 345: :wrench: It would be the refund/cancel timelock in this case. But in any case, I would remove this TODO as we don't yet know in which layer we want to do this.

xmr-btc/tests/harness/mod.rs
```diff
@@ -0,0 +1,36 @@
+pub mod node;
+pub mod transport;
+pub mod wallet;
+
+pub mod bob {
+    use xmr_btc::bob::State;
+
+    // TODO: use macro or generics
+    pub fn is_state5(state: &State) -> bool {
+        matches!(state, State::State5 { .. })
```
luckysori, line 10: :upside_down_face: Given that it's test code we could just inline these and reduce lines of code.

luckysori (approved):
LGTM!

xmr-btc/src/bob.rs
rishflab: correct!

xmr-btc/tests/harness/mod.rs
rishflab: i tried inlining the matches! macro but the run_until function need a function as an argument which takes in the state which the execution has progressed to.

PR #6, opened 2020-10-08
Add messaging layer

tcharding:
Adds a messaging layer similar to what we did in `thor`. Please note this PR duplicates the current happy path test into a new module then adds code to that module. 

With this applied the happy path function is even more horrendous than it was. Have not worked out how to make it easier to read yet. 

If this merges and the approach (as well as the cleanup) is acceptable to the team then we can remove the duplication. This is an attempt to not put others in merge hell while still having the freedom to do useful work as I see it. Open to suggestions.

### DIFF
--- a/xmr-btc/src/happy_path.rs
+++ b/xmr-btc/src/happy_path.rs
@@ -1,0 +1,149 @@
+//! This module shows how a BTC/XMR atomic swap proceeds along the happy path.
+
+use crate::{alice, bitcoin, bob, monero};
+use bitcoin_harness::Bitcoind;
+use monero_harness::Monero;
+use rand::rngs::OsRng;
+use testcontainers::clients::Cli;
+
+const TEN_XMR: u64 = 10_000_000_000_000;
+
+pub async fn init_bitcoind(tc_client: &Cli) -> Bitcoind<'_> {
+    let bitcoind = Bitcoind::new(tc_client, "0.19.1").expect("failed to create bitcoind");
+    let _ = bitcoind.init(5).await;
+
+    bitcoind
+}
+
+#[tokio::test]
+async fn happy_path() {
+    let cli = Cli::default();
+    let monero = Monero::new(&cli);
+    let bitcoind = init_bitcoind(&cli).await;
+
+    // must be bigger than our hardcoded fee of 10_000
+    let btc_amount = bitcoin::Amount::from_sat(10_000_000);
+    let xmr_amount = monero::Amount::from_piconero(1_000_000_000_000);
+
+    let fund_alice = TEN_XMR;
+    let fund_bob = 0;
+    monero.init(fund_alice, fund_bob).await.unwrap();
+
+    let alice_monero_wallet = monero::AliceWallet(&monero);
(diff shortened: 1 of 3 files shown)


xmr-btc/src/happy_path.rs
```diff
@@ -0,0 +1,216 @@
...
+pub async fn init_bitcoind(tc_client: &Cli) -> Bitcoind<'_> {
+    let bitcoind = Bitcoind::new(tc_client, "0.19.1").expect("failed to create bitcoind");
+    let _ = bitcoind.init(5).await;
+
+    bitcoind
+}
+
+/// Create two mock `Transport`s which mimic a peer to peer connection between
+/// two parties, allowing them to send and receive `thor::Message`s.
+pub fn make_transports() -> (Transport, Transport) {
```
bonomat, line 30: This looks similar to what @rishflab did in his PR in https://github.com/comit-network/xmr-btc-swap/pull/2 
see https://github.com/comit-network/xmr-btc-swap/blob/a5e9d12a6c6c489eccd6de6dd4ae85da38f2e60f/xmr-btc/tests/e2e.rs#L41-L60

xmr-btc/src/happy_path.rs
```diff
@@ -0,0 +1,216 @@
...
+    async fn receive_message(&mut self) -> Result<Message> {
+        let msg = self
+            .receiver
+            .next()
+            .await
+            .ok_or_else(|| anyhow!("failed to receive message"))?;
+
+        Ok(msg)
+    }
+}
```
bonomat, line 74: This seems also to be very similar to what @rishflab did in his PR #2 in https://github.com/comit-network/xmr-btc-swap/blob/b6df715b1bdef81f20333145e1705cbdcf0e4fb8/xmr-btc/tests/harness/transport.rs#L16-L55

The difference is that @rishflab made a difference in the message types and uses two different enums, one for Alice's messages and one for Bob's messages. This should prevent anyone from sending the wrong message.

xmr-btc/src/happy_path.rs
tcharding: Yes except he puts two methods on a single trait - a single method per trait is cleaner. Also his traits are never used, writing these traits takes about 2 minutes of cut and pasting from `thor`. I do not see this as an issue.

bonomat: True, he uses a single trait. That can be refactored easily. 

It looks like the traits are used in the `next_state` function for Alice and Bob 

https://github.com/comit-network/xmr-btc-swap/blob/b6df715b1bdef81f20333145e1705cbdcf0e4fb8/xmr-btc/src/alice.rs#L34-L35

and 

https://github.com/comit-network/xmr-btc-swap/blob/b6df715b1bdef81f20333145e1705cbdcf0e4fb8/xmr-btc/src/bob.rs#L37-L41

tcharding: > Tthe traits are used in the next_state function for Alice and Bob

The call sites you reference are using the function parameter that is bound by the trait. This code is not using the implementation of the trait that we are discussing.

PR #12, opened 2020-10-13
Generate actions for Bob's on-chain protocol

luckysori:
Mimics what @thomaseizinger did [here](https://github.com/comit-network/comit-rs/blob/move-nectar-swap-to-comit/nectar/src/swap/comit/herc20_hbit.rs#L57-L184) and [here](https://github.com/comit-network/comit-rs/blob/e584d2b14f3602e2657d09b989e8ea1d483a0626/nectar/src/swap.rs#L716-L751).

This has the advantage that the consumer has more freedom to execute`Action`s without having to implement particular traits. The errorhandling required inside this protocol-executing function is also reduced.

As discussed with Thomas, for this approach to work well, the trait functions such as `receive_transfer_proof` should be infallible,
and the implementer should be forced to hide IO errors behind a retry mechanism.

All of these asynchronous calls need to be "raced" against the abort condition (determined by the `refund_timelock`), which is
missing in the current state of the implementation.

The initial handshake of the protocol has not been included here, because it may not be easy to integrate this approach with libp2p, but a couple of messages still need to exchanged. I need @tcharding to tell me if it's feasible/good to do it like this.

---

I've tried to isolate most changes to the `lib.rs` file for the moment, but we may or may not want to use/modify the `bob` state structs and call functions on them.


xmr-btc/tests/on_chain.rs
```diff
@@ -0,0 +1,67 @@
...
+    bitcoin::{BroadcastSignedTransaction, SignTxLock},
+    bob,
+    monero::CreateWalletForOutput,
+    Action, ReceiveTransferProof,
+};
+
+struct Network;
+
+impl ReceiveTransferProof for Network {
+    fn receive_transfer_proof(&self) -> xmr_btc::monero::TransferProof {
```
thomaseizinger, line 17: This will need to be `async`.

You can probably implement this by having a oneshot where you return the receiver here, keep the sender around and when you get the event from libp2p, pass it in into the sender to continue the protocol.

xmr-btc/src/lib.rs
```diff
@@ -50,3 +50,151 @@ pub mod bitcoin;
...
+            // transaction
+            monero_ledger
+                .check_transfer(S, v.public(), transfer_proof, xmr)
+                .await
+                .expect("TODO: implementor of this trait must make it infallible by retrying");
+
+            let tx_redeem = bitcoin::TxRedeem::new(&tx_lock, &redeem_address);
+            let tx_redeem_encsig = b.encsign(S_a_bitcoin.clone(), tx_redeem.digest());
+
+            co.yield_(Action::SendBitcoinRedeemEncsig(tx_redeem_encsig.clone()))
```
thomaseizinger, line 131: If this is network communication, you might want to yield the message directly, just so that the "driver" on the outside is a stupid as possible and more logic is contained within the protocol execution.

xmr-btc/src/lib.rs
```diff
@@ -50,3 +50,151 @@ pub mod bitcoin;
...
+/// Perform the on-chain protocol to swap monero and bitcoin as Bob.
+///
+/// This is called post handshake, after all the keys, addresses and most of the
+/// signatures have been exchanged.
+pub fn action_generator_bob<N, M, B>(
+    network: &'static N,
+    monero_ledger: &'static M,
+    bitcoin_ledger: &'static B,
+    // TODO: Replace this with a new, slimmer struct?
+    bob::State2 {
```
thomaseizinger, line 87: Why only start at `State2`? Could pack the whole protocol into here :)

xmr-btc/src/lib.rs
```diff
@@ -50,3 +50,151 @@ pub mod bitcoin;
...
+                s_b.into_ed25519(),
+            ));
+            let S = S_a_monero + S_b_monero;
+
+            // TODO: We should require a specific number of confirmations on the lock
+            // transaction
+            monero_ledger
+                .check_transfer(S, v.public(), transfer_proof, xmr)
+                .await
+                .expect("TODO: implementor of this trait must make it infallible by retrying");
```
thomaseizinger, line 126: Yes, plus as you noted in the PR description, it should be gated by a timeout value that enforces the atomicity of the protocol.

xmr-btc/src/lib.rs
```diff
@@ -50,3 +50,151 @@ pub mod bitcoin;
...
+                let sig_a =
+                    adaptor.decrypt_signature(&s_b.into_secp256k1(), tx_refund_encsig.clone());
+                let sig_b = b.sign(tx_refund.digest());
+
+                tx_refund
+                    .add_signatures(&tx_cancel, (A.clone(), sig_a), (b.public(), sig_b))
+                    .expect("sig_{a,b} to be valid signatures for tx_refund")
+            };
+
+            co.yield_(Action::RefundBitcoin {
```
thomaseizinger, line 193: Ideally, our swap execution state machines need to get even more fine granular than that where after yielding such an action, we watch for it to appear in the mempool, monitor if the fees we set are appropriate and if not, emit another action that bumps the fees until the TX gets included in time.

thomaseizinger (commented):
Well executed @luckysori !

I've added some comments inline that highlight some of the key design decision but overall I think this is something you should be able to work with!

xmr-btc/tests/on_chain.rs
thomaseizinger: You may want this to be `&mut self` for that.

xmr-btc/src/lib.rs
luckysori: > you might want to yield the message directly

Do you mean the message type, as opposed to its contents (in this case an encrypted signature)? If so, I agree.

xmr-btc/src/lib.rs
luckysori: It's certainly an option, but @tcharding is working on the libp2p protocol at the moment. I thought that it would be easier (at least for now), to separate the initial handshake (during which signatures and keys are exchanged) from the on-chain protocol, so that we can work in parallel.

xmr-btc/src/lib.rs
luckysori: That is a cool idea with the fee bumping! I had a similar thought when I considered splitting this action into two, and only yield the second one if the first has sufficient confirmations. This means that the caller doesn't need to consider all this logic.

xmr-btc/src/lib.rs
thomaseizinger: Yes the actual message type, just so that this component contains all the logic of what should happen when (which is the essence of a protocol) and the outside component can be dumb and only really does IO without having to convert between types.

xmr-btc/src/lib.rs
thomaseizinger: I guess it depends on how you design the libp2p protocol. Are they always the same messages, unconditionally? If so then it could probably be encapsulated into a single protocol at the `ProtocolsHandler` level and only emit the completed handshake to the `NetworkBehaviour` and sub-sequentially to the rest of the application.

If that is the case, then it would probably also split the protocol logically into these parts so that what you are doing here requires the handshake to be done. Currently it feels like we are starting "in the middle". Maybe that is fine though :)

xmr-btc/src/lib.rs
luckysori: This can also fail if the transaction doesn't lock up enough money. I think that should also result in a refund.

xmr-btc/tests/harness/wallet/monero.rs
```diff
@@ -66,33 +67,56 @@ impl CreateWalletForOutput for AliceWallet<'_> {
...
+                }));
+            }
+
+            if proof.confirmations < expected_confirmations {
+                return Err(backoff::Error::Transient(Error::InsufficientConfirmations));
+            }
+
+            Ok(proof)
+        })
+        .retry(ExponentialBackoff::default())
```
thomaseizinger: Nice! I am excited for how this will turn out!

monero-harness/src/lib.rs
```diff
@@ -48,18 +48,17 @@ const WAIT_WALLET_SYNC_MILLIS: u64 = 1000;
...
+    bob_wallet_rpc_port: u16,
 }
 
-impl<'c> Monero<'c> {
+impl<'c> Monero {
     /// Starts a new regtest monero container.
-    pub fn new(cli: &'c Cli) -> Self {
+    pub fn new(cli: &'c Cli) -> (Self, Container<'c, Cli, image::Monero>) {
         let mut rng = rand::thread_rng();
         let monerod_rpc_port: u16 = rng.gen_range(1024, u16::MAX);
```
thomaseizinger, line 63: What is the reason why you choose these ports yourself?

If you wouldn't specify them, testcontainers (actually `docker`) would automatically find a free one for you that you can query with `Container#get_host_port`. It seems counter-intuitive that you have to pass the `Client` to this function when actually, the `Image` is eventually passed to the `Client` within `run`.

thomaseizinger: Not to mention that the random generation can potentially fail as it might hit one that is already taken :)

luckysori: @tcharding and @bonomat know more than me about this code :D

bonomat: The docker image we are using does not automatically expose those ports. That's why we need to explicitly map them.

+1 on the ports being in used problem.

thomaseizinger: > The docker image we are using does not automatically expose those ports. That's why we need to explicitly map them.

That is a shame :/

thomaseizinger: Let's see what they think: https://github.com/XMRto/monero/issues/9

bonomat: I should have told the whole story: we are slightly misusing their container and are starting 3 wallets instead of only 1. That is why the ports are not exposed by the image.

thomaseizinger: Why don't you just start three containers for that? Sounds like scaling on the wrong level that  creates more work like having to manually assign ports instead of being able to roll with the default behaviour (assuming the ports are exposed by default).

bonomat: The setup is like this: 
* 1 monerod node
* 3 monero-wallet-rpc services

All three monero-wallet-rpc services connect to the same monerod node.

thomaseizinger: The most recent version of testcontainers [supports running containers on a network](https://github.com/testcontainers/testcontainers-rs/blob/dev/CHANGELOG.md#added), so you could actually start those in separate containers and they could still be connected to each other.

bonomat: Cool, we would have needed that a few weeks ago :)

thomaseizinger: Why didn't you raise it on the repo then?

xmr-btc/tests/harness/wallet/monero.rs
luckysori: Fun fact: don't use _exponential_ backoff in tests.

PR #13, opened 2020-10-20
Tor

bonomat:
This PR adds a small lib which allows us to add a Tor service easily. The whole thing is hidden behind a feature-flag so that it does not bother us. 

Note: GH action might change if it fails, the rest is ready for review.

Kudos to @luckysori for helping with the annoying handler type.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,19 @@ jobs:
   build_test:
     runs-on: ubuntu-latest
     steps:
-      - name: Install tor
-        run: sudo apt-get tor
 
       - name: Checkout sources
         uses: actions/checkout@v2
+
+      - name: Install and stop tor in case it was running
+        run: |
+          sudo apt install software-properties-common
+          sudo curl https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | sudo gpg --import
+          sudo gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | sudo apt-key add -
+          sudo add-apt-repository 'deb https://deb.torproject.org/torproject.org bionic main'
+          sudo apt update
+          sudo apt install tor deb.torproject.org-keyring
+          sudo /etc/init.d/tor stop
 
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
   build_test:
     runs-on: ubuntu-latest
     steps:
+      - name: Install tor
+        run: sudo apt-get tor
+
       - name: Checkout sources
         uses: actions/checkout@v2
 
--- a/xmr-btc/src/tor.rs
+++ b/xmr-btc/src/tor.rs
@@ -10,53 +10,78 @@ use torut::{
     onion::TorSecretKeyV3,
 };
 
-lazy_static! {
-    /// The default TOR socks5 proxy address, `127.0.0.1:9050`.
-    pub static ref TOR_PROXY_ADDR: SocketAddrV4 = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9050);
-    /// The default TOR Controller Protocol address, `127.0.0.1:9051`.
-    pub static ref TOR_CP_ADDR: SocketAddr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9051));
-}
-
-/// checks if tor is running
-async fn tor_running() -> Result<()> {
-    // Make sure you are running tor and this is your socks port
-    let proxy = reqwest::Proxy::all(format!("socks5h://{}", *TOR_PROXY_ADDR).as_str())
-        .expect("tor proxy should be there");
-    let client = reqwest::Client::builder()
-        .proxy(proxy)
-        .build()
-        .expect("should be able to build reqwest client");
-
-    let res = client.get("https://check.torproject.org").send().await?;
-
-    let text = res.text().await?;
-    let is_tor = text.contains("Congratulations. This browser is configured to use Tor.");
-
-    if is_tor {
-        Ok(())
-    } else {
-        bail!("Tor is currently not running")
-    }
-}
+// lazy_static! {
+//     The default TOR socks5 proxy address, `127.0.0.1:9050`.
+// pub static ref TOR_PROXY_ADDR: SocketAddrV4 =
+// SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9050); The default TOR Controller
+// Protocol address, `127.0.0.1:9051`. pub static ref TOR_CP_ADDR: SocketAddr =
+// SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9051)); }
 
 type Handler = fn(AsyncEvent<'_>) -> Box<dyn Future<Output = Result<(), ConnError>> + Unpin>;
 
 #[allow(missing_debug_implementations)]
-pub struct AuthenticatedConnection(AuthenticatedConn<TcpStream, Handler>);
+pub struct AuthenticatedConnection {
+    tor_proxy_address: SocketAddrV4,
+    tor_control_port_address: SocketAddr,
+    authenticated_connection: Option<AuthenticatedConn<TcpStream, Handler>>,
+}
+
+impl Default for AuthenticatedConnection {
+    fn default() -> Self {
+        Self {
+            tor_proxy_address: SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9050),
+            tor_control_port_address: SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9051)),
+            authenticated_connection: None,
+        }
+    }
+}
 
 impl AuthenticatedConnection {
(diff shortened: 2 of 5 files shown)


xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,104 @@
...
+    control::{AsyncEvent, AuthenticatedConn, ConnError, UnauthenticatedConn},
+    onion::TorSecretKeyV3,
+};
+
+lazy_static! {
+    /// The default TOR socks5 proxy address, `127.0.0.1:9050`.
+    pub static ref TOR_PROXY_ADDR: SocketAddrV4 = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9050);
+    /// The default TOR Controller Protocol address, `127.0.0.1:9051`.
+    pub static ref TOR_CP_ADDR: SocketAddr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 9051));
+}
```
luckysori, line 18: :wrench: These are only ever used as `&str`. Could we initialise them as such using `const`s without needing `lazy_static`? Or is this providing something more?

xmr-btc/tests/tor.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    #[tokio::test]
+    async fn test_tor_control_port() -> Result<()> {
+        // Setup test HTTP Server
+        let (tx, rx) = tokio::sync::oneshot::channel::<()>();
+        let port = 8080;
+        start_test_service(port, rx);
+
+        // Connect to local Tor service
+        let mut authenticated_connection = AuthenticatedConnection::new().await?;
```
luckysori, line 43: :thinking: I guess this means that we assume that Tor is running before this test is run. We should skip this in CI and document it if that's the case.

luckysori (commented):
Just a couple of things I noticed.

xmr-btc/src/tor.rs
bonomat: Removed in the last patch.

xmr-btc/tests/tor.rs
bonomat: I'm starting Tor from the test now. For that it assumes that Tor is in the path. 
Not sure if this is the nicest way to go. Maybe mentioning it in the readme is enough? 
It's behind a feature flag anyway.

bonomat: I'm starting Tor from the test now. For that it assumes that Tor is in the path. 
Not sure if this is the nicest way to go. Maybe mentioning it in the readme is enough? 
It's behind a feature flag anyway.

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+                control_port,
+            )),
+        }
+    }
+
+    /// checks if tor is running
+    async fn tor_running(&self) -> Result<()> {
+        // Make sure you are running tor and this is your socks port
+        let proxy = reqwest::Proxy::all(format!("socks5h://{}", self.tor_proxy_address).as_str())
+            .expect("tor proxy should be there");
```
tcharding, line 42: Can we return an error here instead of panicing please.

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+            tor_proxy_address: SocketAddrV4::new(Ipv4Addr::LOCALHOST, proxy_port),
+            tor_control_port_address: SocketAddr::V4(SocketAddrV4::new(
+                Ipv4Addr::LOCALHOST,
+                control_port,
+            )),
+        }
+    }
+
+    /// checks if tor is running
+    async fn tor_running(&self) -> Result<()> {
```
tcharding, line 39: Since this returns an error (i.e., this function is not a predicate), perhaps we could call it `assert_tor_running`

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+        let res = client.get("https://check.torproject.org").send().await?;
+
+        let text = res.text().await?;
+        let is_tor = text.contains("Congratulations. This browser is configured to use Tor.");
+
+        if is_tor {
+            Ok(())
+        } else {
+            bail!("Tor is currently not running")
+        }
```
tcharding, line 57: Perhaps
```
        if !text.contains("Congratulations. This browser is configured to use Tor.") {
            bail!("Tor is currently not running")
        }

        Ok(())
```

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    async fn init_unauthenticated_connection(&self) -> Result<UnauthenticatedConn<TcpStream>> {
+        // Connect to local tor service via control port
+        let sock = TcpStream::connect(self.tor_control_port_address).await?;
+        let unauthenticated_connection = UnauthenticatedConn::new(sock);
+        Ok(unauthenticated_connection)
```
tcharding, line 64: This is a local variable with very limited scope, such a long identifier does not add any additional information. Perhaps we could use
```
let uc = UnauthenticatedConn::new(sock);
Ok(uc)
```

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        let mut unauthenticated_connection = match self.init_unauthenticated_connection().await {
+            Err(_) => bail!("Tor instance not running"),
+            Ok(unauthenticated_connection) => unauthenticated_connection,
+        };
+
+        let tor_info = match unauthenticated_connection.load_protocol_info().await {
+            Ok(info) => info,
+            Err(_) => bail!("Failed to load protocol info from Tor."),
+        };
```
tcharding, line 79: We can use `map_err` here
```
        let tor_info = unauthenticated_connection
            .load_protocol_info()
            .await
            .map_err(|_| anyhow!("Failed to load protocol info from Tor."))?;
```

xmr-btc/src/tor.rs
```diff
@@ -0,0 +1,127 @@
...
+                None,
+                &mut [(
+                    port,
+                    SocketAddr::new(IpAddr::from(Ipv4Addr::new(127, 0, 0, 1)), port),
+                )]
+                .iter(),
+            )
+            .await
+            .map_err(|_| anyhow!("Could not add onion service."))?;
+        Ok(())
```
tcharding, line 125: We can just return this directly, no need to use `?;` and then `Ok(())`

tcharding (approved):
Seems functional. This `torut` library makes it hard to write clean code. ACK'ing as is but please consider two points
1. The types `AuthenticatedConnection` and `UnauthenticatedConnection` are not uniform (not to mention `AuthenticatedConn<TcpStream>`, it would be nice if it was all a bit more balanced. What I mean is unauth just has sockets in it and has conversion methods but auth is totally different - this is unexpected but may not be fixable due to the `torut` library.
2. Local variables are extremely long, please consider using `ac` and `uc` instead of `authenticated_connection`.

xmr-btc/src/tor.rs
bonomat: Good point

xmr-btc/src/tor.rs
bonomat: Looks good, will do that.

xmr-btc/src/tor.rs
bonomat: Is this to be preferred over `bail!` in general?

tcharding: Its functionally exactly the same, just more concise (IMO).

.github/workflows/ci.yml
```diff
@@ -77,6 +83,7 @@ jobs:
         run: cargo check --workspace --all-targets --all-features
 
       - name: Cargo test
-        run: cargo test --workspace --all-features
+        run: cargo test -p xmr-btc test_tor_control_port --all-features
```
tcharding, line 86: Why this change please?

tcharding (approved):
LGTM, except the test change CI question.

.github/workflows/ci.yml
bonomat: To get the tests running :D Obviously this should not have been in this PR.

.github/workflows/ci.yml
```diff
@@ -46,9 +46,20 @@ jobs:
...
         uses: actions/checkout@v2
 
+      - name: Install and stop tor in case it was running
+        run: |
+          sudo apt install software-properties-common
+          sudo curl https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | sudo gpg --import
+          sudo gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | sudo apt-key add -
+          sudo add-apt-repository 'deb https://deb.torproject.org/torproject.org bionic main'
+          sudo apt update
+          sudo apt install tor deb.torproject.org-keyring
```
bonomat, line 60: @tcharding : Is this the way how you installed Tor on your machine btw? 
My issue with CI came from an outdated Tor version

xmr-btc/tests/tor.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn start_test_service(port: u16, rx: Receiver<()>) {
+        let make_svc =
+            make_service_fn(|_conn| async { Ok::<_, Infallible>(service_fn(hello_world)) });
+        let addr = ([127, 0, 0, 1], port).into();
+        let server = hyper::Server::bind(&addr).serve(make_svc);
+        let graceful = server.with_graceful_shutdown(async {
+            rx.await.ok();
+        });
+        tokio::spawn(async {
+            // server.await.unwrap();
```
bonomat, line 33: !

PR #14, opened 2020-10-21
Save and recover protocol state from disk

rishflab:
NOTE: This implementation saves secrets to disk! It is not
secure.

The storage API allows the caller to atomically record the state
of the protocol. The user can retrieve this recorded state and
re-commence the protocol from that point. The state is recorded
using a hard coded key, causing it to overwrite the previously
recorded state. This limitation means that this recovery
mechanism should not be used in a program that simultaneously
manages the execution of multiple swaps.

An e2e test was added to show how to save, recover and resume
protocol execution. This logic could also be integrated into the
run_until functions to automate saving but was not included at
this stage as protocol execution is currently under development.

Serialisation and deserialisation was implemented on the states
to allow the to be stored using the database. Currently the
secret's are also being stored to disk but should be recovered
from a seed or wallets.

### DIFF
--- a/xmr-btc/src/serde.rs
+++ b/xmr-btc/src/serde.rs
@@ -1,48 +1,3 @@
-pub mod ecdsa_fun_signature {
-    use serde::{de, de::Visitor, Deserializer, Serializer};
-    use std::{convert::TryFrom, fmt};
-
-    struct Bytes64Visitor;
-
-    impl<'de> Visitor<'de> for Bytes64Visitor {
-        type Value = ecdsa_fun::Signature;
-
-        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-            write!(formatter, "a string containing 64 bytes")
-        }
-
-        fn visit_bytes<E>(self, s: &[u8]) -> Result<Self::Value, E>
-        where
-            E: de::Error,
-        {
-            if let Ok(value) = <[u8; 64]>::try_from(s) {
-                let sig = ecdsa_fun::Signature::from_bytes(value)
-                    .expect("bytes represent an integer greater than or equal to the curve order");
-                Ok(sig)
-            } else {
-                Err(de::Error::invalid_length(s.len(), &self))
-            }
-        }
-    }
-
-    pub fn serialize<S>(x: &ecdsa_fun::Signature, s: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        s.serialize_bytes(&x.to_bytes())
-    }
-
-    pub fn deserialize<'de, D>(
-        deserializer: D,
-    ) -> Result<ecdsa_fun::Signature, <D as Deserializer<'de>>::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let sig = deserializer.deserialize_bytes(Bytes64Visitor)?;
-        Ok(sig)
-    }
-}
-
 pub mod cross_curve_dleq_scalar {
     use serde::{de, de::Visitor, Deserializer, Serializer};
     use std::{convert::TryFrom, fmt};
@@ -178,7 +133,6 @@ pub mod monero_amount {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use ::bitcoin::SigHash;
     use curve25519_dalek::scalar::Scalar;
     use rand::rngs::OsRng;
     use serde::{Deserialize, Serialize};
@@ -187,9 +141,6 @@ mod tests {
     pub struct CrossCurveDleqScalar(
         #[serde(with = "cross_curve_dleq_scalar")] cross_curve_dleq::Scalar,
     );
-
-    #[derive(Debug, Serialize, Deserialize, PartialEq)]
-    pub struct ECDSAFunSignature(#[serde(with = "ecdsa_fun_signature")] ecdsa_fun::Signature);
 
     #[derive(Debug, Serialize, Deserialize, PartialEq)]
     pub struct MoneroPrivateKey(#[serde(with = "monero_private_key")] crate::monero::PrivateKey);
@@ -203,15 +154,6 @@ mod tests {
         let encoded = serde_cbor::to_vec(&scalar).unwrap();
         let decoded: CrossCurveDleqScalar = serde_cbor::from_slice(&encoded).unwrap();
         assert_eq!(scalar, decoded);
-    }
-
-    #[test]
-    fn serde_ecdsa_fun_sig() {
-        let secret_key = crate::bitcoin::SecretKey::new_random(&mut OsRng);
-        let sig = ECDSAFunSignature(secret_key.sign(SigHash::default()));
-        let encoded = serde_cbor::to_vec(&sig).unwrap();
-        let decoded: ECDSAFunSignature = serde_cbor::from_slice(&encoded).unwrap();
-        assert_eq!(sig, decoded);
     }
 
     #[test]
--- a/xmr-btc/src/serde.rs
+++ b/xmr-btc/src/serde.rs
@@ -1,48 +1,3 @@
-pub mod cross_curve_dleq_scalar {
-    use serde::{de, de::Visitor, Deserializer, Serializer};
-    use std::{convert::TryFrom, fmt};
-
-    struct Bytes32Visitor;
-
-    impl<'de> Visitor<'de> for Bytes32Visitor {
-        type Value = cross_curve_dleq::Scalar;
-
-        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-            write!(formatter, "a string containing 32 bytes")
-        }
-
-        fn visit_bytes<E>(self, s: &[u8]) -> Result<Self::Value, E>
-        where
-            E: de::Error,
-        {
-            if let Ok(value) = <[u8; 32]>::try_from(s) {
-                Ok(cross_curve_dleq::Scalar::from(value))
-            } else {
-                Err(de::Error::invalid_length(s.len(), &self))
-            }
-        }
-    }
-
-    pub fn serialize<S>(x: &cross_curve_dleq::Scalar, s: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        // Serialise as ed25519 because the inner bytes are private
-        // TODO: Open PR in cross_curve_dleq to allow accessing the inner bytes
-        s.serialize_bytes(&x.into_ed25519().to_bytes())
-    }
-
-    pub fn deserialize<'de, D>(
-        deserializer: D,
-    ) -> Result<cross_curve_dleq::Scalar, <D as Deserializer<'de>>::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let dleq = deserializer.deserialize_bytes(Bytes32Visitor)?;
-        Ok(dleq)
-    }
-}
-
 pub mod monero_private_key {
     use serde::{de, de::Visitor, Deserializer, Serializer};
     use std::fmt;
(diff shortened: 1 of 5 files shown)


xmr-btc/src/serde.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+pub mod bitcoin_amount {
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S>(value: &bitcoin::Amount, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_u64(value.as_sat())
```
tcharding, line 141: Why not stick with `x` and `s` for parameters like the others?

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,159 @@
...
+    db: sled::Db,
+}
+
+impl Database {
+    const LAST_STATE_KEY: &'static str = "latest_state";
+
+    pub fn open(path: &Path) -> Result<Self> {
+        let path = path
+            .to_str()
+            .ok_or_else(|| anyhow!("The path is not utf-8 valid: {:?}", path))?;
```
tcharding, line 15: No need to convert path to a string, `open` takes `AsRef<Path>`.
```
pub fn open<P: AsRef<Path>>(path: P) -> Result<Db>
```

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,159 @@
...
+    {
+        let key = serialize(&Self::LAST_STATE_KEY)?;
+        let new_value = serialize(&state).context("Could not serialize new state value")?;
+
+        let old_value = self.db.get(&key)?;
+
+        self.db
+            .compare_and_swap(key, old_value, Some(new_value))
+            .context("Could not write in the DB")?
+            .context("Stored swap somehow changed, aborting saving")?; // let _ =
```
tcharding, line 33: `// let _ =`?

tcharding (approved):
LGTM

xmr-btc/tests/harness/storage.rs
rishflab: good pick

tcharding (approved):

luckysori (commented):
Nice work!

I think we need a little bit more control over what we write to the database - I left a comment on that.

Other than that, we should move the storage code to the app crate at some point. The library should not care about storage.

xmr-btc/tests/e2e.rs
```diff
@@ -400,4 +403,126 @@ mod tests {
             initial_balances.bob_btc - swap_amounts.btc - lock_tx_bitcoin_fee
         );
     }
+
+    #[tokio::test]
+    async fn recover_protocol_state_from_db() {
```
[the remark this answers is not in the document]
rishflab: It wrote it to help me iterate on the DB implementation without having to spin up containers. If anyone wants to modify the DB code and test it quickly, it would help them to. I am also testing the case when I insert twice to the same key. I had a bug earlier where the db would break if I saved to the same key but it worked with an empty db.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,157 @@
+use anyhow::{anyhow, Context, Result};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::path::Path;
+
+pub struct Database {
+    db: sled::Db,
+}
+
+impl Database {
+    const LAST_STATE_KEY: &'static str = "latest_state";
```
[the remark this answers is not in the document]
rishflab: If we use several keys we will have to implement logic to know which key corresponds to the latest state. It is not difficult to do but it does increase the number of moving parts in the codebase.

We can achieve type safety by saving the parent state enum instead of saving the enum variants. To resume protocol execution I have to convert the enum variant to the parent enums as the protocol executor loop matches to determine the variant anyway.

Would this be sufficient?

xmr-btc/tests/e2e.rs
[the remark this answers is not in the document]
luckysori: > It wrote it to help me iterate on the DB implementation without having to spin up containers

But this test does spin up the containers, which I don't think you need to test the db. I think the storage is completely decoupled from the protocol execution. 

It would be a different story if you were testing the application and you wanted to verify that the data was being saved during execution.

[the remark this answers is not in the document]
rishflab: Oh right my bad, I mistook this test for the one in the storage module. Yes you are correct, I think this test can be removed.

xmr-btc/src/serde.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    pub fn serialize<S>(x: &cross_curve_dleq::Scalar, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // Serialise as ed25519 because the inner bytes are private
+        // TODO: Open PR in cross_curve_dleq to allow accessing the inner bytes
+        s.serialize_bytes(&x.into_ed25519().to_bytes())
```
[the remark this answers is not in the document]
luckysori: This is now done in `cross_curve_dleq` master.

xmr-btc/tests/e2e.rs
```diff
@@ -18,6 +18,8 @@ mod harness;
 const TEN_XMR: u64 = 10_000_000_000_000;
 const RELATIVE_REFUND_TIMELOCK: u32 = 1;
 const RELATIVE_PUNISH_TIMELOCK: u32 = 1;
+const ALICE_TEST_DB_FOLDER: &str = "../target/e2e-test-alice-recover";
```
bonomat, line 21: If you don't have a real reason for putting these db files into target then I would go for a tmp folder instead. I've introduced this dependency already in my PR so you can just use this:
e.g. 

```
use tempfile::tempdir;
// Create a directory inside of `std::env::temp_dir()`.
let alice_db_dir = tempdir()?;
let bob_db_dir = tempdir()?;
```


Same applies to a similar usage further down.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
+use anyhow::{anyhow, Context, Result};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::path::Path;
+
+pub struct Database<T>
```
bonomat, line 5: Is this on purpose in `tests/harness/`? The db should be used in the binary later-on right?

bonomat (approved):
Great commit messages 👍 

Two points: 
* I'm not sure if it is safe to use `../target` for the path, better use tmp folder for that
* Why is the DB/storage.rs in `tests`?

xmr-btc/src/serde.rs
```diff
@@ -0,0 +1,152 @@
+pub mod cross_curve_dleq_scalar {
```
D4nte, line 1: cross-curbe-dleq is our repo, right? It'd be good if this is pushed in the repo that defines the type (behind a feature flag) that defined in this (xmr-btc) repo.

xmr-btc/tests/e2e.rs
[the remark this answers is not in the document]
D4nte: > 🔧 I think this test is not adding much. The storage APIs are already tested in the `storage` module. And the protocol is already tested in the other tests.

From the name of the function, I would tend to disagree with you as I just discovered a bug in nectar that would reset some data everytime we load the db. reloading the db and expecting same data is a useful test I believe.

But maybe this could just be a test in storage that does not do the full execution of the swap.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+{
+    db: sled::Db,
+    _marker: std::marker::PhantomData<T>,
+}
+
+impl<T> Database<T>
+where
+    T: Serialize + DeserializeOwned,
+{
+    const LAST_STATE_KEY: &'static str = "latest_state";
```
D4nte, line 17: As this is expected to always be used as a key for sled, then serialize at instantiation to avoid bug like the one in nectar: [`d82417b` (#3316)](https://github.com/comit-network/comit-rs/pull/3316/commits/d82417b7eea2d9b09f371a7887a1efc61dee340d)

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+            sled::open(path).with_context(|| format!("Could not open the DB at {:?}", path))?;
+
+        Ok(Database {
+            db,
+            _marker: Default::default(),
+        })
+    }
+
+    pub async fn insert_latest_state(&self, state: &T) -> Result<()> {
+        let key = serialize(&Self::LAST_STATE_KEY)?;
```
D4nte, line 30: As mentioned previously, use `conquer_once::Lazy` to store this value directly serialised in a static variable so you can't forget to serialise it when using it.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+        let new_value = serialize(&state).context("Could not serialize new state value")?;
+
+        let old_value = self.db.get(&key)?;
+
+        self.db
+            .compare_and_swap(key, old_value, Some(new_value))
```
D4nte, line 36: If you are never looking at the old value then just use `insert` instead of `compare_and_swap`.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+            .context("Could not write in the DB")?
+            .context("Stored swap somehow changed, aborting saving")?;
+
+        self.db
+            .flush_async()
```
D4nte, line 41: Do note that sled automatically flushes in the background several time per second (it's configurable) so adding a todo to rethink the usage of `flush_async` is likely to help performance once done.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    pub fn get_latest_state(&self) -> anyhow::Result<T> {
+        let key = serialize(&Self::LAST_STATE_KEY)?;
+
+        let encoded = self
+            .db
+            .get(&key)?
+            .ok_or_else(|| anyhow!("State does not exist {:?}", key))?;
+
+        let state = deserialize(&encoded).context("Could not deserialize state")?;
```
D4nte, line 55: try to stack this with the previous tower.

xmr-btc/tests/harness/storage.rs
```diff
@@ -0,0 +1,158 @@
...
+pub fn deserialize<T>(v: &[u8]) -> anyhow::Result<T>
+where
+    T: DeserializeOwned,
+{
+    Ok(serde_cbor::from_slice(&v)?)
+}
+
+#[cfg(test)]
+mod tests {
+    #![allow(non_snake_case)]
```
D4nte, line 76: I know it has not been introduced with this PR but I would welcome to rethink having var that are drastically different (public key vs private key) with names that only differ by casing (s vs S).

Devs use `priv` and `pub` prefixes.

D4nte (approved):
Please check https://github.com/comit-network/xmr-btc-swap/pull/18

Few comments that would be good to tackle.

xmr-btc/src/serde.rs
luckysori: Just an FYI: it's done already.

xmr-btc/tests/harness/storage.rs
rishflab: yeh, the test hardness was the closest thing we had to a binary when i submitted this

xmr-btc/tests/harness/storage.rs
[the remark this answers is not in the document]
rishflab: Alright going to merge this as is and make the changes we discussed in a follow up PR once the usage is more clear

xmr-btc/tests/harness/storage.rs
rishflab: gonna leave this and other db changes for now as its not completely clear what/how/when we want to store and retrieve

xmr-btc/tests/harness/storage.rs
rishflab:
```
error[E0277]: the trait bound `conquer_once::raw::Lazy<std::vec::Vec<u8>, conquer_once::with_std::ParkThread>: std::convert::AsRef<[u8]>` is not satisfied
  --> xmr-btc/tests/harness/storage.rs:27:37
   |
27 |         let old_value = self.db.get(&LAST_STATE_KEY)?;
   |                                     ^^^^^^^^^^^^^^^
   |                                     |
   |                                     the trait `std::convert::AsRef<[u8]>` is not implemented for `conquer_once::raw::Lazy<std::vec::Vec<u8>, conquer_once::with_std::ParkThread>`
   |                                     help: consider adding dereference here: `&*LAST_STATE_KEY`
   |
```
Cant get it to compile when i use conquer once. I have put a todo

xmr-btc/tests/harness/storage.rs
rishflab: added todo

xmr-btc/tests/harness/storage.rs
rishflab: not sure what you mean here

xmr-btc/tests/harness/storage.rs
rishflab: added a todo

xmr-btc/tests/harness/storage.rs
D4nte: Yes

xmr-btc/tests/harness/storage.rs
D4nte: Read the help that the compiler gave you...

PR #10, opened 2020-10-12
Add initial swap binary crate

tcharding:
This draft PR is just FYI and was done so I have something to add the libp2p stuff on top of.

In preparation for writing a fully fledged XMR/BTC swap client add a new crate
and initial binary with basic structopt cli implementation.

With this applied we can run the swap as Alice with `swap --as-alice` and run
the swap as Bob with `swap --sats 10000000`.

PR #17, opened 2020-10-22
A simple cli proposal

bonomat:
What do you think of this? 

As a maker: 
```bash
 ./target/debug/xmr-btc --maker                                      
Usually we would now wait for takers to connect.
```

As a taker: 
```bash
./target/debug/xmr-btc --taker --address "asd"                      
Received order from maker: 1 BTC for 42 XMR. Take it or leave it? Hit <enter> to accept or CTRL-C to decline and quit.
 
Accepting trade doing the swap
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,30 @@
 XMR to BTC Atomic Swap
 ======================
 
+## How to start
+
+### As a maker
+
+To start the application as a maker type the following into your terminal:  
+```bash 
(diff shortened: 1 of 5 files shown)


bonomat:
Ignore this PR: missed that @tcharding was working on something similar https://github.com/comit-network/xmr-btc-swap/pull/10

PR #21, opened 2020-10-23
Move storage into application crate

rishflab:
PR to move storage into application crate. I was thinking of also integrating the storage in this PR but maybe it is better to break it up. 

I wanted to do it in the same PR so that I don't have to drop it and then reintroduce it later.

### DIFF
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -22,9 +22,12 @@ monero = "0.9"
 rand = "0.7"
 reqwest = { version = "0.10", default-features = false, features = ["socks"] }
 serde = { version = "1", features = ["derive"] }
+serde_cbor = "0.11"
 serde_derive = "1.0"
 serde_json = "1"
+sled = "0.34"
 structopt = "0.3"
+tempfile = "3"
 time = "0.2"
 tokio = { version = "0.2", features = ["rt-threaded", "time", "macros", "sync"] }
 torut = { version = "0.1", optional = true }
--- a/swap/src/storage.rs
+++ b/swap/src/storage.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, Context, Result};
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Serialize};
 use std::path::Path;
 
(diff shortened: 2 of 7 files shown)


luckysori (approved):
`cargo tomlfmt` may complain about the packages not being in alphabetical order.

swap/Cargo.toml
```diff
@@ -34,4 +34,7 @@ tracing-log = "0.1"
 tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "ansi", "env-filter"] }
 url = "2.1"
 void = "1"
-xmr-btc = { path = "../xmr-btc" }
\ No newline at end of file
+xmr-btc = { path = "../xmr-btc" }
+sled = "0.34"
+tempfile = "3"
+serde_cbor = "0.11"
```
tcharding, line 40: These should be sorted alphabetically.

bonomat (approved):
Please fix the order in cargo.toml and merge at your convenience once merge conflicts are resolved.

swap/src/storage.rs
```diff
@@ -81,13 +81,16 @@ mod tests {
...
+        serde::{bitcoin_amount, monero_private_key},
+        CrossCurveScalar, Curve25519Scalar,
+    };
 
     #[derive(Debug, Serialize, Deserialize, PartialEq)]
     pub struct TestState {
         A: xmr_btc::bitcoin::PublicKey,
         a: xmr_btc::bitcoin::SecretKey,
-        s_a: xmr_btc::cross_curve_dleq::Scalar,
+        s_a: CrossCurveScalar,
```
tcharding: I have been very vocal about my dislike for using fully qualified paths but it is the approach we have settled upon, this patch goes against all the other repos. Whilst I'm not against this change it introduces non-uniformity across the codebase, both in this repo and in the rest of the CoBloX repos. In my opinion this change should be discussed on the mailing list and a uniform approach be applied.

rishflab: I can drop this commit for now. Upon second inspection the re-exports should be renamed anyway to cross_curve::Scalar and curve25519::Scalar to be consistent with bitcoin::Amount and monero::Amount etc

PR #23, opened 2020-10-26
Do not re-export curve25519_dalek

luckysori:
Anything that needs to be re-exported by this crate from `curve25519_dalek` can be re-exported from the `monero` module. In
fact, the `Scalar` type was already being re-exported.

What do you think, @rishflab?


rishflab:
ah yes you are correct, I didnt catch that

rishflab (approved):

PR #25, opened 2020-10-26
Remove allow(unused_assignments) attribute

luckysori:
It looks like the compiler can ascertain that `message0` will be initialised by the time we use it, so it doesn't need to be an
`Option` and it doesn't need to be declared as mutable.

### DIFF
--- a/swap/src/alice.rs
+++ b/swap/src/alice.rs
@@ -28,13 +28,12 @@ use xmr_btc::{alice::State0, bob, monero};
 pub type Swarm = libp2p::Swarm<Alice>;
 
 // FIXME: This whole function is horrible, needs total re-write.
-#[allow(unused_assignments)] // Due to the mutable message0?
 pub async fn swap(
     listen: Multiaddr,
     redeem_address: ::bitcoin::Address,
     punish_address: ::bitcoin::Address,
 ) -> Result<()> {
-    let mut message0: Option<bob::Message0> = None;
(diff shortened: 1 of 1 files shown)


tcharding (approved):
Nice one! So one can declare a non-mutable local variable and set it once, wow I did not know that.

swap/src/alice.rs
```diff
@@ -82,10 +81,7 @@ pub async fn swap(
     );
     swarm.set_state0(state0.clone());
 
-    let state1 = match message0 {
-        Some(msg) => state0.receive(msg).expect("failed to receive msg 0"),
-        None => panic!("should have the message by here"),
-    };
+    let state1 = state0.receive(message0).expect("failed to receive msg 0");
```
bonomat, line 84: I don't understand how the compiler knows that this variable was defined... 

//edit// ah, because you only break the loop in a single case in which this variable is defined... very interesting.

swap/src/alice.rs
```diff
@@ -53,7 +52,7 @@ pub async fn swap(
             OutEvent::Message0(msg) => {
                 debug!("Got message0 from Bob");
                 // TODO: Do this in a more Rusty/functional way.
```
thomaseizinger, line 54: Is this TODO now solved? :)

thomaseizinger (commented):
Nice!

PR #27, opened 2020-10-26
Add tor to main

bonomat:
The idea is to have Tor enabled by default if it was compiled with the feature flag enabled. 
This leads to a few added `cfg[feature]` through the code but I think this is nicer than having `if`s everywhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run clippy
+        run: cargo clippy --workspace --all-targets -- -D warnings
+
+      - name: Run clippy all features
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
   build_test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,10 +40,10 @@ jobs:
       - name: Check code formatting
         run: cargo fmt --all -- --check
 
-      - name: Run clippy
+      - name: Run clippy with default features
         run: cargo clippy --workspace --all-targets -- -D warnings
 
-      - name: Run clippy all features
+      - name: Run clippy with all features enabled (e.g. tor)
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
   build_test:
--- a/swap/src/alice.rs
+++ b/swap/src/alice.rs
@@ -125,7 +125,10 @@ fn new_swarm(listen: Multiaddr, port: Option<u16>) -> Result<Swarm> {
     }
     #[cfg(not(feature = "tor"))]
     {
-        transport = transport::build(local_key_pair)?;
+        transport = match port {
+            None => transport::build(local_key_pair)?,
+            Some(port) => anyhow::bail!("local port should not be provided for non-tor usage"),
+        };
     }
 
     let mut swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, local_peer_id.clone())
--- a/swap/src/alice.rs
+++ b/swap/src/alice.rs
@@ -1,6 +1,6 @@
 //! Run an XMR/BTC swap in the role of Alice.
 //! Alice holds XMR and wishes receive BTC.
-use anyhow::Result;
+use anyhow::{bail, Result};
 use libp2p::{
     core::{identity::Keypair, Multiaddr},
     request_response::ResponseChannel,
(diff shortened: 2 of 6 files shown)


swap/src/cli.rs
```diff
@@ -11,4 +11,9 @@ pub struct Options {
     /// Run the swap as Bob and try to swap this many BTC (in satoshi).
     #[structopt(long = "sats")]
     pub satoshis: Option<u64>,
+
+    /// Alice's onion mulitaddr (only required for Bob, Alice will autogenerate
```
luckysori, line 15: :wrench: 
```suggestion
    /// Alice's onion multitaddr (only required for Bob, Alice will autogenerate
```

luckysori (commented):
> My idea is to have Tor enabled by default if it was compiled with the feature flag enabled.
>
> What do you think of this approach?

So we're okay with someone compiling with the feature flag and then not being able to run the binary without a tor node running? Maybe this is safer, as you get what you compiled and you cannot change it at runtime.

> Any idea on how I can test this?

I'm not sure if I would test it automatically, but you could emit a log once the tor service has been created and pass the test if you find that log.

bonomat:
> So we're okay with someone compiling with the feature flag and then not being able to run the binary without a tor node running? Maybe this is safer, as you get what you compiled and you cannot change it at runtime.

I'm ok with this :D Eventually I see this a default config to protect all of our users. 

> "Some people want privacy, everyone needs privacy", 
me, just now

swap/src/alice.rs
```diff
@@ -105,15 +106,24 @@ pub async fn swap(
...
+    let transport = {
+        #[cfg(feature = "tor")]
+        {
+            transport::build(local_key_pair, Some((listen.clone(), port)))?
+        }
+        #[cfg(not(feature = "tor"))]
+        {
+            transport::build(local_key_pair)?
+        }
+    };
```
tcharding, line 126:
```suggestion
    #[cfg(feature = "tor")]
    let transport = transport::build(local_key_pair, Some((listen.clone(), port)))?
    #[cfg(not(feature = "tor"))]
    let transport = transport::build(local_key_pair)?
```

swap/src/cli.rs
```diff
@@ -11,4 +11,9 @@ pub struct Options {
     /// Run the swap as Bob and try to swap this many BTC (in satoshi).
     #[structopt(long = "sats")]
     pub satoshis: Option<u64>,
+
+    /// Alice's onion multitaddr (only required for Bob, Alice will autogenerate
+    /// one)
```
tcharding, line 16: This comment is not clear to me. If Alice autogenerates an address how is Bob supposed to know it?

swap/src/main.rs
```diff
@@ -37,13 +36,36 @@ pub const PORT: u16 = 9876; // Arbitrarily chosen.
...
 async fn main() -> Result<()> {
     let opt = Options::from_args();
 
     trace::init_tracing(LevelFilter::Debug)?;
 
-    let addr = format!("/ip4/{}/tcp/{}", ADDR, PORT);
+    let (addr, _tor_connection) = {
+        #[cfg(feature = "tor")]
+        {
+            let tor_secret_key = TorSecretKeyV3::generate();
```
tcharding, line 55: If we generate a new address for every instance (and we currently use one instance per swap) that means that Alice must share this address with Bob _after_ she has started the node? Is this what we want?

swap/src/cli.rs
bonomat: Bob needs to know the address of "an Alice" to trade with her. So yes, Alice needs to communicate that with her.

swap/src/alice.rs
bonomat: Maybe I did something wrong but that did not work for me...

swap/src/main.rs
bonomat: There are multiple solutions to this. Simplest is: yes, Alice needs to start first and then share it with Bob. This is fine as we have a single-trade-and-terminate application. 

In a later version we can store this key, derive it from a seed, etc.

swap/src/main.rs
```diff
@@ -12,38 +12,58 @@
...
 // Alice's address and port until we have a config file.
 pub const PORT: u16 = 9876; // Arbitrarily chosen.
 pub const ADDR: &str = "127.0.0.1";
-pub const BITCOIND_JSON_RPC_URL: &str = "127.0.0.1:8332";
+pub const BITCOIND_JSON_RPC_URL: &str = "http://127.0.0.1:8332";
+
+#[cfg(feature = "tor")]
+use swap::tor::{AuthenticatedConnection, UnauthenticatedConnection};
+#[cfg(feature = "tor")]
+use torut::onion::TorSecretKeyV3;
```
luckysori, line 42: :wrench: For feature-flagged imports I prefer fully-qualified names as opposed to use statements. It should be a bit more concise and less noisy.

luckysori (approved):
LGTM.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,9 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run clippy
+        run: cargo clippy --workspace --all-targets -- -D warnings
+
+      - name: Run clippy all features
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
```
tcharding, line 47: I was confused by this at first since `--all-features` implies a superset but actually this is 'with tor' and the one above is 'without tor'. We could use `-p swap --features tor`. Or even just a comment to say so no one removes the clippy command above?

tcharding (approved):
Added a refactor patch on top. Nice work.

PR #28, opened 2020-10-26
Re work swap functions

tcharding:
Tidy up a bit in preparation for calling the generator functions from alice/bob binaries. The only notable change is that we now send back an empty response to Bob from Alice on receiving `alice::Message2`


luckysori (approved):
LGTM

PR #29, opened 2020-10-27
Add config file

rishflab:
PR to add the beginnings of a config file.

Dont feel comfortable adding any more config params right now such as the seed.pem until we have some sort of "e2e" test. Not sure if adding seed.pem is high priority right now.

### DIFF
--- a/swap/src/config.rs
+++ b/swap/src/config.rs
@@ -1,0 +1,47 @@
+use anyhow::Result;
+use libp2p::multiaddr::Multiaddr;
+use serde::{Deserialize, Serialize};
+use std::{fs, path::Path};
+use url::Url;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Config {
+    pub bitcoind_url: Url,
+    pub listen_addr: Multiaddr,
+}
+
+impl Config {
+    pub fn load(path: &Path) -> Result<Config> {
+        let contents = fs::read_to_string(path).expect("Something went wrong reading the file");
+        let file: Config = toml::from_str(contents.as_str())?;
+        Ok(file)
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            bitcoind_url: Url::parse("https://127.0.0.1:8332")
+                .expect("Failed to generate default config"),
+            listen_addr: "/ip4/127.0.0.1/tcp/9876"
+                .parse()
+                .expect("Failed to generate default config"),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Monero {
+    pub ip: Option<String>,
+    pub port: Option<u64>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_generate_default_config() {
+        let _ = Config::default();
+    }
+}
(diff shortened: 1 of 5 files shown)


swap/src/config.rs
```diff
@@ -0,0 +1,47 @@
...
+        let file: Config = toml::from_str(contents.as_str())?;
+        Ok(file)
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            bitcoind_url: Url::parse("https://127.0.0.1:8332")
+                .expect("Failed to generate default config"),
```
tcharding, line 25:
```suggestion
                .expect("Failed to parse static string"),
```

swap/src/config.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub ip: Option<String>,
+    pub port: Option<u64>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_generate_default_config() {
```
tcharding, line 44: This test is pointless :)

swap/src/config.rs
```diff
@@ -0,0 +1,47 @@
+use anyhow::Result;
+use libp2p::multiaddr::Multiaddr;
+use serde::{Deserialize, Serialize};
+use std::{fs, path::Path};
+use url::Url;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Config {
+    pub bitcoind_url: Url,
+    pub listen_addr: Multiaddr,
```
tcharding, line 10: Perhaps this should be an option? If the node is being run as Bob we don't need this.

swap/src/main.rs
```diff
@@ -13,47 +13,43 @@
...
-// TODO: Add a config file with these in it.
-// Alice's address and port until we have a config file.
-pub const PORT: u16 = 9876; // Arbitrarily chosen.
-pub const ADDR: &str = "127.0.0.1";
-pub const BITCOIND_JSON_RPC_URL: &str = "127.0.0.1:8332";
-
 #[tokio::main]
 async fn main() -> Result<()> {
-    let opt = Options::from_args();
+    let opt: Options = Options::from_args();
```
tcharding, line 35: Do we really need this change?

tcharding (approved):
LGTM

swap/src/config.rs
rishflab: something like "parse static string" is printed anyway if parse fails. The string in expect() is providing context

swap/src/config.rs
rishflab: I put this is in because the I am unwrapping when generating default config. The test will catch any changes that result it a default config that does not generate.

swap/src/main.rs
rishflab: nope, my IDE was unable to resolve the type. I can remove it.

swap/src/config.rs
rishflab: Ah right.... It is actually the address to dial to connect to a peer, not the listen addr.

rishflab:
I think we can do this without a config file and use CLI params instead. There are not that many options and prevents the filesystem from being cluttered

rishflab:
Closing, replacing with CLI based config

PR #30, opened 2020-10-27
Execute on-chain protocol after handshake

luckysori:
This now implements and tests the ability to swap between Alice and Bob using the `swap` binary.


tcharding (approved):
BOOM!

rishflab (approved):

PR #32, opened 2020-10-28
[No merge] Replace descriptor with Pubkey tuple

bonomat:
This does not seem to improve compile time :(


bonomat:
compile time on MacOS: 
```bash
cargo build --all-features  1220.47s user 71.40s system 266% cpu 8:03.96 total
```

PR #34, opened 2020-11-03
Add CLI config and remove conditional compile flags

rishflab:
The hardcoded configuration was replaced with CLI
configuration options. CLI based config was chosen
over a config file as it does not access and clutter
the user's file system. By CLI options depend on whether
the program is run in Alice or Bob mode.

The tor conditional compile flags were removed by
extracting transport creation to the main statement. A tor
transport is created if Alice specifies a tor port using the CLI.

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -4,41 +4,32 @@ use url::Url;
 #[derive(structopt::StructOpt, Debug)]
 pub enum Options {
     Alice {
-        /// Run the swap as Bob and try to swap this many BTC (in satoshi).
-        #[structopt(long = "sats")]
-        satoshis: u64,
-
         // /// Run the swap as Bob and try to swap this many XMR (in piconero).
         // #[structopt(long = "picos", conflicts_with = "sats"))]
         // pub piconeros: u64,
         #[structopt(default_value = "http://127.0.0.1:8332", long = "bitcoind")]
         bitcoind_url: Url,
 
-        #[structopt(default_value = "127.0.0.1", long = "listen_addr")]
-        listen_addr: String,
+        #[structopt(default_value = "/ip4/127.0.0.1/tcp/9876", long = "listen-addr")]
+        listen_addr: Multiaddr,
 
-        #[structopt(default_value = 9876, long = "list_port")]
-        listen_port: u16,
+        /// Run the swap as Bob and try to swap this many BTC (in satoshi).
+        // #[cfg(feature = "tor")]
+        #[structopt(long = "tor-port")]
+        tor_port: Option<u16>,
     },
     Bob {
-        /// Alice's multitaddr (only required for Bob, Alice will autogenerate
-        /// one)
-        #[structopt(long = "alice_addr")]
-        alice_addr: Multiaddr,
-
         /// Run the swap as Bob and try to swap this many BTC (in satoshi).
         #[structopt(long = "sats")]
         satoshis: u64,
+
+        #[structopt(long = "alice-addr")]
+        alice_addr: Multiaddr,
 
         // /// Run the swap as Bob and try to swap this many XMR (in piconero).
         // #[structopt(long = "picos", conflicts_with = "sats"))]
         // pub piconeros: u64,
         #[structopt(default_value = "http://127.0.0.1:8332", long = "bitcoind")]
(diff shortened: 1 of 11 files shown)


swap/src/main.rs
```diff
@@ -30,95 +37,97 @@ mod trace;
...
-pub const ADDR: &str = "127.0.0.1";
-pub const BITCOIND_JSON_RPC_URL: &str = "http://127.0.0.1:8332";
-
-#[cfg(feature = "tor")]
-pub const TOR_PORT: u16 = PORT + 1;
-
 #[tokio::main]
 async fn main() -> Result<()> {
-    let opt = Options::from_args();
+    let opt: Options = Options::from_args();
```
luckysori, line 42: :wrench: Looks unnecessary.

swap/src/cli.rs
```diff
@@ -1,19 +1,35 @@
+use libp2p::core::Multiaddr;
+use url::Url;
+
 #[derive(structopt::StructOpt, Debug)]
-pub struct Options {
-    /// Run the swap as Alice.
-    #[structopt(long = "as-alice")]
-    pub as_alice: bool,
+pub enum Options {
```
luckysori, line 5: :wrench: Looks like some docstrings are missing and/or outdated here.

swap/src/main.rs
```diff
@@ -30,95 +37,97 @@ mod trace;
...
-        info!("running swap node as Alice ...");
-
-        if opt.piconeros.is_some() || opt.satoshis.is_some() {
-            bail!("Alice cannot set the amount to swap via the cli");
+    match opt {
+        Options::Alice {
+            bitcoind_url: url,
+            listen_addr,
+            tor_port,
+            ..
```
luckysori, line 51: :question: Aren't you matching against all fields?

swap/src/main.rs
```diff
@@ -30,95 +37,97 @@ mod trace;
...
+                .expect("failed to create bitcoin wallet");
+
+            let redeem = bitcoin_wallet
+                .new_address()
+                .await
+                .expect("failed to get new redeem address");
+            let punish = bitcoin_wallet
+                .new_address()
+                .await
+                .expect("failed to get new punish address");
```
luckysori, line 88: Once you rebase address generation should be inside of the `{alice, bob}::swap` functions. Just an FYI.

swap/src/main.rs
```diff
@@ -30,95 +37,97 @@ mod trace;
...
+            listen_addr,
+            tor_port,
+            ..
+        } => {
+            info!("running swap node as Alice ...");
+
+            let behaviour = Alice::default();
+            let local_key_pair = behaviour.identity();
+
+            let (listen_addr, _ac, transport) = match tor_port {
```
luckysori, line 58: :question: Is it only Alice that can run her binary through `tor`? @bonomat is this how it's meant to be?

luckysori (approved):
LGTM!

swap/src/main.rs
bonomat: If Bob always dials Alice and Alice is behind Tor then both parties need Tor.

Differently said: 
* `Tor` --> _dials_ --> `No-Tor`: ✔️
* `Tor` --> _dials_ --> `Tor`: ✔️
* `No-Tor` --> _dials_ --> `No-Tor`: ✔️
* `No-Tor` --> _dials_ --> `Tor`: x

Bob does not need to add a tor service as Alice won't dial Bob.

swap/src/main.rs
rishflab: far out this one keeps getting me, doesn't IDE can't handle it for some reason

swap/src/main.rs
luckysori: Thanks for the explanation! We therefore must support Bob starting the binary with Tor. Is that currently possible, @rishflab?

swap/src/main.rs
bonomat: `swap` is completely broken in my IDE :(

swap/src/main.rs
rishflab: yep :)

swap/src/main.rs
rishflab: Yep, fixed

PR #35, opened 2020-11-04
No need to send Monero transfer proof from Alice to Bob

luckysori:


rishflab (approved):
LGTM!

monero-harness/src/lib.rs
```diff
@@ -136,11 +136,7 @@ impl<'c> Monero {
         monerod.start_miner(&miner_address).await?;
 
         tracing::info!("Waiting for miner wallet to catch up...");
-        let block_height = monerod.client().get_block_count().await?;
-        miner_wallet
-            .wait_for_wallet_height(block_height)
-            .await
-            .unwrap();
+        miner_wallet.refresh().await?;
```
bonomat, line 139: 🚧
This is unfortunately not equal and I had problems here: 
* `refresh` asks the monerod for blockheight and refreshes until then
* `wait_for_blockheight` waits until the wallet has seen `block_height`. 

This can be a problem here as monerod might not be finished with initializing the 70 blocks and a request from the wallet to monerod could return anything < 70.

swap/Cargo.toml
```diff
@@ -36,7 +36,7 @@ torut = { version = "0.1" }
 tracing = { version = "0.1", features = ["attributes"] }
 tracing-core = "0.1"
 tracing-futures = { version = "0.2", features = ["std-future", "futures-03"] }
-tracing-log = "0.1"
+tracing-log = { version = "0.1" }
```
bonomat, line 39: ?

monero-harness/src/lib.rs
luckysori: > This can be a problem here as monerod might not be finished with initializing the 70 blocks and a request from the wallet to monerod could return anything < 70

But we call this a few lines above:

```rust
let block = monerod.client().generate_blocks(70, &miner_address).await?;
```

Doesn't that mean that we wait for the blocks to be generated before we continue?

Also, all the tests pass on CI, so it would seem like this doesn't cause any problems... If you are very confident that it is necessary we can revert the change. It was just an optimisation anyway.

bonomat: My assumption is it depends on how fast your machine is. Let's leave it as you propose and if it becomes a problem we can revert.

swap/Cargo.toml
luckysori: I was messing around with tracing and forgot to revert to the simplest version of this. I'll change it.

monero-harness/src/rpc/wallet.rs
```diff
@@ -236,14 +236,14 @@ impl Client {
...
-        spend_key: &str,
+        spend_key: Option<&str>,
         view_key: &str,
     ) -> Result<GenerateFromKeys> {
         let params = GenerateFromKeysParams {
             restore_height: 0,
             filename: view_key.into(),
             address: address.into(),
-            spendkey: spend_key.into(),
+            spendkey: spend_key.map(|sk| sk.into()),
```
D4nte, line 246:
```suggestion
            spendkey: spend_key.map(Into::into),
```

monero-harness/tests/wallet.rs
```diff
@@ -61,3 +64,51 @@ async fn fund_transfer_and_check_tx_key() {
...
+    let (view_sk, view_pk) = {
+        let scalar = Scalar::random(&mut OsRng);
+        let sk = PrivateKey::from_scalar(scalar);
+        let pk = PublicKey::from_private_key(&sk);
+
+        (sk, pk)
+    };
+    let address = Address::standard(Network::Mainnet, spend_pk, view_pk);
+
+    let one_xmr = 1_000_000_000_000;
```
D4nte, line 92: If often used in test then it could be a const.

What's the name of the smallest monero unit again?

swap/src/monero.rs
```diff
@@ -1,58 +1,60 @@
...
 use url::Url;
+
 pub use xmr_btc::monero::{
     Amount, CreateWalletForOutput, InsufficientFunds, PrivateViewKey, PublicKey, PublicViewKey,
-    Transfer, TransferProof, TxHash, WatchForTransfer, *,
+    Transfer, *,
 };
 
-pub struct Wallet(pub wallet::Client);
+pub struct Facade {
```
D4nte, line 15: IMO, `Facade` is a very generic term. Why is it not a wallet anymore?

Maybe just call it `Wallets` as I can see it wraps both the user wallet and the readonly wallet.

swap/src/bob.rs
```diff
@@ -142,13 +105,19 @@ pub async fn swap(
     };
 
     swarm.send_message2(alice.clone(), state2.next_message());
+    // NOTE: We must poll the swarm after `send_messageX` to actually trigger
```
D4nte, line 108: Did you chat about that with @thomaseizinger? :)

swap/src/network/request_response.rs
```diff
@@ -40,7 +40,7 @@ pub enum AliceToBob {
     Amounts(SwapAmounts),
     Message0(alice::Message0),
     Message1(alice::Message1),
-    Message2(alice::Message2),
+    Message2, // empty response
```
D4nte, line 43: Why responding if it's empty?

xmr-btc/src/alice.rs
```diff
@@ -93,10 +92,13 @@ where
         + Send
         + Sync
         + 'static,
+    M: monero::WatchForTransfer + Send + Sync + 'static,
 {
+    #[allow(clippy::enum_variant_names)]
     #[derive(Debug)]
     enum SwapFailed {
```
D4nte, line 99:
```suggestion
    enum SwapStateAtFailure {
```

Am I correct?

monero-harness/tests/wallet.rs
luckysori: Piconero (or atomic unit) https://web.getmonero.org/resources/moneropedia/atomic-units.html

swap/src/monero.rs
luckysori: Yeah, it's a bad name.

swap/src/bob.rs
luckysori: Nope, I never got to the bottom of this.

swap/src/network/request_response.rs
luckysori: This is how I understand it: each message exchange is part of a request-response protocol, always initiated by Bob (since Alice is the owner of the known address). I would assume that you could just omit empty responses, but then the individual request-response protocol would be unfinished.

xmr-btc/src/alice.rs
luckysori: Sure, it's just so long.

luckysori:
@D4nte feel free to take over the PR and even close it if you don't need it. It is ready to be merged in terms of functionality, so there's nothing else that I wanted to do with it.

xmr-btc/src/alice.rs
D4nte: just call it `SwapState`?

D4nte:
> @D4nte feel free to take over the PR and even close it if you don't need it. It is ready to be merged in terms of functionality, so there's nothing else that I wanted to do with it.

I'll probably catch-up with @rishflab when he is back and we can go through open PR and dispatch work.

swap/src/bob.rs
thomaseizinger: You should definitely be polling the swarm outside of this in a loop somewhere. In a production application, the swarm can emit all kinds of events all the time, not just the message you are waiting for.

There are several ways of how you can do this, one of them being the way nectar does things and connect the swaps with the main loop through a channel.

thomaseizinger: To receive a message you will probably want to have a channel in the opposite direction. The sending end could be sent as part of the first message.

thomaseizinger: This is effectively creating an ad-hoc actor system, so you might also want to consider using one.

xmr-btc/src/alice.rs
```diff
@@ -146,15 +150,30 @@ where
...
-            // TODO: Watch for LockXmr using watch-only wallet. Doing so will prevent Alice
-            // from cancelling/refunding unnecessarily.
+            let monero_joint_address = monero::Address::standard(
+                monero::Network::Mainnet,
+                public_spend_key,
+                public_view_key.into(),
+            );
+
+            if let Either::Right(_) = select(
+                monero_client.watch_for_transfer_improved(monero_joint_address, xmr, v),
```
D4nte: Shouldn't Bob be the one watching for the transfer?
Before: Alice locks xmr, sends a transfer proof to Bob, Bob assumes xmr is locked because of the proff
After: Alice locks xmr, Bob watches the blockchain using the view key to confirm the xmr is locked.

luckysori: You are correct that Bob has to watch the blockchain using the view key.

But in this generator-based approach, for Alice we yield an action which may or may not be performed by the consumer of this generator, so we need to verify that this happened before we continue. Now, this is not strictly necessary because one can assume that if we receive the redeem encrypted signature from Bob then the monero must have been locked up correctly.

Therefore, in your re-implementation of the protocol I think you can safely omit this if you want.

PR #36, opened 2020-11-04
Recover from a failed swap

rishflab:
Adds a `recover <swap-id>` command which will attempt to finish a swap saved to the database.

The whole `recover.rs` module is a bit of a temporary hack in my view, as recovery should be handled via the `action_generator`s, like normal swap execution (also in my view). 

Alternatively, we could have gone for a simpler manual recovery like what is currently offered in `cnd`, where a user asks for specific transactions to be constructed so that they can do them themselves. That would be cleaner and simpler, but this protocol is so much more complicated than the one currently supported by `cnd` that I'm not sure it's a good idea.


swap/src/recover.rs
```diff
@@ -134,10 +134,87 @@ pub async fn alice_recover(
...
+        Alice::BtcRedeemable { redeem_tx, state } => {
+            let tx_lock_height = bitcoin_wallet
+                .transaction_block_height(state.tx_lock.txid())
+                .await;
+
+            let block_height = bitcoin_wallet.0.block_height().await?;
+            let refund_absolute_expiry = tx_lock_height + state.refund_timelock;
+
+            // bob cannot cancel
+            if block_height > refund_absolute_expiry {
```
luckysori, line 146: 📣 Always tricky with these things, but I think it's this way round.

```suggestion
            if block_height < refund_absolute_expiry {
```

swap/src/recover.rs
```diff
@@ -134,10 +134,87 @@ pub async fn alice_recover(
...
+                    .await?;
+            } else {
+                let tx_cancel = TxCancel::new(
+                    &state.tx_lock,
+                    state.refund_timelock,
+                    state.a.public(),
+                    state.B.clone(),
+                );
+
+                let tx_cancel_height = bitcoin_wallet
```
luckysori, line 158: 📣 I think we're missing broadcasting the cancel transaction. If we just wait for the other party to do it, it may never happen. I think we can simply broadcast it unconditionally just before this line, because we know that it can be included on the blockchain i.e. timelock has been reached.

swap/src/recover.rs
```diff
@@ -134,10 +134,87 @@ pub async fn alice_recover(
...
+                        monero_wallet
+                            .create_and_load_wallet_for_output(s_a + s_b, state.v)
+                            .await?;
+                    }
+                    Either::Right(_) => {
+                        let tx_punish =
+                            TxPunish::new(&tx_cancel, &state.punish_address, state.punish_timelock);
+
+                        let sig_a = state.a.sign(tx_punish.digest());
+                        let sig_b = state.tx_cancel_sig_bob.clone();
```
luckysori, line 204: 📣 Wrong signature.

```suggestion
                        let sig_b = state.tx_punish_sig_bob.clone();
```

luckysori (changes requested):
Caught some mistakes 😛

luckysori (approved):

swap/src/recover.rs
```diff
@@ -0,0 +1,481 @@
...
+            );
+
+            info!("Checking if the Bitcoin cancel transaction has been published");
+            if bitcoin_wallet
+                .0
+                .get_raw_transaction(tx_cancel.txid())
+                .await
+                .is_err()
+            {
+                info!("Bitcoin cancel transaction not yet published");
```
D4nte, line 68: Mention that we are now waiting for it.

swap/src/recover.rs
```diff
@@ -0,0 +1,481 @@
...
+
+                let tx_lock_height = bitcoin_wallet
+                    .transaction_block_height(state.tx_lock.txid())
+                    .await;
+                poll_until_block_height_is_gte(
```
D4nte, line 73:
```suggestion
                poll_until_block_height(
```
We can only poll until it is greater than, the name can be shorten

swap/src/state.rs
```diff
@@ -0,0 +1,88 @@
...
+            Swap::Alice(alice) => Display::fmt(alice, f),
+            Swap::Bob(bob) => Display::fmt(bob, f),
+        }
+    }
+}
+
+impl Display for Alice {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Alice::Handshaken(_) => f.write_str("Handshake complete"),
```
D4nte, line 66:
```suggestion
            Alice::HandshakeComplete(_) => f.write_str("Handshake complete"),
```
If the text that explain a code name differs from it, then it's a good education that the code naming could be improved :)

swap/tests/e2e.rs
```diff
@@ -1,117 +1,124 @@
...
+use std::sync::Arc;
+use swap::{alice, bob, network::transport::build, storage::Database};
+use tempfile::tempdir;
+use testcontainers::clients::Cli;
+use xmr_btc::bitcoin;
+
+// NOTE: For some reason running these tests overflows the stack. In order to
+// mitigate this run them with:
+//
+//     RUST_MIN_STACK=100000000 cargo test
```
D4nte, line 14: Add this to the readme please.

D4nte (commented):
I agree this feels very copy-pastie. Being able to use the asme flow to execute and resume would reduce maintenance effort and the chances of updating the execution but forgetting to update the recovery path.

swap/tests/e2e.rs
luckysori: This is no longer the case after #39 was merged.

luckysori:
@D4nte as with the other PR, this is feature-complete, so I think now you guys can make some minor changes and merge it or close it if you prefer. Other than actually using the storage and adding the recovery module, we did a few other useful things:

- https://github.com/comit-network/xmr-btc-swap/pull/36/commits/58931e3a05f2e2517e04bcd4e3cddce7c346a4c2.
- https://github.com/comit-network/xmr-btc-swap/pull/36/commits/c5ce22d851b7bd38fafa0c4ed62e71e46229e9dc (debatable whether the URL provided should have `/json_rpc` appended or if we should append it in the code, but this should fix it for now).
- https://github.com/comit-network/xmr-btc-swap/pull/36/commits/ec1a96ccc5c03de10cf77e2b67b13d80970d6882#diff-76d0cc2f843ff08a24e4a15bb19c42e3ee12070a676f8bd7c8b604bec5a1e4e1R190-R192 (the limitations documented in swap/src/bob.rs).
- https://github.com/comit-network/xmr-btc-swap/pull/36/commits/9d7cba8a50f67b1134eac12f26ce8509943f4a9b.
- https://github.com/comit-network/xmr-btc-swap/pull/36/commits/e60a2bdcfe543581e33831adc24e448b14fb88bd (fixed a bug in xmr-btc/src/alice.rs, explained in the commit message).

PR #37, opened 2020-11-05
Tor e2e test

bonomat:
First patch allows to configure Tor more. 

Second patch is questionable if we want it, it's a bit ugly imho. 

Why did we remove the feature flag for Tor please? Since we don't have any atm someone who does not have Tor installed on his machine cannot run the e2e tests. Is that what we want?

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -14,8 +14,18 @@ pub enum Options {
         #[structopt(default_value = "/ip4/127.0.0.1/tcp/9876", long = "listen-addr")]
         listen_addr: Multiaddr,
 
-        #[structopt(long = "tor-port")]
-        tor_port: Option<u16>,
+        #[structopt(long = "use-tor")]
+        tor: bool,
+
+        /// Local port Alice will be listenning on if tor is used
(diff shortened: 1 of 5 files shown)


luckysori:
@bonomat I did not make the decision to remove the feature flag, but I think it did introduce quite a bit of complexity for not a lot of gain.

To ensure that anyone can run the tests locally we can either:

1. Not have a tor e2e test.
2. Have a tor e2e test but mark it with `#[ignore]`, so that it's not required for CI or for local test runs of the entire crate/codebase.
3. Re-introduce the feature flag `tor`, but only use it to control whether the tor test is run or not. That way you can still keep it as part of CI, but only someone running the tests with `--all-features` or similar would execute it locally.

D4nte:
I would question whether the PoC needs tor at this moment in time.

bonomat:
> I would question whether the PoC needs tor at this moment in time.

My plan was to push for this early on and eventually make it a must-have default :(

rishflab:
We decided to focus on robustness for now and come to adding tor functionality later

PR #39, opened 2020-11-09
Update cross-curve-dleq dependency

luckysori:
Making build times considerably faster.

On my machine, after running `cargo clean`, `cargo build -p swap` takes 2min 19s.

The updated dependency also comes with a critical fix to the `Scalar` type, which originally wrongly assumed that secp256k1 and ed25519 scalars had the same endianness. For this reason, we now have to reverse the bytes of recovered scalars if we are to use them on a different chain.

Finally, there is no need to append `RUST_MIN_STACK=100000000` to avoid stack overflows in tests and when running the binary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,4 @@ jobs:
       - name: Cargo test
         run: cargo test --workspace --all-features
         env:
-          # To avoid stack overflows
-          RUST_MIN_STACK: 100000000
           MONERO_ADDITIONAL_SLEEP_PERIOD: 60000
--- a/xmr-btc/Cargo.toml
+++ b/xmr-btc/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 anyhow = "1"
 async-trait = "0.1"
 bitcoin = { version = "0.23", features = ["rand", "serde"] }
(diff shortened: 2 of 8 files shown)


bonomat:
How did you stop the time? 
I used `time` like this: 

```bash 
cargo clean && time cargo build --workspace
```
and get
```bash 
cargo build --workspace  815.78s user 76.55s system 378% cpu 3:55.47 total
```

Which is quite some improvement though :)

luckysori:
@bonomat Instead of building the whole workspace I only built `swap` which has always been the slow step, getting:

`cargo build -p swap  813,26s user 47,76s system 617% cpu 2:19,37 total`

bonomat:
I was more confused about the `2min 19s.`

bonomat (approved):

luckysori:
> I was more confused about the `2min 19s.`
> 
> Edit: I get this for swap only:
> 
> ```shell
> cargo build -p swap  782.58s user 68.97s system 458% cpu 3:05.66 total
> ```

image

bonomat:
> Finally, there is no need to append RUST_MIN_STACK=100000000 to avoid stack overflows in tests and when running the binary.

if this is not needed anymore, can we remove it from the CI setup file?

luckysori:
> > Finally, there is no need to append RUST_MIN_STACK=100000000 to avoid stack overflows in tests and when running the binary.
> 
> if this is not needed anymore, can we remove it from the CI setup file?

We can and we did :D

PR #41, opened 2020-11-13
Simplify swap and recovery execution

rishflab:
Consolidate and simplify swap execution. Generators are no longer
needed. Consolidate recovery and swap data structures. The
recursive calls can be replaced with a loop if returning prior to
completion is desired for testing purposes.


swap/src/alice.rs
```diff
@@ -433,8 +446,8 @@ impl Default for Alice {
 }
 
 fn calculate_amounts(btc: ::bitcoin::Amount) -> SwapAmounts {
-    // TODO: Get this from an exchange.
-    // This value corresponds to 100 XMR per BTC
+    // TODO (Franck): This should instead verify that the received amounts matches
```
da-kami, line 449: ❓ I am not sure I understand this comment correctly. Is the goal to verify that the amount received (after the swap) matches the amount that was passed in?

swap/src/bob/swap.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+        BobState::Cancelled(_state) => Ok(BobState::BtcRefunded),
+        BobState::BtcRefunded => Ok(BobState::BtcRefunded),
+        BobState::Punished => Ok(BobState::Punished),
+        BobState::SafelyAborted => Ok(BobState::SafelyAborted),
+        BobState::XmrRedeemed => Ok(BobState::XmrRedeemed),
+    }
+}
+
+// // State machine driver for recovery execution
```
da-kami, line 208: Maybe remove this?

D4nte (approved):

da-kami (approved):
Let's get it in, totally fine to address the cleanup comments in follow ups :)

PR #42, opened 2020-11-17
XMR<>BTC user interface prototype details

da-kami:
This document sums up assumptions and limitations of the current setup and protocol.
It outlines how the UI prototypes for XMR<>BTC swaps are planned and why.

FYI @bonomat @thomaseizinger @luckysori - I don't expect a detailed review from you :)

### DIFF
--- a/docs/research/user-experience/xmr-btc-user-interface-definition.md
+++ b/docs/research/user-experience/xmr-btc-user-interface-definition.md
@@ -98,7 +98,6 @@ We could enable communication between blockchain nodes and the browser through e
 ### Browser Extension
 
 Browser extensions, are not as restrictive as webpages as can be seen in the [Google Chrome's documentsion](https://developer.chrome.com/extensions/xhr).
-The code of a browser extension can be signed and is easily verifiable.
 A browser extension would be able to offer a well-defined API, that allows multiple different web-sites to offer services within the extension.
 This could range from discovery, negotiation to execution. 
 
@@ -125,7 +124,7 @@ For previous MVPs we used [Electron](https://www.electronjs.org/) to implement c
 ## Use-case specific prototypes
 
 Given the assumptions, technical limitations and resulting possible solutions described above this section specifies details for the prototypes described in the [initial project description](https://github.com/coblox/spikes/blob/master/0003-xmr-btc-product.md).
-In the initial project description we define two products to be be mocked as prototypes, (A) single market maker and (B) peer-to-peer decentralized trading platform. 
+In the initial project description we define two products to be mocked as prototypes, (A) single market maker and (B) peer-to-peer decentralized trading platform. 
 
 Note, that the use-cases of a single market maker product and a peer-to-peer decentralized trading platform are agnostic to the user interface.
 Both products can be realised as CLI or UI taken into account the assumption listed above.
@@ -136,7 +135,7 @@ When it comes to prototyping specific user interface details, like for example t
 ### A. Single Market Maker product
 
 The point of this product is to keep `finding a trading partner` (discovery) and `agreeing on a trade` (negotiation) simple by only having a single maker. 
-A single-maker runs a webpage that gives a price. A taker can "take it of leave it".
+A single-maker runs a webpage that gives a price. A taker can "take it or leave it".
 
 We plan to create two prototypes:
 
--- a/docs/research/user-experience/xmr-btc-user-interface-definition.md
+++ b/docs/research/user-experience/xmr-btc-user-interface-definition.md
@@ -1,0 +1,145 @@
+# XMR<>BTC Atomic Swap - User Interface Prototypes
+
+This document specifies assumptions concerning possible user interfaces on top of the [Monero-Bitcoin swap as described by Lucas on the COMIT-blog](https://comit.network/blog/2020/10/06/monero-bitcoin/).
+
+We first specify assumptions and limitations imposed by the current setup and software needed for the swap, then outline possible solutions. 
+This document will be used as a basis for defining and evaluating low-fidelity UI prototypes in the prototyping tool Figma.
+
+## Assumptions
+
+This section sums up assumptions around the current setup, protocol and existing software. 
+
+### Technical Limitations
+
+Swap-solutions that run completely in the browser (webpage) can be problematic because blockchain nodes, notably [bitcoind](https://github.com/bitcoin/bitcoin/issues/11833), do not specifying [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) in their response.
+This means the browser will reject a response from a locally running Bitcoin node.
+
+Theoretically this could be overcome by creating a node infrastructure behind a proper REST API, or with a locally running proxy. However, it is questionable if these are favourable solution, as they adds additional software on top of the blockchain nodes.  
+We will focus on solutions that favour the use of locally running blockchain nodes.
+
+### Current Setup
+
+To be able to provide a reasonably trustless setup, where the user is in control of his private keys the user is currently advised to run his own Bitcoin and Monero node.
+
+Given the current setup this results in the following software necessary to achieve a swap: 
+
+* Bitcoin Blockchain Node
+* Monero Blockchain Node
+* Bitcoin Wallet
+* Monero Wallet
+* **Swap Execution Daemon**: Monitor the blockchain and move the swap forward to the next action.
+* **Interface to connect the trading partners**: The two parties have to find each other and start the `Swap Execution Daemon` with the correct parameters. The interface can be realised in multiple different ways, from single-maker, P2P trading platform to P2P marketplace. 
+
+For Bitcoin the current setup relies on a synced bitcoind node and bitcoind's wallet. 
+
+For Monero the current setup relies on a synced monerod node and monero's wallet RPC. For funding Monero an existing wallet is unlocked and used for the swap. The current setup reliese on generating a *new* Monero wallet for `redeeming` or `refunding`.
+This is not optimal for the user experience, as it means that a user cannot redeem or refund to an existing Monero wallet in the current setup. We decided **not** to model this restriction of the current setup in the prototypes, but to model the prototypes using an existin Monero wallet. 
+We believe that it should be possible to use an existing wallet with investing more implementation time into the setup and protocol.
+
+### Current Protocol
+
+The POC protocol is asymmetric and comes with several constraints that affect the user experience: 
+
+1. The party that funds Bitcoin has to fund first.
+2. The party that funds Bitcoin is always in the role of Bob.
+3. There is an information exchange (Bob sends Alice a key-part for redeeming) during the execution of the swap which requires (P2P) communication between the parties.
+
+Even though that is not per se limiting we have to acknowledge that the protocol will behave different for the maker and taker for the respective direction of the swap.
+
+The protocol comes with different incentives for the two roles Alice and Bob. 
+We think that, for a maker it is favourable to be in the role of Alice, because in Alice is in the powerful position of holding the secret. Additionally Bob funds first in the POC.
+However, to be able to allow a product that offers XRM->BTC and BTC->XMR swaps, we decided to let the maker and taker be in either role.
+The UX of the swap might be different depending on the role a party takes, especially when depicting the swap steps.
+
+### Additional Assumptions
+
+The first version of a prototype will not include hardware wallet (e.g. Ledger Nano S) support.
+In previous products we focused on swaps directly from and to cold storage, however, this does not concern a large enough user base to support it in the initial version. 
+Furthermore, hardware wallets add certain restrictions (forced user-interaction to sign transactions) that might conflict with the current prototol design.
+Hardware wallet support can be added at a later point if requested as feature by the community.
+
+## Possible Solutions
+
+This section outlines possible solutions based on the technical limitations and assumptions stated above.
+This section does not focus on the UX of the swap and the use-case specifics, but on the flow of the application considering restrictions to the browser and applications running next to it.
+
+We have explored the following options:
+
+1. Extending existing (wallet) UIs (e.g. monero-wallet-gui) for swaps.
+2. Webpage based swaps (completely running in browser).
+3. Browser extension for swaps.
+4. Desktop application for swaps.
+
+Note that for desktop applications (4.) we explore the possibility of triggering the application through the browser.
+
+### Extending existing (wallet) UIs
+
+The nature of a swap requires management of both assets of the swap to be able to fund and redeem. 
+This means, that a swap execution software has to interact with wallet software on both sides. 
+
+Additionally, due to the complexity of the current swap protocol's setup and the fact that there is information exchange during the execution phase, it is highly questionable that communities on both sides would extend wallet functionality to allow swaps. 
+For the time being the way forward seems to be a swap tool that plugs into existing wallet and blockchain node software on both sides, but handles additional requirements necessary for swap execution without modifying the existing software.
+
+### Webpage
+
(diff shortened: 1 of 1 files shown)


docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,145 @@
...
+* Monero Blockchain Node
+* Bitcoin Wallet
+* Monero Wallet
+* **Swap Execution Daemon**: Monitor the blockchain and move the swap forward to the next action.
+* **Interface to connect the trading partners**: The two parties have to find each other and start the `Swap Execution Daemon` with the correct parameters. The interface can be realised in multiple different ways, from single-maker, P2P trading platform to P2P marketplace. 
+
+For Bitcoin the current setup relies on a synced bitcoind node and bitcoind's wallet. 
+
+For Monero the current setup relies on a synced monerod node and monero's wallet RPC. For funding Monero an existing wallet is unlocked and used for the swap. The current setup reliese on generating a *new* Monero wallet for `redeeming` or `refunding`.
+This is not optimal for the user experience, as it means that a user cannot redeem or refund to an existing Monero wallet in the current setup. We decided **not** to model this restriction of the current setup in the prototypes, but to model the prototypes using an existin Monero wallet.
```
bonomat, line 36: Note: there is a relatively simple solution to that problem: we can temporarily create a new wallet for redeem/refund key and spend the output directly to an address from the user's wallet. This means the protocol has one more on-chain transaction but you can "ignore" this problematic.

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,145 @@
...
+The nature of a swap requires management of both assets of the swap to be able to fund and redeem. 
+This means, that a swap execution software has to interact with wallet software on both sides. 
+
+Additionally, due to the complexity of the current swap protocol's setup and the fact that there is information exchange during the execution phase, it is highly questionable that communities on both sides would extend wallet functionality to allow swaps. 
+For the time being the way forward seems to be a swap tool that plugs into existing wallet and blockchain node software on both sides, but handles additional requirements necessary for swap execution without modifying the existing software.
+
+### Webpage
+
+A complete webpage based solution is unfortunately not easily possible because of the restrictions of CORS headers and blockchain nodes not being able to supply CORS headers.
+The swap protocol implementation could be done in javascript, or run as web-assembly, but since the current setup requires extensive communication with a (local) wallet
```
bonomat, line 86: Sentence incomplete?

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,145 @@
...
+
+Mozilla's [web-ext tool](https://github.com/mozilla/web-ext) - besides [other tools](https://extensionworkshop.com/documentation/develop/browser-extension-development-tools/) - can help enable cross-browser extensions.
+
+### Desktop Application
+
+A desktop application is a valid solution that is not subject to restrictions by the browser.
+For previous MVPs we used Electron to implement cross-platform desktop applications. 
+
+The problem of finding a trading partner can be outsourced to a webpage (e.g. single-maker only).
+Custom URL schemes can help to provide a better user experience, as they allow us to open the swap desktop application with parameters passed to the application through the customer URL.
```
bonomat, line 109: COMIT Links!!! 🥳

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,145 @@
...
+We first specify assumptions and limitations imposed by the current setup and software needed for the swap, then outline possible solutions. 
+This document will be used as a basis for defining and evaluating low-fidelity UI prototypes in the prototyping tool Figma.
+
+## Assumptions
+
+This section sums up assumptions around the current setup, protocol and existing software. 
+
+### Technical Limitations
+
+Swap-solutions that run completely in the browser (webpage) can be problematic because blockchain nodes, notably [bitcoind](https://github.com/bitcoin/bitcoin/issues/11833), do not specifying [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) in their response.
```
D4nte, line 14: My opinion on this matter is that it is a feature, not a technical limitation. I would not want random downloaded javascript to control my bitcoind, even if we provide a way to verify the JS, users usually don't and then someone could run a single market maker website with malicious JS that steal coins.

The client that access bitcoind/monerod should be explicitly downloaded by the user. Whether it's a desktop app (self built or signed) or an browser extension (signed).

I'd be more keen to work with this from a security point of view than "boo bitcoind restricts us" :)

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,145 @@
...
+For Monero the current setup relies on a synced monerod node and monero's wallet RPC. For funding Monero an existing wallet is unlocked and used for the swap. The current setup reliese on generating a *new* Monero wallet for `redeeming` or `refunding`.
+This is not optimal for the user experience, as it means that a user cannot redeem or refund to an existing Monero wallet in the current setup. We decided **not** to model this restriction of the current setup in the prototypes, but to model the prototypes using an existin Monero wallet. 
+We believe that it should be possible to use an existing wallet with investing more implementation time into the setup and protocol.
+
+### Current Protocol
+
+The POC protocol is asymmetric and comes with several constraints that affect the user experience: 
+
+1. The party that funds Bitcoin has to fund first.
+2. The party that funds Bitcoin is always in the role of Bob.
```
D4nte, line 44: This does not matter in itself without explaining who as a probabilistic american option and who has not.

D4nte (approved):

docs/research/user-experience/xmr-btc-user-interface-definition.md
da-kami: I re-worded it.

docs/research/user-experience/xmr-btc-user-interface-definition.md
da-kami: As far as I understand, both of them have an option. 
Bob has the option to NOT hand the `key_part` over that Alice needs for redeeming. Bob can then refund without Alice being able to stop him.
Alice has the option of NOT redeeming (after she received the `key_part`).

As far as I understand Alice is still in the more powerful position, because she has a longer time window to decide if she wants to redeem. Bob can shorten that time window by deciding when to hand the redeem key to her. But Bob has to be careful, because without Alice redeeming he is forced to refund before `t_2` has passed, otherwise he will be punished.

-- 
I was not planning to reason about this in detail in this document, but just state that the role is tied to the asset in this protocol. Do you think more reasoning is necessary?

da-kami: Additionally Alice is more powerful, because she can decide not to fund at all. Bob's funds would be locked until he can refund then. I reckon, that was the argument for saying it's better if the maker is in the role of Alice.

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -9,13 +9,16 @@ This document will be used as a basis for defining and evaluating low-fidelity U
 
 This section sums up assumptions around the current setup, protocol and existing software. 
 
-### Technical Limitations
+### Security Assumptions
 
-Swap-solutions that run completely in the browser (webpage) can be problematic because blockchain nodes, notably [bitcoind](https://github.com/bitcoin/bitcoin/issues/11833), do not specifying [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) in their response.
+Swap-solutions that run completely in the browser (webpage) and rely on connecting to a locally running blockchain node are currently not possible because specifying [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) is a security risk.
 This means the browser will reject a response from a locally running Bitcoin node.
+Discussions have been ongoing, notably for [bitcoind](https://github.com/bitcoin/bitcoin/issues/11833), to ad features that would allow webpages to communicate with locally running blockchain nodes directly.
```
D4nte, line 16:
```suggestion
Discussions have been ongoing, notably for [bitcoind](https://github.com/bitcoin/bitcoin/issues/11833), to add features that would allow webpages to communicate with locally running blockchain nodes directly.
```

D4nte (approved):

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,154 @@
...
+We think that, for a maker it is favourable to be in the role of Alice, because in Alice is in the powerful position of holding the secret. Additionally Bob funds first in the POC.
+However, to be able to allow a product that offers XRM->BTC and BTC->XMR swaps, we decided to let the maker and taker be in either role.
+The UX of the swap might be different depending on the role a party takes, especially when depicting the swap steps.
+
+### Additional Assumptions
+
+The first version of a prototype will not include hardware wallet (e.g. Ledger Nano S) support.
+In previous products we focused on swaps directly from and to cold storage, however, this does not concern a large enough user base to support it in the initial version. 
+Furthermore, hardware wallets add certain restrictions (forced user-interaction to sign transactions) that might conflict with the current prototol design.
+Hardware wallet support can be added at a later point if requested as feature by the community.
```
da-kami, line 67:
```suggestion
Using the wallets of the standard implementations (monerod and bitcoind) is the easiest, most straightforward starting point.
Hardware wallet support can be added at a later point if requested as feature by the community.
```

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,155 @@
...
+### Webpage
+
+Allowing webpages access to a locally running 
+A complete webpage based solution is not easily possible because of the CORS header restriction of browsers.
+We could enable communication between blockchain nodes and the browser through e.g. a proxy, but such a solution would not have an advantage over running a desktop application, because either way one has to run additional software next to the blockchain nodes to enable the swap.
+
+### Browser Extension
+
+Browser extensions, are not as restrictive as webpages as can be seen in the [Google Chrome's documentsion](https://developer.chrome.com/extensions/xhr).
+The code of a browser extension can be signed and is easily verifiable.
```
bonomat, line 101: Regarding the _easily verifiable_: I'd be careful with this in the sense of it being more secure than _X_: it might be true that initially a user could/would verify the build, but subsequent updates can easily be sneaked in. For example, who of you verified metamask or even noticed that it updated to a new version? 

I like the idea of the browser extension but I think _easily verifiable_ is not justified here as it sounds that it would be better (safer) than other solutions. 

One example I can think of which sounds safer to me would be a static website hosted on github pages where you can verify the exact code which is run. 

This doesn't solve the CORS issue I know. Maybe it's better to ignore this for now and not try to provide a solution which circumvents it.

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,155 @@
...
+
+A stand-alone desktop applications is possible, but can be seen as a hurdle, because every user first has to download the binary to see any form of user interface.
+For complex applications that involve a decentralized orderbook a desktop application might be advisable for the moment.
+
+For previous MVPs we used [Electron](https://www.electronjs.org/) to implement cross-platform desktop applications. 
+
+## Use-case specific prototypes
+
+Given the assumptions, technical limitations and resulting possible solutions described above this section specifies details for the prototypes described in the [initial project description](https://github.com/coblox/spikes/blob/master/0003-xmr-btc-product.md).
+In the initial project description we define two products to be be mocked as prototypes, (A) single market maker and (B) peer-to-peer decentralized trading platform.
```
bonomat, line 128:
```suggestion
In the initial project description we define two products to be mocked as prototypes, (A) single market maker and (B) peer-to-peer decentralized trading platform. 
```

docs/research/user-experience/xmr-btc-user-interface-definition.md
```diff
@@ -0,0 +1,155 @@
...
+Note, that the use-cases of a single market maker product and a peer-to-peer decentralized trading platform are agnostic to the user interface.
+Both products can be realised as CLI or UI taken into account the assumption listed above.
+
+The "product specific prototypes" focus on the setup and interaction of different software. 
+When it comes to prototyping specific user interface details, like for example the swap execution steps, we may opt for creating separate detail-mocks focussing on these details.
+
+### A. Single Market Maker product
+
+The point of this product is to keep `finding a trading partner` (discovery) and `agreeing on a trade` (negotiation) simple by only having a single maker. 
+A single-maker runs a webpage that gives a price. A taker can "take it of leave it".
```
bonomat, line 139:
```suggestion
A single-maker runs a webpage that gives a price. A taker can "take it or leave it".
```

docs/research/user-experience/xmr-btc-user-interface-definition.md
da-kami: Fair point! @D4nte and me came to the conclusion that the extension would be more verifiable than a webpage, because you can download the extension before install and compare checksums of the installer, whereas for the webpage that's hard to achieve.

But your example using GitHub pages is interesting, did not think about that one. 

In any case I'm happy to spare us this discussion for now and remove the sentence :)

da-kami:
```suggestion
```

D4nte: > Regarding the easily verifiable: I'd be careful with this in the sense of it being more secure than X: it might be true that initially a user could/would verify the build, but subsequent updates can easily be sneaked in. For example, who of you verified metamask or even noticed that it updated to a new version?

That is exactly it. The user does not need to verify the signature for an extension, Firefox enforces extension to be sign by them. You have to go through a process to get them to sign it [1]
which includes their review of the code source. I do not know if we will be able to go through this process but if we are, then it's _easily verifiable_ by the user: if Firefox installed it in then it was signed. I would expect firefox to not allow another developer to submit an update for our add-on, meaning they are protected against malicious updates. It also mean the verification is compulsory. The user needs to go through extra steps (change firefox config) to install a non-signed add-on.

> I like the idea of the browser extension but I think easily verifiable is not justified here as it sounds that it would be better (safer) than other solutions.

In any other solution (GUI, CLI) , the user would need to build themselves from source or verify a signed binary, which is less easily verifiable (still easy).

> One example I can think of which sounds safer to me would be a static website hosted on github pages where you can verify the exact code which is run.

While this is an interesting proposal, I am not convinced that a static website can do REST API calls to a local daemon to drive a swap and show a status. It's _static_ by definition.

[1] https://extensionworkshop.com/documentation/publish/signing-and-distribution-overview/

bonomat: Thanks for the clarification. I'm keen on learning more about the verification process of browser plugins. 

> While this is an interesting proposal, I am not convinced that a static website can do REST API calls to a local daemon to drive a swap and show a status. It's static by definition.

_Static_ was maybe the wrong term? Anyway, it's possible that you host a _static_ website on Github pages which has some JS/WASM libraries to do REST calls. I did a PoC using rust to wasm some time ago where I load GPX files from Github and render the path on an OSM. 

Code is terrible but it worked: https://github.com/bonomat/trackerz

docs/research/user-experience/xmr-btc-user-interface-definition.md
D4nte: Sorry was not clear enough: at this stage I would simply remove these questions/not consider the probabilistic option. If we reach the stage where people (ab)use it then we can look at solutions to mitigate it.

PR #43, opened 2020-11-19
Collect questions that we would like to answer

da-kami:
So far I just noted down some question. Please add questions if I missed something.

Should add more structure? 
At the moment split by role (taker, maker) and loosely grouped into topics (discovery, ...).
Should we rank by importance? (As in "what do we consider to be more important to answer")

Once we agree on the question set I would start mapping the questions on prototypes.
Potentially change/design prototypes to target specific questions more clearly.
(The first bunch of single-maker spot prototypes will be sent out without taking this into account.)

PR #45, opened 2020-11-24
Collect validation questions and map them to prototypes

da-kami:
Unfortunately the other PR went against the wrong base, which was not changed automatically because there were changes on the other branch in between.
Now properly cleaned up - nothing changed compared to #43 - I will just merge this one once it builds :)


D4nte:
Try to use your fork next time :)

PR #46, opened 2020-11-25
Adding peers manually could enhance privacy

da-kami:
This opens up different use cases and might preserve more privacy.
Validate if it is important for users to be able to control who to add (and how).


D4nte:
I think it's more something about "connecting to specific peer" vs "auto-discovery".
Also, it is not clear what "adding a peer" mean.
Finally, as we did not yet define the negotiation protocol then we don't know whether or not adding a peer manually helps with privacy.

Maybe just have a like of "discover peer automatically" vs "reach to peers manually".

rishflab (approved):

PR #51, opened 2020-12-01
Adapt the CLI, replace execution with refactored one

da-kami:
Still to be validated, that what happens in the `e2e` test can just be used like this in the actual execution triggered through the `cli`. Verify that swap can just be `await`ed.


rishflab:
Looks good so far

swap/src/cli.rs
```diff
@@ -1,38 +1,74 @@
...
-        monerod_url: Url,
+        #[structopt(
+            short = "m",
+            long = "monerod",
+            default_value = "http://127.0.0.1:18083/json_rpc"
+        )]
+        monero_wallet_rpc_url: Url,
 
-        #[structopt(long = "tor")]
+        #[structopt(short = "t", long = "tor")]
```
D4nte, line 64: Does tor even work? I don't think it does.

swap/src/cli.rs
```diff
@@ -1,38 +1,74 @@
...
+            default_value = "http://127.0.0.1:8332"
+        )]
         bitcoind_url: Url,
 
-        #[structopt(default_value = "http://127.0.0.1:18083/json_rpc", long = "monerod")]
-        monerod_url: Url,
+        #[structopt(
+            short = "m",
+            long = "monero-wallet-rpc",
+            default_value = "http://127.0.0.1:18083/json_rpc"
```
D4nte, line 20: Does a default value make sense? From our discussion looks like not.

swap/src/cli.rs
```diff
@@ -1,38 +1,74 @@
-use libp2p::core::Multiaddr;
+use libp2p::{core::Multiaddr, PeerId};
+use std::str::FromStr;
 use url::Url;
 use uuid::Uuid;
 
 #[derive(structopt::StructOpt, Debug)]
 #[structopt(name = "xmr-btc-swap", about = "Trustless XMR BTC swaps")]
 pub enum Options {
     Alice {
```
D4nte, line 9: Does this means the user needs to enter `swap alice`?
IMO, alice/bob roles are an implementation detail. Because There is only one possible direction given a role, then the role should be deduced from whether the user is sending or receiving sats (recip. piconeros)

D4nte (changes requested):
The role should not be exposed to the user.

swap/src/cli.rs
da-kami: Yes, it does. This is the wallet-rpc andpoint, that typically runs at `18083`, but there is no official default port defined.

swap/src/cli.rs
da-kami: I don't care at the moment.

swap/src/cli.rs
da-kami: yes, at the moment I kept that because it would be hard to remove it. My objective is to achieve a mainnet swap to see what other obstacles we run into. Once that is done I am happy to refactor and make it more usable.

da-kami: I totally agree that we should change this to imply the role, similar to what is depicted in the mocks. I would just get it into a state where we do a swap first and then iterate on it :)

D4nte (approved):
As discussed, UX does not matter right now, we want to do mainnet and recovery.
We can polish the UX once we decide on the first product to build.

da-kami:
Closing this in favor of https://github.com/comit-network/xmr-btc-swap/pull/61

PR #55, opened 2020-12-02
Remove generator code

D4nte:


D4nte:
Will do progressively as we replace the tests

PR #54, opened 2020-12-02
Punish Test

rishflab:
Add unhappy path test for `alice_punishes` scenario. There is a todo to assert on balances as the rest of team needs the changes to the executor function asap and I am stuck on trying to get Alice's bitcoin balance to update. Alice successfully exit on the Punish state however.

Added functionality to allow the execution to exit and restart at a specified state to allow for such tests. The public API remains unchanged.

PR #61, opened 2020-12-04
Prepare CLI for mainnet swap

da-kami:


swap/src/alice/swap.rs
```diff
@@ -102,6 +103,7 @@ pub async fn swap(
             s_a,
             v_a,
         } => {
+            info!("Alice State: started");
```
rishflab, line 106: These log statements can probably go once the punish test PR is merged

rishflab (approved):

swap/src/alice/swap.rs
da-kami: yeah, we can potentially drop those. It would be nice to print a bit more information than in the publish tests though. I see them as output that should help a user to understand what already happened.

swap/src/bin/swap.rs
```diff
@@ -36,21 +43,33 @@ extern crate prettytable;
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    init_tracing(LevelFilter::Trace).expect("initialize tracing");
```
da-kami, line 46: @rishflab @D4nte I am using tracing for user output at the moment. I wonder if that is what we want, or if we should distinguish between user-output (`println`) and logging (`tracing`). Strong opinions anyone?

xmr-btc/src/bitcoin.rs
```diff
@@ -212,6 +219,11 @@ pub trait GetRawTransaction {
     async fn get_raw_transaction(&self, txid: Txid) -> Result<Transaction>;
 }
 
+#[async_trait]
+pub trait Network {
```
da-kami, line 223: @rishflab @D4nte Like all the other traits that the wallet must fulfill I added a `Network` trait so the wallet is initialized for a network. But at the moment there is no validation for this. 
Additionally, we added `Config` that allows specifying a network so the boundaries when to act depend on the network. Should these concepts be combined?

da-kami:
Will rebase after https://github.com/comit-network/xmr-btc-swap/pull/54 was merged and then adapt/remove the log prints.

swap/src/bin/swap.rs
thomaseizinger: A good way to think about stdout vs stderr is that `stdout` should play nice with piping and hence be "machine-readable" vs `stderr` is for humans and hence can contain all the diagnostic information.

If your CLI doesn't have anything that it could feed to another CLI, probably everything should go on stderr :)

xmr-btc/src/bitcoin.rs
D4nte: I guess they should be combined yes, is `Wallet` the best place to store the network?

da-kami:
Resolve merge conflict, would go for merge after that and then add the refactoring of removing the roles from the `cli` on top of that.

xmr-btc/src/bitcoin.rs
da-kami: The wallet still has to be initialized with `Network`, because the `xmr_btc` crate requires the network (when constructing `TxLock` - only the wallet is available). 

for now I added the `bitcoin_network` to the `Config` and always just pass that network (always use the `Config`. I think that is fine for now, eventually we can come up with a better solution.

da-kami:
Rebased, cleaned up and ready for review again please! Let's get this in.

rishflab (approved):
Looks good to me!

PR #65, opened 2020-12-09
Drive swarm in separate async task

rishflab:
Previously the libp2p swarm had to be manually polled within the
protocol execution code to execute actions such as sending a
message. The swarm is now wrapped in SwarmDriver which polls the
swarm in a seperate task

### DIFF
--- a/swap/src/alice/execution.rs
+++ b/swap/src/alice/execution.rs
@@ -1,6 +1,6 @@
 use crate::{
     alice::swarm_driver::SwarmDriverHandle, bitcoin, monero, network::request_response::AliceToBob,
-    SwapAmounts, PUNISH_TIMELOCK, REFUND_TIMELOCK,
+    SwapAmounts,
 };
 use anyhow::{bail, Context, Result};
 use ecdsa_fun::{adaptor::Adaptor, nonce::Deterministic};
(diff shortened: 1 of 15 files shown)


swap/src/alice/execution.rs
```diff
@@ -88,40 +70,21 @@ pub async fn negotiate(
     );
 
     // TODO(Franck): Understand why this is needed.
-    swarm.set_state0(state0.clone());
+    // swarm.swarm.set_state0(state0.clone());
```
da-kami, line 73: So, is this not needed anymore?

da-kami, line 73:
```suggestion
```

swap/src/alice/execution.rs
```diff
@@ -88,40 +70,21 @@ pub async fn negotiate(
     );
 
     // TODO(Franck): Understand why this is needed.
```
da-kami, line 72:
```suggestion
```

da-kami:
Should this PR be based on `bob-unhappy-paths` instead of `master`? I feel some changes overlap with that branch.

D4nte (approved):
Looks good, let's merge that first by itself. We can merge #71 after

PR #72, opened 2020-12-10
Add refund test and complete punish test

rishflab:
Added a test for the scenario when both Alice and Bob refund

Added assertions to check balances are correct after the Alice punish test


xmr-btc/src/bob.rs
```diff
@@ -799,6 +798,68 @@ impl State4 {
...
+        let t2 = t1 + self.punish_timelock;
+
+        match (current_block_height < t1, current_block_height < t2) {
+            (true, _) => Ok(Epoch::T0),
+            (false, true) => Ok(Epoch::T1),
+            (false, false) => Ok(Epoch::T2),
+        }
+    }
+
+    pub async fn refund_btc<W: bitcoin::BroadcastSignedTransaction>(
```
da-kami, line 820: if this was removed somewhere else please remove the dead code!

da-kami (changes requested):

da-kami (approved):

PR #76, opened 2020-12-10
Allow returning XMR from monero::Amount as decimal-string

da-kami:
This allows us to implement `Display` on `monero::Amount` that shows `XMR`.
We use `rust_decimal` to convert from piconeros to decimal-string.
`rust_decimal` uses `i64` internally, so only values within `i64` can be returned by `as_monero()`.
For `Display` either the result of `as_monero()` is returned, or the piconeros if `as_monero()` errors.

In the long run we could add `Denomination` similar to [rust bitcoin](https://docs.rs/bitcoin/0.25.2/src/bitcoin/util/amount.rs.html#209-244), but that is quite complex to add so I decided to go with this first implementation of `Display`. It will only fail for very large amounts larger than `9223372.036854775807` XMR, which should be rare...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2889,6 +2889,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "rust_decimal"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


da-kami:
Reopening later

PR #81, opened 2020-12-14
Database & Resume Swap

rishflab:
Rebased @D4nte work on the database onto current master.
Update tests to use event-loop and properly restart.

Notable changes: 

* Refactored the e2e test to have re-usable building blocks to init containers, wallets, and initial state
* Refactored the network behavior to be stateless (we finally did that 🎉) so we can overcome re-starting issues where the behaviour would need an initial state that is not available
* Add `[role]::recover` function next to `[role]::swap`, (used in tests, to be plugged into CLI so we can recover :)


da-kami:
Database and recovery done :)
Will adapt the CLI in the morning as small follow-up PR, the `recover` functions are already in place and just have to be called with the correct params which includes the additional param to pass a db dir in. 

For completion we should add balance checks to the `happy_path_restart` tests. @rishflab would be cool if you could tackle that in the morning :)

da-kami:
Let's not forget to squash the earlier commits on this branch 😬

da-kami:
The tests pass locally when run separately. When running in CI there was a failure when running the tests that are in same file. Could be an initialization issue with the containers, but needs more investigation. I split the tests into multiple files for now to see if that resolves the issue.

thomaseizinger:
Yeah you can't have them listen on the same port :)
I would suggest to use the `MemoryTransport` like we did in `comit-rs` to not run into this issue.
Alternatively you can listen on TCP port 0 to get one automatically assigned by the OS but that is a bit tricky because you have to get it out again of the `NewListenAddress` event before you can dial.

Note: I did not read any of the code :)

swap/src/alice/negotiate.rs
```diff
@@ -10,6 +10,7 @@ use futures::{
 };
 use libp2p::request_response::ResponseChannel;
```
D4nte, line 11: Why the file name change? this does more than just negotiation.

swap/src/bin/swap.rs
```diff
@@ -214,24 +222,7 @@ async fn main() -> Result<()> {
...
-                bitcoind_url,
-                config.bitcoin_network,
-            )
-            .await
-            .expect("failed to create bitcoin wallet");
-            let monero_wallet = monero::Wallet::new(monerod_url);
-
-            recover(bitcoin_wallet, monero_wallet, state).await?;
-        }
+        Options::Recover { .. } => todo!("implement this"),
```
D4nte, line 225: What is `recover` for? Do you mean, `resume`?

D4nte (commented):
I don't think "recovery" is a good terminology because if possible, the swap will continue. I suggest "Resume" instead.

Also, my work was on https://github.com/comit-network/xmr-btc-swap/pull/83 I will see if I can combine our changes.

Can you please clean-up your commit tree?

swap/src/alice/negotiate.rs
da-kami: Does it? I think negotiate is more accurate than execute. We can rename it again, but I think it is not the execution.

swap/src/bin/swap.rs
da-kami: This is the terminology that was previously used in the project. I am happy to rename both the CLI option and the function to `resume`.

D4nte: The terminology previously used represented a different feature.

swap/tests/refund_restart.rs
```diff
@@ -0,0 +1,162 @@
...
+use testutils::init_tracing;
+use uuid::Uuid;
+use xmr_btc::{bitcoin, config::Config};
+
+pub mod testutils;
+
+// Bob locks btc and Alice locks xmr. Alice fails to act so Bob refunds. Alice
+// then also refunds.
+#[tokio::test]
+async fn both_refund() {
```
D4nte, line 17: This only tests for Alice to restart and refund. We need a test for Bob to restart to ensure that database save/restore works as expected.

swap/src/alice/negotiate.rs
D4nte: It provides an abstraction that allows to do one function call per step for the protocol execution.

Look at the functions present in the file, they are not related to "negotiation":

- `wait_for_locked_bitcoin`
- `lock_xmr`
- etc

would `steps.rs` make more sense to you?

swap/src/alice/swap.rs
```diff
@@ -393,67 +393,72 @@ pub async fn run_until(
...
-                                }
+                match state3.current_epoch(bitcoin_wallet.as_ref()).await? {
+                    Epoch::T0 => {
+                        let wait_for_enc_sig = wait_for_bitcoin_encrypted_signature(
+                            &mut event_loop_handle,
+                            config.monero_max_finality_time,
+                        );
+                        let t1_timeout = state3.wait_for_t1(bitcoin_wallet.as_ref());
+
+                        tokio::select! {
```
D4nte, line 423: In my try I replaced this with `futures::future::select`. It allows to return a value, the state, removing the triplicate db save and `run_until` call.

swap/src/bin/swap.rs
rishflab: my vote goes for resume

da-kami:
Closing this PR in favor of https://github.com/comit-network/xmr-btc-swap/pull/85

PR #85, opened 2020-12-15
Database & Swap Resume

D4nte:
- Introduce Database for Alice and Bob
- Save states in database
- Resume from database (broken for Bob, will create test + fix it in follow-up) PR
- Tests when alice restarts: Both happy and refund path


rishflab (approved):
Lets get this in and do clean up in later PRs

swap/src/alice.rs
```diff
@@ -185,3 +184,19 @@ impl Behaviour {
...
+            pt: PeerTracker::default(),
+            amounts: Amounts::default(),
+            message0: Message0::default(),
+            message1: Message1::default(),
+            message2: Message2::default(),
+            message3: Message3::default(),
+            identity,
+        }
+    }
+}
```
D4nte, line 202:
```suggestion
impl Default for Behaviour {
    fn default() -> Self {
        let identity = Keypair::generate_ed25519();

        Self {
            identity,
            ..Default::default()
        }
    }
}
```

Will do in follow-up PR

da-kami (approved):

PR #86, opened 2020-12-15
Database clean-up

D4nte:
Small clean-ups I spotted when reviewing #85.

### DIFF
--- a/swap/src/alice/swap.rs
+++ b/swap/src/alice/swap.rs
@@ -19,7 +19,6 @@ use crate::{
     SwapAmounts,
 };
 use anyhow::{bail, Result};
-use async_recursion::async_recursion;
 use futures::{
     future::{select, Either},
     pin_mut,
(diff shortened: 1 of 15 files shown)


da-kami (approved):

rishflab (approved):

swap/tests/happy_path.rs
```diff
@@ -94,8 +92,8 @@ async fn happy_path() {
         Uuid::new_v4(),
     );
 
-    let _bob_swarm_fut = tokio::spawn(async move { bob_event_loop.run().await });
-
+    tokio::spawn(async move { alice_event_loop.run().await });
+    tokio::spawn(async move { bob_event_loop.run().await });
```
da-kami, line 96: Could the stackoverflow problem be related to removing the futures being kept in scope with `_` variables?
I don't think it should, but that is the only notable change I can see that could cause this.

D4nte: The  `JoinHandle` returned by `tokio::spawn` is only meant to be used to await on the stack if one wishes to do so.

xmr-btc/src/bob.rs
```diff
@@ -373,6 +371,8 @@ impl State0 {
         let s_b = cross_curve_dleq::Scalar::random(rng);
         let v_b = monero::PrivateViewKey::new_random(rng);
 
+        let dleq_proof_s_b = cross_curve_dleq::Proof::new(rng, &s_b);
```
D4nte, line 374: @luckysori what do you think? Do you see any downsize in doing so?

luckysori: Nope, seems okay!

By the way, I would still recommend replacing this proof with Lloyd's version of it in `secp256kfun`. It's faster, smaller and more likely to be correct. Just a reminder.

D4nte: Could you open an issue to describe in a bit more details please?

luckysori: Is it okay if instead I open a simple issue and I do it quickly once you think the codebase is a bit more stable?

I can write a detailed explanation of what to do, but it feels like a waste of time.

luckysori: https://github.com/comit-network/xmr-btc-swap/issues/90

D4nte: > Is it okay if instead I open a simple issue and I do it quickly once you think the codebase is a bit more stable?
> 
> I can write a detailed explanation of what to do, but it feels like a waste of time.

Sounds great, thanks!

swap/src/alice/swap.rs
```diff
@@ -502,22 +502,22 @@ impl Swap {
                         let refund_tx_seen =
                             bitcoin_wallet_clone.watch_for_raw_transaction(tx_refund.txid());
 
-                        pin_mut!(punish_tx_finalised);
-                        pin_mut!(refund_tx_seen);
-
-                        state = match select(punish_tx_finalised, refund_tx_seen).await {
-                            Either::Left(_) => AliceState::Punished,
-                            Either::Right((published_refund_tx, _)) => {
+                        select! {
```
da-kami, line 505: You can actually return the state from `select!` and assign it to the variable outside. 
Did you implement it with state being set inside on purpose?

da-kami:
Note that there is another `pin_mut` in `steps.rs` that can also be replaced with `select!` - I tried that, but am still running into the `stackoverlfow` issue.

swap/src/alice/swap.rs
D4nte: I was convinced of the opposite. Thanks!

D4nte:
> Note that there is another `pin_mut` in `steps.rs` that can also be replaced with `select!` - I tried that, but am still running into the `stackoverlfow` issue.

I intend to check type sizes using `-Zprint-type-sizes`

D4nte:
Close for now due to test failure

PR #87, opened 2020-12-15
Reorganise modules

rishflab:
Our previous structure did not make sense. We had a state.rs which contained the
storage models at the top level. The actual protocol state was a level lower in
swap::alice and swap::bob. We had protocol code (negotiate.rs/steps.rs) in the
same module as our stateless network code. Created 3 modules for protocol,
network and storage and put the files where they should belong.


da-kami:
I understand the changes and don't really have a problem with them, but could you maybe describe the motivation for this a bit more? Why do we need this?

rishflab:
> I understand the changes and don't really have a problem with them, but could you maybe describe the motivation for this a bit more? Why do we need this?

Our previous structure did not make sense. We had a state.rs which contained the storage models at the top level. The actual protocol state was a level lower in swap::alice and swap::bob. We had protocol code in the same module as our stateless network code. I created 3 modules for protocol, network and storage and put the files where the should belong.

Will add this explanation to the PR description/commit.

D4nte (changes requested):
Can we avoid re-organisation when we are sprinting towards an end goal? (mainnet).

rishflab:
Closing this as it is too tricky to rebase it. Easier to remake the PR.

PR #89, opened 2020-12-15
Bugfix: Remove fragile assertions

rishflab:
The the Alice and Bob restart tests we were asserting that both
Alice's and Bob's balances were correct. This was problematic
because we are not waiting for both parties to complete causing
the test to fail (sometimes). This issue is resolved by removing
the assertions for the party that is moved into an async task as
we do not what state they have reached when we check their
balances.

I am not sure whether we should merge this commit or change the tests so that we wait (join) until both parties have hit redeemed?

### DIFF
--- a/swap/tests/happy_path_restart_alice.rs
+++ b/swap/tests/happy_path_restart_alice.rs
@@ -66,11 +66,6 @@ async fn given_alice_restarts_after_encsig_is_learned_resume_swap() {
         )
         .await;
 
-    // TODO: we are making a clone of Bob's wallets here to keep them in scope after
-    // Bob's wallets are moved into an async task.
-    let bob_btc_wallet_clone = bob_btc_wallet.clone();
-    let bob_xmr_wallet_clone = bob_xmr_wallet.clone();
-
     let _ = tokio::spawn(async move {
         bob::swap::swap(
             bob_state,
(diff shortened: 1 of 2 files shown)


D4nte:
> I am not sure whether we should merge this commit or change the tests so that we wait (join) until both parties have hit redeemed?

I believe we should wait until both party have redeemed.

rishflab:
> > I am not sure whether we should merge this commit or change the tests so that we wait (join) until both parties have hit redeemed?
> 
> I believe we should wait until both party have redeemed.

In that case does it make sense to merge the Alice and bob tests into one. We can abort and resume from db for both parties?

da-kami:
> > I am not sure whether we should merge this commit or change the tests so that we wait (join) until both parties have hit redeemed?
> 
> I believe we should wait until both party have redeemed.

I agree with that we should do that, it is not trivial though. 

I am taking Alice restarts (and Bob does not) as an example:

We initialize Alice's eventloop and handle.
We initialize Bob's eventloop and handle.
We spawn away Bob's swap and eventloop.
We swapn away Alice's eventloop.
We use `run_until` to `await` Alice's specific restart state we want.
We init another Alice eventloop.
We spawn away Alice new eventloop and await her `swap` with a specific state.

In this setup we cannot know when Bob is finished.

For happy-path we solve this problem by using a `try_join` for both to make sure both resolve before asserting. But that does not work here because of using `run_until` and the interruption. 

@D4nte would you have a solution in mind how to solve that problem?

da-kami:
> > > I am not sure whether we should merge this commit or change the tests so that we wait (join) until both parties have hit redeemed?
> > 
> > 
> > I believe we should wait until both party have redeemed.
> 
> In that case does it make sense to merge the Alice and Bob tests into one. We can stop and restart both parties in the same test. Do we gain anything by having seperate tests?

Wouldn't it be enough to move the assertion for Bob into the async block where we `await` Bob's future?

D4nte:
> Do we gain anything by having seperate tests?

Just focusing the test on one functionality at a time.

PR #91, opened 2020-12-17
Remove breaks and continues

rishflab:
Optional Change. 

Replaces the breaks and continues in run_until with a returns. There is one function to transition the state (`next_state`) and another to run the state machine. The state transition function no longer has to have knowledge of how it is being called.

### DIFF
--- a/swap/src/alice/swap.rs
+++ b/swap/src/alice/swap.rs
@@ -19,7 +19,6 @@ use crate::{
     SwapAmounts,
 };
 use anyhow::{bail, Result};
-use async_recursion::async_recursion;
 use futures::{
     future::{select, Either},
     pin_mut,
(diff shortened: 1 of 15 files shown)


swap/src/alice/swap.rs
```diff
@@ -253,288 +253,292 @@ impl Swap {
...
+        mut self,
+        state: AliceState,
+        is_target_state: fn(&AliceState) -> bool,
+    ) -> Result<AliceState> {
+        let mut result = state;
+        loop {
+            info!("Current state: {}", result);
+            result = self.next_state(result).await?;
+            if is_target_state(&result) {
+                return Ok(result);
```
D4nte, line 541:
```
	break Ok(result);
```

`break` should be used to return a value from a loop, not `return`: https://doc.rust-lang.org/rust-by-example/flow_control/loop/return.html

D4nte (commented):
Probably need to park this one until I fixed thee branch

PR #94, opened 2020-12-17
T1Expired state for Bob and check expiries in states after both locked

da-kami:


swap/src/bob/swap.rs
```diff
@@ -359,6 +364,31 @@ where
                 )
                 .await
             }
+            BobState::T1Expired(state4) => {
```
luckysori, line 367: I haven't looked too deeply into the state of the codebase now, so this may already be accounted for.

There is a difference between T1 expiring (I assume by this you mean the cancel transaction being publishable) before or after Bob has given Alice his encrypted signature on `bitcoin::TxRedeem`:

- If Bob has not given Alice his encrypted signature yet, then Bob only has to make sure that he gets `TxCancel` and `TxRefund` in before Alice can punish him.
- If Bob _has_ given Alice his encrypted signature already, then there is a race between Alice redeeming the bitcoin and Bob publishing `TxCancel`. It's obviously not in Alice's interest to risk this as she can end up revealing `s_a` via the mempool and not getting the bitcoin, but there's still a chance that after you reach `BobState::T1Expired` Alice redeems.

Tl;dr: I think you should be selecting between cancel and redeem futures and if you find the redeem transaction, you should allow Bob to redeem the monero.

da-kami: Thanks @luckysori, that is an interesting edge case!

I will write down the solution in my own words, please let me know if I misunderstood something.

Case: Bob has already sent `encsig` to Alice AND `t1` expires

Bob has to simultanously try to send `tx_cancel` and watch for `tx_redeem`. If `tx_cancel` is confirmed Bob goes for `refund`. If the BTC `redeem` tx sent by Alice is obseved by Bob, Bob has to redeem the Monero.

I will see to add this tomorrow.

swap/src/bob/swap.rs
```diff
@@ -272,18 +276,27 @@ where
...
+                    // Bob sends Alice his key
+                    let tx_redeem_encsig = state.tx_redeem_encsig();
+
+                    let state4_clone = state.clone();
+                    let enc_sig_sent_watcher =
+                        event_loop_handle.send_message3(alice_peer_id.clone(), tx_redeem_encsig);
+                    let bitcoin_wallet = bitcoin_wallet.clone();
+                    let t1_timeout = state4_clone.wait_for_t1(bitcoin_wallet.as_ref());
+
+                    pin_mut!(enc_sig_sent_watcher);
```
D4nte, line 290: Let's prefer `select!` over `pin_mut`.

swap/src/bob/swap.rs
```diff
@@ -300,28 +313,20 @@ where
...
-                        BobState::Cancelled(state)
+                let state = if let Epoch::T0 = state.current_epoch(bitcoin_wallet.as_ref()).await? {
+                    let state_clone = state.clone();
+                    let redeem_watcher = state_clone.watch_for_redeem_btc(bitcoin_wallet.as_ref());
+                    let t1_timeout = state_clone.wait_for_t1(bitcoin_wallet.as_ref());
+
+                    pin_mut!(redeem_watcher);
+                    pin_mut!(t1_timeout);
+
+                    match select(redeem_watcher, t1_timeout).await {
```
D4nte, line 324: use `select!`

swap/src/bob/swap.rs
D4nte: >  T1 expiring (I assume by this you mean the cancel transaction being publishable) 

Can you clarify why you have to assume that? This is core of the protocol and I am worried when I read that "t1 expired" is not obviously equivalent to "cancel tx can be published". What edge case am I missing?

D4nte: A proper solution to this edge case is properly calculate safe expiries. Something we can do at a later stage.

swap/src/bob/swap.rs
```diff
@@ -359,6 +364,31 @@ where
                 )
                 .await
             }
+            BobState::T1Expired(state4) => {
+                if state4
+                    .check_for_tx_cancel(bitcoin_wallet.as_ref())
+                    .await
+                    .is_err()
+                {
+                    state4.submit_tx_cancel(bitcoin_wallet.as_ref()).await?;
```
D4nte, line 373: Once #97 is done, then we should just try to broadcast, no need to verify beforehand.

swap/src/state.rs
```diff
@@ -106,9 +107,10 @@ impl Display for Bob {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Bob::Negotiated { .. } => f.write_str("Handshake complete"),
-            Bob::BtcLocked { .. } | Bob::XmrLocked { .. } | Bob::BtcCancelled(_) => {
-                f.write_str("Bitcoin refundable")
-            }
+            Bob::BtcLocked { .. }
+            | Bob::XmrLocked { .. }
+            | Bob::T1Expired(_)
+            | Bob::BtcCancelled(_) => f.write_str("Bitcoin refundable"),
```
D4nte, line 113: Not sure why we started to combine these states in the first place.
Please have a display implementation per state. The discrepancy can only lead to confusion.

also fix `BtcRedeemed` while you are there :)

D4nte (approved):

swap/src/bob/swap.rs
da-kami: I think what @luckysori wanted to point out is the fact that it can be a race here - even if Bob does not want that he cannot stop it. 
If Bob has sent the `encsig` to Alice but Alice does not redeem until `t1` has passed Bob will try to get the `cancel` tx in. But Alice might race him with the `redeem` transaction (which would make `cancel` invalid).
If we assume that both Alice and Bob act according to our implementation I don't think we will run into this problem at the moment, but eventually we might have to make Bob's side more resilient against this problem, so he still goes on with `redeem` the XMR in such a case and does not get "stuck" in trying to cancel. 

Shall we just create a ticket for this edge-case for now?

swap/src/bob/swap.rs
da-kami: OK, I will change ALL of the `pin_mut` to `select` then.
I created this PR under the assumption that the cleanup would rebase on these changes and then change this. Did not want to mix it... :D

swap/src/bob/swap.rs
da-kami: That is true #97 actually branched off this branch so the functionality is added on both sides :)

swap/src/bob/swap.rs
luckysori: > > T1 expiring (I assume by this you mean the cancel transaction being publishable)
> 
> Can you clarify why you have to assume that? This is core of the protocol and I am worried when I read that "t1 expired" is not obviously equivalent to "cancel tx can be published". What edge case am I missing?

I only say assume because of the context i.e. the rest of the code in this PR.

Intuitively (to me) `t0` would actually be the cancel transaction timelock and `t1` would be the punish transaction timelock, but I think you are modelling it differently. I just haven't looked at the codebase deeply enough to understand the new model.

D4nte: First time I here of `t0` :) I always heard of `t1` and `t2`.

swap/src/bob/swap.rs
D4nte: Don't over do it, just don't introduce new `pin_mut`.

swap/src/state.rs
da-kami: Thanks! I should have actioned that one right away

swap/src/bob/swap.rs
luckysori: > A proper solution to this edge case is properly calculate safe expiries. Something we can do at a later stage.

Can you explain how safe expiries help here? I think no matter how safe your expiries are, Alice can always publish the redeem transaction (if she has Bob's encrypted signature) to compete with a cancel transaction. It's just that it's generally not in her interest, but the code should eventually handle the race anyway.

luckysori: > First time I here of `t0` :) I always heard of `t1` and `t2`.

Haha, there goes my confusion then.

D4nte (approved):
Maybe one last comment, if `T1` can be confusing, we could rename to `CancelTimelock`

PR #97, opened 2020-12-17
Resilient cancel transaction error handling

da-kami:
This should be merged after https://github.com/coblox/bitcoin-harness-rs/pull/19 was merged; and we should properly point `bitcoin_harness` to a release or master.

I am still keeping this as Draft until https://github.com/coblox/bitcoin-harness-rs/pull/19 and we fixed the dependency.

### DIFF
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -13,13 +13,12 @@ atty = "0.2"
 backoff = { version = "0.2", features = ["tokio"] }
 base64 = "0.12"
 bitcoin = { version = "0.25", features = ["rand", "use-serde"] }
-bitcoin-harness = { git = "https://github.com/coblox/bitcoin-harness-rs", rev = "864b55fcba2e770105f135781dd2e3002c503d12" }
+bitcoin-harness = { git = "https://github.com/coblox/bitcoin-harness-rs", rev = "f3fbbba403e1d3329d8fd79d7549775183bcbc74" }
 conquer-once = "0.3"
 derivative = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", rev = "cdfbc766045ea678a41780919d6228dd5acee3be", features = ["libsecp_compat", "serde"] }
 futures = { version = "0.3", default-features = false }
 genawaiter = "0.99.1"
-jsonrpc_client = { git = "https://github.com/thomaseizinger/rust-jsonrpc-client", rev = "c7010817e0f86ab24b3dc10d6bb0463faa0aace4", features = ["reqwest"] }
 libp2p = { version = "0.29", default-features = false, features = ["tcp-tokio", "yamux", "mplex", "dns", "noise", "request-response"] }
 libp2p-tokio-socks5 = "0.4"
 log = { version = "0.4", features = ["serde"] }
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -19,6 +19,7 @@ derivative = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", rev = "cdfbc766045ea678a41780919d6228dd5acee3be", features = ["libsecp_compat", "serde"] }
 futures = { version = "0.3", default-features = false }
 genawaiter = "0.99.1"
+jsonrpc_client = { git = "https://github.com/thomaseizinger/rust-jsonrpc-client", rev = "c7010817e0f86ab24b3dc10d6bb0463faa0aace4", features = ["reqwest"] }
 libp2p = { version = "0.29", default-features = false, features = ["tcp-tokio", "yamux", "mplex", "dns", "noise", "request-response"] }
 libp2p-tokio-socks5 = "0.4"
 log = { version = "0.4", features = ["serde"] }
--- a/swap/src/bob/swap.rs
+++ b/swap/src/bob/swap.rs
@@ -7,7 +7,6 @@ use crate::{
 };
 use anyhow::{bail, Result};
 use async_recursion::async_recursion;
-use libp2p::{core::Multiaddr, PeerId};
 use rand::{CryptoRng, RngCore};
 use std::{convert::TryFrom, fmt, sync::Arc};
 use tokio::select;
--- a/swap/src/bob/swap.rs
+++ b/swap/src/bob/swap.rs
@@ -5,7 +5,7 @@ use crate::{
     storage::Database,
     SwapAmounts, TRANSACTION_ALREADY_IN_BLOCKCHAIN_ERROR_CODE,
 };
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 use async_recursion::async_recursion;
 use bitcoin_harness::bitcoind_rpc::jsonrpc_client::JsonRpcError;
 use libp2p::{core::Multiaddr, PeerId};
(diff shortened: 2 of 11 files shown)


swap/Cargo.toml
```diff
@@ -19,6 +19,7 @@ derivative = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", rev = "cdfbc766045ea678a41780919d6228dd5acee3be", features = ["libsecp_compat", "serde"] }
 futures = { version = "0.3", default-features = false }
 genawaiter = "0.99.1"
+jsonrpc_client = { git = "https://github.com/thomaseizinger/rust-jsonrpc-client", rev = "c7010817e0f86ab24b3dc10d6bb0463faa0aace4", features = ["reqwest"] }
```
D4nte, line 22: We are bringing this dependency via `bitcoin-harness`. Having it twice is risking to have compatibility issues between version.

Why are you bringing the dependency in?

swap/src/bob/swap.rs
```diff
@@ -370,7 +370,18 @@ where
                     .await
                     .is_err()
                 {
-                    state4.submit_tx_cancel(bitcoin_wallet.as_ref()).await?;
+                    let result = state4.submit_tx_cancel(bitcoin_wallet.as_ref()).await;
+                    if let Err(error) = result {
+                        let json_rpc_err = error
+                            .downcast_ref::<jsonrpc_client::JsonRpcError>()
+                            .ok_or_else(|| anyhow!("Failed to downcast JsonRpcError"))?;
```
D4nte, line 377:
```suggestion
                            .context("Failed to broadcast cancel transaction")?;
```
If you are not able to downcast it means that something else went wrong. Maybe Bitcoind was not reachable and hence it is not a `JsonRpcError`.
In this case, we need to have a useful error message.

swap/src/bob/swap.rs
```diff
@@ -370,7 +370,18 @@ where
...
+
+                        if json_rpc_err.code == -27 {
+                            warn!("Failed to send cancel transaction, assuming that is was already included by the other party...");
```
D4nte, line 380: what is error code `-27`? Isn't it ONLY when bitcoind already knows the transaction?
In this case, just do an `info`.
A warn is something that the user should be warned about. The user does not need to be warned here, just informed. It all went well.
```suggestion
                            info!("Cancel Transaction is already published");
```

D4nte (changes requested):
Really not keen to bring json-rpc twice.

Just update json-rpc in bitcoin-harness first to do it properly.

swap/Cargo.toml
da-kami: Because it is the easiest way to get the type `JsonRpcError` in. I discussed this with @thomaseizinger yesterday. Can we solve this by re-exporting the type from `bitcoin_harness`? Happy to do so, but we can also just do it in a follow up.

swap/src/bob/swap.rs
da-kami: True, that has to be improved! I think it would be better to pass on the original error in this case instead of creating our own `anyhow` error.

swap/Cargo.toml
D4nte: ah, thanks for clarifying. I think it'd be safer to re-export the type than shooting ourselves in the foot with version conflicts. re-export the whole crate.

swap/src/bob/swap.rs
D4nte: You were not passing the original error in the first place tho.

swap/Cargo.toml
da-kami: Agreed! https://github.com/coblox/bitcoin-harness-rs/pull/19

swap/src/bob/swap.rs
da-kami: The level was changed to `info` but the message remains the same. Probably better to give a bit more information back to the user at this point. We can change this at a later point if it is confusing.

swap/src/bob/swap.rs
da-kami: Now we do :)

rishflab (approved):

da-kami:
@rishflab it seems that the logic is flawed somehow. It fails with:

```
 Caused by:
    0: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)
    1: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)', swap/tests/punish.rs:99:82
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

From this error my educated guess would be that we are trying to submit a `refund` but `cancel` was not confirmed yet.

rishflab:
> @rishflab it seems that the logic is flawed somehow. It fails with:
> 
> ```
>  Caused by:
>     0: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)
>     1: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)', swap/tests/punish.rs:99:82
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> From this error my educated guess would be that we are trying to submit a `refund` but `cancel` was not confirmed yet.

It wasn't happening earlier before the squash. Do you know what could of changed?

da-kami:
> > @rishflab it seems that the logic is flawed somehow. It fails with:
> > ```
> >  Caused by:
> >     0: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)
> >     1: JSON-RPC request failed with code -26: non-mandatory-script-verify-flag (Signature must be zero for failed CHECK(MULTI)SIG operation) (code 64)', swap/tests/punish.rs:99:82
> > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> > ```
> > 
> > 
> > From this error my educated guess would be that we are trying to submit a `refund` but `cancel` was not confirmed yet.
> 
> It wasn't happening earlier before the squash. Do you know what could of changed?

I think it is unrelated to the squash and could have happened before, because we are not properly checking that the transaction that the other party submitted is already confirmed.

My guess is: 

1. `tx_cancel` was sent by Alice
2. Bob's tx cancel submission fails (so he assumes that is it already confirmed),
3. Alice's tx cancel is not confirmed yet 
4. Bob submits `tx_refund` without `tx_cancel` being confirmed which fails. 

I think we need more elaborate handling. We might have to bring the previous check for the actual transaction back and ensure that it is confirmed as well (or wait for confirmation). This is especially important for mainnet where confirmation time can take significant time.

xmr-btc/src/bob.rs
```diff
@@ -660,51 +660,34 @@ impl State4 {
...
     where
         W: BroadcastSignedTransaction,
     {
-        let tx_cancel =
-            bitcoin::TxCancel::new(&self.tx_lock, self.refund_timelock, self.A, self.b.public());
+        let tx_cancel = self.tx_cancel();
 
-        let sig_a = self.tx_cancel_sig_a.clone();
-        let sig_b = self.b.sign(tx_cancel.digest());
+        let sig_theirs = self.tx_cancel_sig_a.clone();
```
da-kami, line 673:
```suggestion
        let sig_theirs = self.tx_cancel_sig_alice.clone();
```

❗️ This code is becoming more and more confusing because different terminologies are mixed together.
Above you refer to `tx_punish_sig_bob` at this point, should this be `tx_cancel_sig_alice`?
Please use the same terminology throughout to avoid confusion.

da-kami (commented):
I still think we should be able to handle the transaction failure inside a function that can be used by Alice and Bob, but it might require a bit of refactoring.

In the end both roles use the same signature to add the signatures to `tx_cancel`, just their inputs are different, so we should be able to extract that.

xmr-btc/src/bob.rs
rishflab: Removed theirs and hours

I can change to `tx_cancel_sig_alice`. I didn't make it `tx_cancel_sig_a`, it was already like that.

swap/src/bob/swap.rs
```diff
@@ -357,13 +358,18 @@ where
...
             }
             BobState::T1Expired(state4) => {
-                if state4
-                    .check_for_tx_cancel(bitcoin_wallet.as_ref())
-                    .await
-                    .is_err()
-                {
-                    state4.submit_tx_cancel(bitcoin_wallet.as_ref()).await?;
-                }
+                let result = state4.submit_tx_cancel(bitcoin_wallet.as_ref()).await;
```
D4nte, line 361: This should be done closer to the RPC call and we should avoid duplication this error handling code. I am preparing a patch to show how it can be done.

D4nte: https://github.com/comit-network/xmr-btc-swap/pull/103

D4nte (approved):

PR #100, opened 2020-12-18
 Bob refunds swap after restart that requires communication

D4nte:
As Bob is dialing Alice, we now ensure that we are connected to Alice
at each step that needs communication.
If we are not connected, we proceed with dialing.

In an attempt to improve libp2p usage, we also add known address of
Alice first and only use peer_id to dial.
This ensures that we use the expected peer id.

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -1,4 +1,4 @@
-use libp2p::core::Multiaddr;
+use libp2p::{core::Multiaddr, PeerId};
 use url::Url;
 use uuid::Uuid;
 
@@ -47,6 +47,9 @@ pub enum Command {
         receive_bitcoin: bitcoin::Amount,
     },
     BuyXmr {
+        #[structopt(short = "p", long = "connect-peer-id")]
+        alice_peer_id: PeerId,
+
         #[structopt(short = "a", long = "connect-addr")]
         alice_addr: Multiaddr,
 
@@ -77,6 +80,12 @@ pub enum Command {
     Resume {
         #[structopt(short = "id", long = "swap-id")]
         swap_id: Uuid,
+
+        #[structopt(short = "p", long = "connect-peer-id")]
+        alice_peer_id: PeerId,
+
+        #[structopt(short = "a", long = "connect-addr")]
+        alice_addr: Multiaddr,
 
         #[structopt(
             short = "b",
--- a/swap/tests/happy_path.rs
+++ b/swap/tests/happy_path.rs
@@ -1,5 +1,6 @@
 use crate::testutils::{init_alice, init_bob};
 use futures::future::try_join;
+use get_port::get_port;
 use libp2p::Multiaddr;
 use rand::rngs::OsRng;
 use swap::{alice, bob};
@@ -35,8 +36,8 @@ async fn happy_path() {
     let xmr_alice = xmr_to_swap * 10;
     let xmr_bob = xmr_btc::monero::Amount::ZERO;
 
-    // todo: This should not be hardcoded
-    let alice_multiaddr: Multiaddr = "/ip4/127.0.0.1/tcp/9876"
+    let port = get_port().expect("Failed to find a free port");
+    let alice_multiaddr: Multiaddr = format!("/ip4/127.0.0.1/tcp/{}", port)
         .parse()
         .expect("failed to parse Alice's address");
 
--- a/swap/tests/happy_path.rs
+++ b/swap/tests/happy_path.rs
@@ -63,7 +63,8 @@ async fn happy_path() {
 
     let (bob_state, bob_event_loop, bob_event_loop_handle, bob_btc_wallet, bob_xmr_wallet, bob_db) =
         init_bob(
-            alice_multiaddr,
+            alice_multiaddr.clone(),
+            alice_event_loop.peer_id(),
             &bitcoind,
             &monero,
             btc_to_swap,
@@ -83,6 +84,8 @@ async fn happy_path() {
         alice_db,
     );
 
+    let alice_peer_id = alice_event_loop.peer_id();
+
     let _alice_swarm_fut = tokio::spawn(async move { alice_event_loop.run().await });
 
     let bob_swap_fut = bob::swap::swap(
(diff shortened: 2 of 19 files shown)


rishflab (approved):
LGTM!

D4nte:
> LGTM!
> 
> This PR means we no longer have to work on this issue #89 or merge this PR #101?

Yes for #101. Not sure about #89.

swap/src/cli.rs
```diff
@@ -78,6 +81,12 @@ pub enum Command {
         #[structopt(short = "id", long = "swap-id")]
         swap_id: Uuid,
 
+        #[structopt(short = "p", long = "connect-peer-id")]
```
da-kami, line 84: ❗️ Please remove the option `p2p-address` (line 109) below in favor of these options. As far as I can see this is duplicated now. Note that for convenience it would help to keep  the `default` value, potentially also for the `alice_addr` above.

Please add another comment similar to the one on option `p2p-address` - the connection options are actually only relevant for Bob (there is a mistake in the current comment) because he has to be able to dial Alice. 
Since `Resume` is used by both parties this is a bit confusing but we can deal with it later.

swap/tests/happy_path.rs
```diff
@@ -35,8 +36,8 @@ async fn happy_path() {
     let xmr_alice = xmr_to_swap * 10;
     let xmr_bob = xmr_btc::monero::Amount::ZERO;
 
-    // todo: This should not be hardcoded
-    let alice_multiaddr: Multiaddr = "/ip4/127.0.0.1/tcp/9876"
+    let port = get_port().expect("Failed to find a free port");
+    let alice_multiaddr: Multiaddr = format!("/ip4/127.0.0.1/tcp/{}", port)
```
da-kami, line 40: 🙃 Would potentially be nicer to let `init_alice` return the connection information she will use. That way we can change this code in one place, independent of each test.

swap/tests/happy_path_restart_bob.rs
```diff
@@ -158,3 +163,149 @@ async fn given_bob_restarts_after_encsig_is_sent_resume_swap() {
...
+    let alice_fut_handle = tokio::spawn(alice_fut);
+
+    // We are selecting with bob_event_loop_1 so that we stop polling on it once
+    // bob reaches `xmr locked` state.
+    let bob_restart_state = select! {
+        res = bob_xmr_locked_fut => res.unwrap(),
+        _ = bob_event_loop_1.run() => panic!("The event loop should never finish")
+    };
+
+    // let tx_lock_id = if let AliceState::BtcRefunded { state3, .. } = alice_state
```
da-kami, line 264: ❓ Is this code still relevant?

swap/tests/happy_path_restart_bob.rs
```diff
@@ -158,3 +163,149 @@ async fn given_bob_restarts_after_encsig_is_sent_resume_swap() {
...
+
+    let bob_final_state = select! {
+     bob_final_state = bob_fut => bob_final_state.unwrap(),
+     _ = bob_event_loop_2.run() => panic!("Event loop is not expected to stop")
+    };
+
+    assert!(matches!(bob_final_state, BobState::XmrRedeemed));
+
+    // Wait for Alice to finish too.
+    let alice_final_state = alice_fut_handle.await.unwrap().unwrap();
```
da-kami, line 292: ❗️ Nice! I think we have to do this for the other restart tests as well otherwise they might fail. This relates to what was discussed here: https://github.com/comit-network/xmr-btc-swap/pull/89

da-kami (changes requested):
Nice work! 
I would vote for fixing the CLI `resume` command and changing the other restart tests to wait for the other party properly before we merge.

swap/tests/happy_path.rs
rishflab: I think we did this or planned to do it #101

swap/src/cli.rs
```diff
@@ -47,6 +47,9 @@ pub enum Command {
         receive_bitcoin: bitcoin::Amount,
     },
     BuyXmr {
+        #[structopt(short = "p", long = "connect-peer-id")]
```
rishflab, line 50: I don't think it makes to have  peer-id in the CLI if we don't have a way to configure multiple multiaddr's for a peer.

rishflab (changes requested):

swap/src/cli.rs
D4nte: Multiple multiaddr for a CLI sounds like a feature.

da-kami:
I thought with making Alice's listening port random we would not have to use separate files anymore. Why was https://github.com/comit-network/xmr-btc-swap/pull/100/commits/1e4b1155564edbf479d57d9a28b0e778c292c2c7 needed? 
Is the logic of `get_port` to get a random unused port?

D4nte:
> I thought with making Alice's listening port random we would not have to use separate files anymore. Why was [1e4b115](https://github.com/comit-network/xmr-btc-swap/commit/1e4b1155564edbf479d57d9a28b0e778c292c2c7) needed?
> Is the logic of `get_port` to get a random unused port?

It looked like I had performance issues in the CI with two tests in one file.

da-kami (approved):
We can do the following things in follow-ups: 

* All tests that restart have to properly wait until all parties have finished execution before asserting balances.
* Cleanup the CLI `resume` command (duplicated address information).
* Investigate why we still need separate files for executing tests, #101 had fixed that by using https://docs.rs/portpicker/0.1.0/portpicker/, but we closed in in favor of this PR using https://docs.rs/get-port/3.0.0/get_port/

da-kami:
> > I thought with making Alice's listening port random we would not have to use separate files anymore. Why was [1e4b115](https://github.com/comit-network/xmr-btc-swap/commit/1e4b1155564edbf479d57d9a28b0e778c292c2c7) needed?
> > Is the logic of `get_port` to get a random unused port?
> 
> It looked like I had performance issues in the CI with two tests in one file.

Interesting. I would have guessed the performance would get better if we had multiple test in one file.

swap/src/cli.rs
D4nte: What I mean is that it looks like that can be a future feature.

swap/src/cli.rs
D4nte: Done in https://github.com/comit-network/xmr-btc-swap/pull/102

swap/tests/happy_path_restart_bob.rs
D4nte: Done

PR #101, opened 2020-12-18
Randomly select unused ports to listen on for testing

rishflab:
Randomly select unused ports to listen on for testing.

Used port picker library. It is not maintained but we had a look at the code and it is fairly straightforward and only uses rand 0.6 as a dependency


da-kami:
I still think it would be more elegant to let `init_alice` return her multi-address. Not that we are using `port_picker` that should be possible without obstacles.

da-kami:
Closed in favor of https://github.com/comit-network/xmr-btc-swap/pull/100

PR #102, opened 2020-12-21
Separate Buy/Sell for resume command

D4nte:
And other clean ups from #100.

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -60,15 +60,14 @@ pub enum Command {
         receive_monero: xmr_btc::monero::Amount,
     },
     History,
-    Resume {
+    Resume(Resume),
+}
+
+#[derive(structopt::StructOpt, Debug)]
+pub enum Resume {
+    SellXmr {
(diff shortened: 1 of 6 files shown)


swap/src/cli.rs
```diff
@@ -104,15 +103,40 @@ pub enum Command {
         )]
         monero_wallet_rpc_url: Url,
 
-        // TODO: The listen address is only relevant for Alice, but should be role independent
-        //  see: https://github.com/comit-network/xmr-btc-swap/issues/77
         #[structopt(
-            short = "a",
-            long = "p2p-address",
+            short = "l",
```
da-kami, line 107: I'd remove all the short-codes in accordance with https://github.com/comit-network/xmr-btc-swap/pull/108 which will remove them for all other things.

da-kami (approved):

swap/src/cli.rs
D4nte: Let's get #108 merged first so I can rebase once not twice :)

rishflab (approved):

PR #104, opened 2020-12-21
Make application runable by changing shortcode to avoid duplication

da-kami:


swap/src/cli.rs
```diff
@@ -107,7 +107,7 @@ pub enum Command {
         // TODO: The listen address is only relevant for Alice, but should be role independent
         //  see: https://github.com/comit-network/xmr-btc-swap/issues/77
         #[structopt(
-            short = "a",
+            short = "la",
```
D4nte, line 110:
```suggestion
            short = "pa",
```
`l` is not even present in the `long` version.

D4nte (changes requested):
Please use later present in the long description.

rishflab (approved):

swap/src/cli.rs
da-kami: Crap, was looking at the variable name when creating the short-code 😅
Maybe we should just get rid of the short-codes altogether, they have caused nothing but trouble...

da-kami:
@D4nte after thinking about this again: How about we get rid of all the shortcodes for now? They have cause nothing but trouble, I regret adding them when working on the CLI... 
They could easily be added later.

da-kami:
Closing this in favor of https://github.com/comit-network/xmr-btc-swap/pull/108

PR #106, opened 2020-12-21
Move tests under `src`

D4nte:
The aim is to remove `lib.rs` to ensure we have no dead code.

However, because we have different clippy linter in `lib.rs` and `main.rs` there is some work to be done before we can do so.

This is the first self-contained step.


D4nte:
Need to investigate CI test failures

PR #107, opened 2020-12-21
Ensure that Bob can cancel correctly if T1 expired and Alice did not …

da-kami:
Bob has to check for the possibility to cancel in every state after he locked the BTC.
Otherwise Bob will try to perform actions that don't have any point and it might be impossible to use the `resume` command because it will always fail in trying to go on with Alice even though that might not be possible.

### DIFF
--- a/swap/src/alice/swap.rs
+++ b/swap/src/alice/swap.rs
@@ -429,6 +429,10 @@ pub async fn run_until(
                 state3,
                 encrypted_signature,
             } => {
+                // TODO: Evaluate if it is correct for Alice to Redeem no matter what.
+                //  If T1 expired she should potentially not try redeem. (The implementation
+                //  gives her an advantage.)
+
                 let signed_tx_redeem = match build_bitcoin_redeem_transaction(
                     encrypted_signature,
                     &state3.tx_lock,
--- a/swap/src/bob/swap.rs
+++ b/swap/src/bob/swap.rs
@@ -227,13 +227,32 @@ where
                 {
                     event_loop_handle.dial().await?;
 
-                    // todo: watch until t1, not indefinitely
-                    let msg2 = event_loop_handle.recv_message2().await?;
-                    let state4 = state3
-                        .watch_for_lock_xmr(monero_wallet.as_ref(), msg2)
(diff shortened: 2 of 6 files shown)


xmr-btc/src/lib.rs
```diff
@@ -60,4 +60,26 @@ pub mod monero;
 pub mod serde;
 pub mod transport;
 
+use crate::bitcoin::{BlockHeight, TransactionBlockHeight, WatchForRawTransaction};
 pub use cross_curve_dleq;
+
+pub async fn current_epoch<W>(
```
D4nte, line 66: This should go in `src/bitcoin.rs`.

Please keep in mind that `lib.rs` should and will be deleted.

swap/src/bob/swap.rs
```diff
@@ -236,16 +236,22 @@ where
...
+                let state = if let Epoch::T0 = state3.current_epoch(bitcoin_wallet.as_ref()).await?
+                {
+                    event_loop_handle.dial(alice_peer_id.clone()).await?;
 
-                // todo: watch until t1, not indefinitely
-                let msg2 = event_loop_handle.recv_message2().await?;
-                let state4 = state3
-                    .watch_for_lock_xmr(monero_wallet.as_ref(), msg2)
-                    .await?;
+                    // todo: watch until t1, not indefinitely
```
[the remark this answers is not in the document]
da-kami: Done

swap/src/alice/swap.rs
```diff
@@ -429,6 +429,10 @@ pub async fn run_until(
                 state3,
                 encrypted_signature,
             } => {
+                // TODO: Evaluate if it is correct for Alice to Redeem no matter what.
```
D4nte, line 432: True, there can be a race condition between the redeem and cancel transaction after T1.
Bob learns the secret to redeem Monero if he sees the transaction in the mempool and then can refund Bitcoin if he can race to get the cancel tx in.

I propose to solve this todo with part 2 of this plan: https://lists.comit.network/pipermail/comit-dev/2020-December/000031.html

D4nte (approved):

PR #108, opened 2020-12-21
Remove all CLI short-codes because they don't add value at the moment

da-kami:
`struct_opt` does not check at compile time, if there is a duplicate. Only when running the application. We have been running in this annoyance twice, so let's remove the shortcodes for now.


D4nte (approved):

da-kami:
bors retry

PR #109, opened 2020-12-22
Allow Bob to try to refund after T2 has passed

da-kami:
If we don't allow this, then Bob cannot refund Bitcoin if:
Alice started a swap with Bob but never acts after the start (and remains offline).
Bob is offline until T2 has passed.

Note: This helps recovering funds, but strictly speaking we would not have to implement it because in a scenario where T2 is expired Alice should punish. Bob has to act to refund after T1 but before T2 expired. I would still add this, because it should not change the behavior and makes recovery for Bob easier.

### DIFF
--- a/swap/src/bob/swap.rs
+++ b/swap/src/bob/swap.rs
@@ -10,7 +10,7 @@ use async_recursion::async_recursion;
 use rand::{CryptoRng, RngCore};
 use std::{convert::TryFrom, fmt, sync::Arc};
 use tokio::select;
-use tracing::info;
+use tracing::{info, warn};
 use uuid::Uuid;
 use xmr_btc::{
     bob::{self, State2},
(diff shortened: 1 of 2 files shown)


D4nte (approved):

swap/src/bob/swap.rs
```diff
@@ -386,15 +386,24 @@ where
...
                         state.refund_btc(bitcoin_wallet.as_ref()).await?;
                         BobState::BtcRefunded(state)
                     }
-                    Epoch::T2 => BobState::Punished,
+                    Epoch::T2 => {
+                        // If T2 expired we still try to refund in case Alice has not acted.
+                        // This is especially important for scenarios where Alice is offline
+                        // indefinitely after starting the swap.
+                        let refund_result = state.refund_btc(bitcoin_wallet.as_ref()).await;
+                        match refund_result {
```
D4nte, line 402: I think we should analyse the error before assuming the failure is due to Alice punishing

da-kami:
I just tried to do this (I bypassed the `Epoch` check), and it actually did run into an error (unfortunately I did not log it yet). So, I would no merge this, but in such a scenario Alice has to `punish`. We could eventually add this for Bob, but it's probably not a good idea to do it now as it adds more unknowns.

swap/src/bob/swap.rs
da-kami: I added a log statement.

da-kami:
After another look I noticed we are sending `tx_cancel` when `refunding` again. I don't think this is expected behavior, given that `T1Expired` was introduced for Bob as well. Se last commit remove sending `tx_cancel` when refunding and relies on `T1Expired` correctly handling `tx_cancel`.

da-kami:
Please have another look @D4nte.

swap/src/bob/swap.rs
```diff
@@ -390,15 +390,28 @@ where
...
+                    Epoch::T2 => {
+                        // If T2 expired we still try to refund in case Alice has not acted.
+                        // This is especially important for scenarios where Alice is offline
+                        // indefinitely after starting the swap.
+                        info!("Trying to refund even though T2 has passed...");
+                        let refund_result = state.submit_tx_refund(bitcoin_wallet.as_ref()).await;
+                        match refund_result {
+                            Ok(()) => BobState::BtcRefunded(state),
+                            Err(err) => {
+                                warn!("Bob tried to refund after T2 has passed, but sending the refund tx errored with {}", err);
```
D4nte, line 410:
```suggestion
                                error!("Bob tried to refund after T2 has passed, but sending the refund tx errored with {}", err);
```
This is a pretty bad scenario.

D4nte (approved):

swap/src/bob/swap.rs
```diff
@@ -10,7 +10,7 @@ use async_recursion::async_recursion;
 use rand::{CryptoRng, RngCore};
 use std::{convert::TryFrom, fmt, sync::Arc};
 use tokio::select;
-use tracing::info;
+use tracing::{info, warn};
```
da-kami, line 13:
```suggestion
use tracing::{info, error};
```

rishflab (approved):

PR #118, opened 2021-01-05
Add minimum monero confirmations to config use it for Bob

da-kami:
After discussion with @bonomat about enforced `10` confirmations until a tx is spendable on Monero I gave it a try with only `1` confirmation on regtest happy-path-test and `5` for `stagenet` and did not encounter problems. Both swaps passed without problems so I am proposing to merge these changes. 

One additional change crept in: Changed the average blocktime for `stagenet` to `4` mins, because the blocktime fluctuates a lot. See: https://monero-stagenet.exan.tech/

### DIFF
--- a/xmr-btc/src/config.rs
+++ b/xmr-btc/src/config.rs
@@ -8,6 +8,7 @@ pub struct Config {
     pub bitcoin_finality_confirmations: u32,
     pub bitcoin_avg_block_time: Duration,
     pub monero_max_finality_time: Duration,
+    pub monero_finality_confirmations: u32,
     pub bitcoin_cancel_timelock: Timelock,
     pub bitcoin_punish_timelock: Timelock,
     pub bitcoin_network: ::bitcoin::Network,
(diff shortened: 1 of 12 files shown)


da-kami:
Note that that spending from confirmations with less than `10` confirmations should not work on `stagenet`. 
Further investigation is needed why the swap worked. It might be that the tx was held back because the newly created wallet was still not done rescanning. Once the rescan optimization is implemented rebase this PR and try again on `stagenet`. It should fail with `Transaction spends at least one output which is too young`.

PR that merged the enforced 10 confirmations: https://github.com/monero-project/monero/pull/5882

xmr-btc/src/config.rs
```diff
@@ -24,6 +25,7 @@ impl Config {
             // blockchain is slow
             monero_max_finality_time: (*mainnet::MONERO_AVG_BLOCK_TIME).mul_f64(1.5)
                 * mainnet::MONERO_FINALITY_CONFIRMATIONS,
+            monero_min_confirmations: mainnet::MONERO_FINALITY_CONFIRMATIONS,
```
D4nte, line 28:
```suggestion
            monero_finality_confirmations: mainnet::MONERO_FINALITY_CONFIRMATIONS,
```
Please use the same terminology across the codebase.

D4nte (approved):

da-kami:
Merging this now, further changes concerning removing the redeem confirmations (because they don't add value at the moment) can be done in follow up PR. This definitely helps us with the e2e tests making them quite a lot faster :)

PR #119, opened 2021-01-06
Remember the block-height before XMR lock for generated monero wallet

da-kami:
The first approach https://github.com/comit-network/xmr-btc-swap/pull/121 was using `get_transfer_by_txid` that allows extrancting the exact tx-lock 1st confirmation block height. 
But that introduced an additional error scenario, and I actually ran into that error scenario (`transaction not found`) once I ran it on `stagenet`. Might be that `get_transfer_by_txid` requires running the node in a specific way (like `txindex` on bitcoin).  
I am not sure at this stage and don't want to invest more time.

Long story short: 
I opted for just recording the height before watching for XMR locked. This means that we record a height right after sending the Bitcoin lock tx. (Because we start watching for XMR lock right after that.) 
Bob's new wallet unnecessarily scans an additional 7+ blocks (assuming inclusion in the next Bitcoin block and one confirmation for Monero lock) every time which is a matter of milliseconds. Not worth optimising this further at this stage. This solution is more resilient as well, because it does not add another error scenario.

### DIFF
--- a/monero-harness/src/rpc/wallet.rs
+++ b/monero-harness/src/rpc/wallet.rs
@@ -186,7 +186,7 @@ impl Client {
     }
 
     /// Get wallet block height, this might be behind monerod height.
-    pub(crate) async fn block_height(&self) -> Result<BlockHeight> {
+    pub async fn block_height(&self) -> Result<BlockHeight> {
         let request = Request::new("get_height", "");
 
         let response = self
@@ -238,9 +238,16 @@ impl Client {
         address: &str,
         spend_key: &str,
         view_key: &str,
+        restore_height: Option<u32>,
     ) -> Result<GenerateFromKeys> {
+        let restore_height = if let Some(restore_height) = restore_height {
+            restore_height
+        } else {
+            0
+        };
+
         let params = GenerateFromKeysParams {
-            restore_height: 0,
+            restore_height,
             filename: view_key.into(),
             address: address.into(),
             spendkey: spend_key.into(),
--- a/swap/src/monero.rs
+++ b/swap/src/monero.rs
@@ -210,6 +210,7 @@ pub trait CreateWalletForOutput {
         &self,
         private_spend_key: PrivateKey,
         private_view_key: PrivateViewKey,
+        restore_height: Option<u32>,
     ) -> anyhow::Result<()>;
 }
 
--- a/swap/src/monero/wallet.rs
+++ b/swap/src/monero/wallet.rs
@@ -68,6 +68,7 @@ impl CreateWalletForOutput for Wallet {
         &self,
         private_spend_key: PrivateKey,
         private_view_key: PrivateViewKey,
+        restore_height: Option<u32>,
     ) -> Result<()> {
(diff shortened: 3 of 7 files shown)


swap/src/bob/swap.rs
```diff
@@ -181,11 +181,19 @@ where
                     let cancel_timelock_expires =
                         state3.wait_for_cancel_timelock_to_expire(bitcoin_wallet.as_ref());
 
+                    // Record the current monero wallet block height so we don't have to scan from
```
D4nte, line 184: People with a wallet few years old would still have an issue, could you go for the correct solution to pass the lock transaction height?

xmr-btc/src/alice.rs
```diff
@@ -884,10 +884,11 @@ impl State5 {
 
         let s = s_b.scalar + self.s_a.into_ed25519();
 
+        // TODO: Optimized rescan height should be passed for refund as well.
```
D4nte, line 887: Can you do this now?

xmr-btc/src/bob.rs
```diff
@@ -620,6 +558,7 @@ impl State3 {
             tx_lock: self.tx_lock,
             tx_cancel_sig_a: self.tx_cancel_sig_a,
             tx_refund_encsig: self.tx_refund_encsig,
+            monero_rescan_block_height,
```
D4nte, line 561:
```suggestion
            lock_transaction_height,
```

Instead of naming it after *how* we want to use it, let's name it after *what* it is.

D4nte (approved):
It's ok to go but I think we should implement the comments first.
I am happy to take over if you don't want, just ping me please.

swap/src/bob/swap.rs
da-kami: I don't understand what you mean. We are creating a new wallet every time and pass the `restore_height` to it. I don't understand how this involves "People with a wallet few years old ".

da-kami:
@D4nte If you are keen to understand why https://github.com/comit-network/xmr-btc-swap/pull/121 please go ahead and implement the solution that records the exact lock tx height. I'd be keen to see it, but would not give it priority at this point.

swap/src/bob/swap.rs
D4nte: My understanding was incorrect :) thanks.

xmr-btc/src/bob.rs
da-kami: I would opt for doing this in the follow up. At the moment it is not precisely the `lock_transaction_height`, but a height that we record before we see the lock tx that is recorded for the sole purpose of using it for the wallet creation. I would make the name explicit, otherwise it could be confusing because one might think the height is relevant for watching the transactions. Once we implement extracting the exact height from the transfer most of what is currently there would be removed and the remaining parts can be renamed.

xmr-btc/src/alice.rs
da-kami: I would action it later, because we are focusing on happy-path and I would not optimize for refund yet.

PR #121, opened 2021-01-06
Use get_transfer_by_txid to extract the exact block height of txlock

da-kami:
This is just for reference. If you spot what the problem could be we could still go for this, otherwise I would opt for the solution proposed by https://github.com/comit-network/xmr-btc-swap/pull/119
This fails with `transaction not found` on `stagenet`, not sure why.

### DIFF
--- a/monero-harness/src/rpc/wallet.rs
+++ b/monero-harness/src/rpc/wallet.rs
@@ -185,7 +185,7 @@ impl Client {
         Ok(r.result)
     }
 
-    /// Get wallet block height, this might be behind monerod height.
+    /// Get wallet block height.
     pub async fn block_height(&self) -> Result<BlockHeight> {
         let request = Request::new("get_height", "");
 
--- a/monero-harness/src/rpc/wallet.rs
+++ b/monero-harness/src/rpc/wallet.rs
@@ -186,7 +186,7 @@ impl Client {
     }
 
     /// Get wallet block height, this might be behind monerod height.
-    pub(crate) async fn block_height(&self) -> Result<BlockHeight> {
+    pub async fn block_height(&self) -> Result<BlockHeight> {
         let request = Request::new("get_height", "");
 
         let response = self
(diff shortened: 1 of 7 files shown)


D4nte:
> This is just for reference. If you spot what the problem could be we could still go for this, otherwise I would opt for the solution proposed by #119
> This fails with `transaction not found` on `stagenet`, not sure why.

Is it fine in regtest?

da-kami:
> > This is just for reference. If you spot what the problem could be we could still go for this, otherwise I would opt for the solution proposed by #119
> > This fails with `transaction not found` on `stagenet`, not sure why.
> 
> Is it fine in regtest?

Happy path test failed on CI, so there is definitely a problem. It might be time consuming to pin it down.

D4nte:
`get_transfer_by_txid` only work if the transfer is to or from the current wallet.

da-kami:
> `get_transfer_by_txid` only work if the transfer is to or from the current wallet.

In my understanding the transfer is to Bob's wallet so it should work. But true, we are creating this wallet later, so it might not be recorded as transaction to said wallet. I wonder how we can extract the height from the tx then.

D4nte:
> > `get_transfer_by_txid` only work if the transfer is to or from the current wallet.
> 
> In my understanding the transfer is to Bob's wallet so it should work. But true, we are creating this wallet later, so it might not be recorded as transaction to said wallet. I wonder how we can extract the height from the tx then.

The transfer is not to Bob's wallet as the transfer is locked with keys that are revealed by the Bitcoin redeem/punish adaptor signatures.

I don't think we can extract the height but elementary maths should help.

da-kami:
> > > `get_transfer_by_txid` only work if the transfer is to or from the current wallet.
> > 
> > 
> > In my understanding the transfer is to Bob's wallet so it should work. But true, we are creating this wallet later, so it might not be recorded as transaction to said wallet. I wonder how we can extract the height from the tx then.
> 
> The transfer is not to Bob's wallet as the transfer is locked with keys that are revealed by the Bitcoin redeem/punish adaptor signatures.
> 
> I don't think we can extract the height but elementary maths should help.

If we really cannot extract the exact height I would stick to the simple solution of https://github.com/comit-network/xmr-btc-swap/pull/119 - this is bulletproof and does not cost much (scanning a few more blocks won't hurt). I would like to avoid further error scenarios.

D4nte:
> > > > `get_transfer_by_txid` only work if the transfer is to or from the current wallet.
> > > 
> > > 
> > > In my understanding the transfer is to Bob's wallet so it should work. But true, we are creating this wallet later, so it might not be recorded as transaction to said wallet. I wonder how we can extract the height from the tx then.
> > 
> > 
> > The transfer is not to Bob's wallet as the transfer is locked with keys that are revealed by the Bitcoin redeem/punish adaptor signatures.
> > I don't think we can extract the height but elementary maths should help.
> 
> If we really cannot extract the exact height I would stick to the simple solution of #119 - this is bulletproof and does not cost much (scanning a few more blocks won't hurt). I would like to avoid further error scenarios.

I am happy for us to move forward with the solution implemented in #119 where Bob Remembers the block height after Bitcoin is locked and before he sends the encrypted signature (to redeem BTC) to Alice.

PR #126, opened 2021-01-08
Deterministic peer ids

da-kami:
The first commit introduces a `seed` file similar to Nectar. Note, that the parameter `--database` was changed to `--data-dir` where the database is stored in the sub-folder `database` inside that data directory. This is breaking change, run commands will have to be adapted.
I opted for keeping the structure of generating an overall seed and then deriving the network see from it (as in Nectar, where we use the seed for wallet creation as well). I feel this is cleaner than just using the seed for the network only. 

The second commit applies the deterministic peer id to Bob as well. We don't have to do this, in the current setup Bob can have a new identity every time. I would still harmonize this to avoid confusion in the future. I don't see a reason why Bob's setup should be different from Alice here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2243,6 +2243,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "pem"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


da-kami:
Needs a bit more refactoring. The identity should be part of the EventLoop.

PR #127, opened 2021-01-08
Deterministic peer id for alice and bob

da-kami:
The first commit introduces a seed file similar to Nectar. Note, that the parameter --database was changed to --data-dir where the database is stored in the sub-folder database inside that data directory. This is breaking change, run commands will have to be adapted.
I opted for keeping the structure of generating an overall seed and then deriving the network seed from it (as in Nectar, where we use the seed for wallet creation as well). I feel this is cleaner than just using the seed for the network only.

The second commit applies the deterministic peer id to Bob as well. We don't have to do this because in the current setup Bob can have a new identity every time. I would still harmonize this to avoid confusion in the future. I don't see a reason why Bob's setup should be different from Alice here.

### DIFF
--- a/swap/src/network.rs
+++ b/swap/src/network.rs
@@ -25,17 +25,17 @@ pub struct Seed([u8; SEED_LENGTH]);
 
 impl Seed {
     /// prefix "NETWORK" to the provided seed and apply sha256
-    pub fn new(seed: [u8; crate::seed::SEED_LENGTH]) -> Self {
+    pub fn new(seed: crate::seed::Seed) -> Self {
         let mut engine = sha256::HashEngine::default();
 
-        engine.input(&seed);
+        engine.input(&seed.bytes());
         engine.input(b"NETWORK");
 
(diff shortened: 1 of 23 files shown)


rishflab (approved):

swap/src/network.rs
```diff
@@ -17,3 +19,35 @@ impl Executor for TokioExecutor {
...
+        let mut engine = sha256::HashEngine::default();
+
+        engine.input(&seed);
+        engine.input(b"NETWORK");
+
+        let hash = sha256::Hash::from_engine(engine);
+        Self(hash.into_inner())
+    }
+
+    pub fn bytes(&self) -> [u8; SEED_LENGTH] {
```
D4nte, line 38:
```suggestion
    fn bytes(&self) -> [u8; SEED_LENGTH] {
```
This is not publicly used

D4nte (approved):

D4nte (approved):

PR #125, opened 2021-01-07
Merge xmr_btc crate

rishflab:
Created network, storage and protocol modules. Organised
files into the modules where the belong.

xmr_btc crate moved into isolated modulein swap crate.

Remove the xmr_btc module and integrate into swap crate.

Consolidate message related code

Reorganise imports

Remove unused parent Message enum

Remove unused parent State enum

Remove unused dependencies from Cargo.toml

PR #128, opened 2021-01-10
Remove lib and move tests into src

rishflab:
This commits were approved cut out out of PR #125 because CI was failing.


swap/src/main.rs
```diff
@@ -14,25 +14,31 @@
...
+pub mod database;
+pub mod monero;
+pub mod network;
+pub mod protocol;
+#[cfg(test)]
+mod tests;
+pub mod trace;
+
 #[macro_use]
 extern crate prettytable;
```
D4nte, line 43: Is this needed? I don't think the `#[macro_use]` and `extern` keywords are necessary anymore.

D4nte (approved):

da-kami:
Can be re-tried after test refactoring and adding additional tests is done. Closing for now, as not pressing at the moment.

PR #138, opened 2021-01-13
Add sequence diagram

D4nte:


docs/sequence.puml
```diff
@@ -0,0 +1,58 @@
...
+
+Bob --> Alice: Establish connection
+
+group Request Response Channel
+    Bob -> Alice: Request Amounts
+    note left: Btc Amount
+
+    Alice --> Bob: Send Amounts
+    note right: Btc & Xmr Amounts
+end
```
bonomat, line 15: I'm a bit confused on the first step: 
Bob is the maker (wanting BTC) and Alice the taker (wanting XMR).
Bob comes up with the rate. So the first request should be from Alice to Bob : `Get Rate ` or similar.

docs/sequence.puml
```diff
@@ -0,0 +1,58 @@
...
+    Bob -> Alice: bob::Message0
+    note left: Pubkeys\ndleq proof sb\nxmr viewkey vb\nbtc refund addr
+
+    Alice --> Bob: alice::Message0
+    note right: Pubkeys\ndleq proof sa\nxmr view key va\nbtc redeem addr\nbtc punish addr
+end
+
+group Request Response Channel
+    Bob -> Alice: bob::Message1
+    note left: btc lock tx
```
bonomat, line 27: What's the level of detail you want to have? Might be important to point out that this is not yet signed.

docs/sequence.puml
```diff
@@ -0,0 +1,58 @@
...
+
+    Alice --> Bob: alice::Message1
+    note right: cancel tx sig\ntx refund enc sig
+end
+
+group Request Response Channel
+    Bob -> Alice: bob::Message2
+    note left: punish tx sig\ncancel tx sig
+
+== Negotiated ==
```
bonomat, line 37: What does this mean please? (:

bonomat (commented):
Looks good from Memory. The messages make sense but I do not remember if this reflects the reality (code). Maybe @luckysori has a look as well. 

For the next reviewer: This might be useful: [plantuml.com](https://www.planttext.com/?text=hLIzJiCm4Dxz53Un8T6oHIfgYxe1GcLXwE8ubzHgScxj2vItv-eaBKqZX80iJdpdxyzsfe6rvxegLQPPfpeWp2qRPqbfmyx3tALg5jzcfJLuQdjqXDuf9GrmVNtNb1DO261Qsh06uuZGi7MaLE5TlOKbxceCB6FOEXAC-xMMdbA1F17d2DErpYfN4uVOGOuHIimvWJcRjgHYAQwACWGcWHUax7AfjyMwNNi5hvK_TaZtBoIcmfGyOWYwm9k-mESwtU17M55MuWwsthaSGhgYlT2-MNoN8hp9TofYFEQrwDLPvWVTwCWtGDKQ6kJI9ow8PBeZon2hIDPCR6iwx5Pa_tiGetuG1uhIcGtm_WVcHXVcZ2Q3fQo6O8iLoTZ6XMGEK_-WUTpNt4RISPv9K6eoWISi75lDcC5aehx0ttMt9u47CNtcM8hDdHce3QSnlaXZNm5xJI57tsnqO_zxVxTzVuiskfwwe_HbF2CjR3FPDCIxPGlIN7iyeusbnLx46Mvj7exTUJQT_Gv_AkBbr_C9)

docs/sequence.puml
```diff
@@ -0,0 +1,58 @@
...
+    Bob ->> Bitcoin: Lock
+
+    Alice ->> Monero: Lock
+
+    Alice --> Bob: alice::Message2
+    note right: xm transfer proof
+end
+
+    Bob -> Alice: bob::Message3
+    note left: Enc sig
```
luckysori, line 48: :thinking: If your intention with the diagram is to reflect the state of the codebase I think this should still be part of a `Request Response Channel` group and `Alice` will send back an empty response. But you probably already know this and it's a deliberate omission.

docs/sequence.puml
D4nte: This is a direct translation of the code. I intend to remove this part as indeed, it is not helpful.

luckysori (approved):
It looks good to me! I'll push a commit with some minor edits, but you can drop it if you don't like them.

docs/sequence.puml
D4nte: Good to ensure it's correct, I am not going to modify the content of the execution part. The intent of this PR is to have a starting document from which I can plan the next steps:
- replace request amounts/amounts with appropriate messages for single market maker
- Replace request response channels with one shot messages

docs/sequence.puml
D4nte: It means that negotiation phase ended. will replace.

docs/sequence.puml
D4nte: I missed that, thanks!

D4nte: Actually, looking at the codebase it does not seem to be the case. Alice does not seem to reply.

docs/sequence.puml
bonomat: Interesting, for me the negotiation phase ends after the rate and amounts are exchanged (or before the first protocol-related message was exchanged).

docs/sequence.puml
luckysori: I was referring to this line, but maybe it's no longer used: https://github.com/comit-network/xmr-btc-swap/blob/07270df12a6805466a5c1cba12bae9e07fdbe1e8/swap/src/protocol/alice/message3.rs#L81

docs/sequence.puml
bonomat: My initial comment was bollocks: Alice has XMR and wants BTC and Bob has BTC and has XMR.

docs/sequence.puml
D4nte: Indeed, it looks like we have introduced a terminology in the code that does not match what we previously agree in terms of negotiation vs execution. I will propose a new terminology as part this exercise.

docs/sequence.puml
D4nte: I actually look for that initially and did not find it! Thanks.

PR #140, opened 2021-01-13
[doc] Remove Req-Res and update for single market maker use case

D4nte:
- Bob initiate swap with BTC Amount
- Alice acks with approx. xmr amount
- Rework terminology, make a difference between negotiations and
execution setup
- Remove request response channels in favor of one shots
- Give exact amount to Bob once assets are locked
- Improve message naming

[plantuml](https://www.planttext.com/?text=XPHDZzem48Rl-HMZFLJRmrRzl7102nMtKY5efKgbMZdEa5YRZBEsmsR_VSTs4WWRMYG8icVUUUSPn_VM2UEQgamouKGgBC9IEQaL9K8wRM2fKn0MRlppzgT5y_OcMOITHQaaXhtuxtQ7vkZ_R8m-gYn43g9hJMXqakH6DpLynrmx9PpIb2J07wz-TpUFcZDOiQkqLBO0gObGdW9Z_XQV6hIEdxRMnF9V2y5nPGWPz0QvtREemSIS8OIqGoZnu6QmT18MbMx8xKbacK5heL3ax3ZOgBpWwDYjPGHeeKOqe39u0wF4FJdjlJk6wCqaWrBCzhtoyc0jjZ5PRjLwZL_hBOXWCGGXPOd_TZnNBSh6Gu8TkgO-8xKfFFq5_5wpDv6pOoa8KWHi9JFBCGD58EW5jE6H_Rb2C6MliqxY_HJ2JPC-uWltdPNu1BNH-W3s8TrJMnauAdpcNJZwXPHX6pmqb0LahyXqO4GeVALcXsQwg3WLXLXLz5Kpn2fKZGjrG_wC1HizrLVibcVigeOFtIKyLo3aweZK8xp2i5FvC8Ml8QjoOZrlgzJo4LnxZKodyt4IYS_d2KiiMG5OT0ISbv2aNuRTG_eVev-cZ9vXEzSVbKp-UKHFx_jytjyrC_Z6FSUyljbvTsTCxOr0V4uk8VZnTn316K7sm2zfE1vx-b6euHqmMEcZ7-8X67eMJXOyPJ_GrEVpa1MzExkXBX5zIIxmT4Tin5bSzZKqlGtHdNvSZ07XPk038LnZS5G_RArQxeel1TjhzpXxOfrmm7_Fl_v-_mi0)

### DIFF
--- a/docs/sequence.puml
+++ b/docs/sequence.puml
@@ -19,23 +19,23 @@ group Negotiation
 end
 
 group Execution Setup
-    group Phase 0 [Messages can be exchanged in any order]
-        Bob -> Alice: bob::Message0
+    group Phase A [Messages can be exchanged in any order]
+        Bob -> Alice: Message0
         note left: Pubkeys\ndleq proof s_b\nxmr viewkey v_b\nbtc refund addr
 
-        Alice -> Bob: alice::Message0
+        Alice -> Bob: Message1
         note right: Pubkeys\ndleq proof s_a\nxmr view key v_a\nbtc redeem addr\nbtc punish addr
     end
 
-    group Phase 1 [Messages must be exchanged in the given order]
+    group Phase B [Messages must be exchanged in the given order]
 
-        Bob -> Alice: bob::Message1
+        Bob -> Alice: Message2
         note left: unsigned btc lock tx
 
-        Alice -> Bob: alice::Message1
+        Alice -> Bob: Message3
         note right: btc cancel tx sig\nbtc refund tx enc sig S_b
 
-        Bob -> Alice: bob::Message2
+        Bob -> Alice: Message4
         note left: btc punish tx sig\nbtc cancel tx sig
 
     end
@@ -46,10 +46,10 @@ group Execution
 
     Alice ->> Monero: Lock
 
-    Alice -> Bob: alice::Message2
+    Alice -> Bob: Message5
     note right: Exact xmr amount\nzxmr lock tx transfer proof\nThis can be removed if Bob watches the blockchain.
 
-    Bob -> Alice: bob::Message3
+    Bob -> Alice: Message6
     note left: redeem tx enc sig S_a
 
     Alice ->> Bitcoin: Redeem
--- a/docs/sequence.puml
+++ b/docs/sequence.puml
@@ -4,52 +4,53 @@ actor Bob
 Actor Alice
 database Monero
 
-Bob --> Alice: Establish connection
+group Negotiation
 
-group Request Response Channel
-    Bob -> Alice: Request Amounts
-    note left: Btc Amount
+    Bob --> Alice: Establish connection
 
-    Alice --> Bob: Send Amounts
-    note right: Btc & Xmr Amounts
+    group Request Response Channel
+        Bob -> Alice: Swap request
+        note left: Btc Amount\naddress hints
+        note right: Alice stores peer id & address hints\nto contact Bob
+
+        Alice --> Bob: Swap response
+        note right: ACK or NACK\n Approx. XMR amount
+    end
 end
 
-group Request Response Channel
-    Bob -> Alice: bob::Message0
-    note left: Pubkeys\ndleq proof s_b\nxmr viewkey v_b\nbtc refund addr
+group Execution Setup
+    group Phase 0 [Messages can be exchanged in any order]
+        Bob -> Alice: bob::Message0
+        note left: Pubkeys\ndleq proof s_b\nxmr viewkey v_b\nbtc refund addr
 
-    Alice --> Bob: alice::Message0
-    note right: Pubkeys\ndleq proof s_a\nxmr view key v_a\nbtc redeem addr\nbtc punish addr
+        Alice -> Bob: alice::Message0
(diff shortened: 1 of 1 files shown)


docs/sequence.puml
```diff
@@ -4,52 +4,53 @@ actor Bob
...
 
-    Alice --> Bob: Send Amounts
-    note right: Btc & Xmr Amounts
+    group Request Response Channel
+        Bob -> Alice: Swap request
+        note left: Btc Amount\naddress hints
+        note right: Alice stores peer id & address hints\nto contact Bob
+
+        Alice --> Bob: Swap response
+        note right: ACK or NACK\n Approx. XMR amount
```
da-kami, line 17: At this stage I don't understand why it is still an approximate amount. I understand that the amount on a SP website has to be approximate, but right before the swap I suppose it would be the actual amount this is agreed upon. 
Is there no agreement needed to the exact amounts of the swap prior to the execution? The amounts are not part of calculating tx proofs?

D4nte: I will have to verify by trying to code it but as far as I can see for Monero, the amount is not needed until Monero is actually locked. Maybe I am incorrect, I'll double check tomorrow.

We want Alice to be able to provide the amount at the last minute as otherwise, Bob can wait before locking Btc (at now cost) to see whether the trade will go in his favour.

I think this solution is more elegant than putting in place arbitrary time outs to prevent that.

da-kami: Is the approximation necessary because of the funding step for the Bitcoin on the user side? 
Are we planning to just trigger the actual swap once the Bitcoin funds arrived in the swap-tool wallet?

thomaseizinger (commented):
I would make the message counter global and not per party. Makes it easier to reference and there isn't really an advantage to counting them individually.

D4nte:
> I would make the message counter global and not per party. Makes it easier to reference and there isn't really an advantage to counting them individually.

Totally agree!

docs/sequence.puml
bonomat: Imho Alice should tell Bob the **exact** XMR amount she is willing to commit. 
If Bob is not happy with this amount (rate), he can still bail out before publishing a single transaction. 
Obviously, the amounts are not enforced in anyway and Alice could use a different amount at the time of funding. 
But this would be really bad, because now both parties have issued a transaction on-chain, and the fees are wasted.

D4nte: > Imho Alice should tell Bob the **exact** XMR amount she is willing to commit.
> If Bob is not happy with this amount (rate), he can still bail out before publishing a single transaction.
> Obviously, the amounts are not enforced in anyway and Alice could use a different amount at the time of funding.
> But this would be really bad, because now both parties have issued a transaction on-chain, and the fees are wasted.
> 
> // Edit //
> 
> Different said: Bob should know how much XMR he will receive **before** issuing the first transaction (`tx_fund`).

As a single market maker, I want to be sure that I provide the right rate to not loose money.
If you look at the current ecosystem, whether it's crypto or forex, single market maker always provide an approximate rate. Then the user sends money and the rate gets locked once the money is received by the service provider.
Some single market makers also provide a "lock rate" service where the rate is guaranteed but is generally higher than the estimated flexible rate.

We could do fixed rates now but I worried that when we want to make the single market maker software MVP ready and we start to look at the functionalities that potential operators would want, we will have to revert to a flexible approximative rate.

D4nte: At this stage, I am happy to roll with a fixed rate and defer the discussion to when we try to make the Service Provider MVP.

bonomat: > As a single market maker, I want to be sure that I provide the right rate to not loose money.
If you look at the current ecosystem, whether it's crypto or forex, single market maker always provide an approximate rate. Then the user sends money and the rate gets locked once the money is received by the service provider.
> Some single market makers also provide a "lock rate" service where the rate is guaranteed but is generally higher than the estimated flexible rate.

For centralized services (SP) this works well, because the trader cannot change its mind half way through. The trader sends the money and hopes to receive at least something in return. If this is not good enough, the SP reputation suffers.

Now I think the story is different: 
The trader receives a promise from the maker to execute the trade at a certain rate. If the maker does not go through with the promised rate, he is in breach if this promise. 

Since we can't enforce this promise, we can argue whether this promise should never have been made in the first place and use an approximate rate instead. 
Personally, I prefer to know exactly how much money I will receive. 

In my head this preferably because of this use case: 
Bob receives a rate from Alice and locks up funds on Bitcoin. 
Alice locks up a different amount on XMR and now wants Bob to co-sign `tx_redeem` for Bitcoin.
If the XMR amount is not _good enough_ he will simply decline and both parties lost transaction fees. Hence, Alice best chance to go through with the trade is to tell Bob in advance how much she is willing to commit.

bonomat: > At this stage, I am happy to roll with a fixed rate and defer the discussion to when we try to make the Service Provider MVP.

Didn't see your reply. Eitherway, I don't think it's a big issue to change later one.

da-kami (approved):

PR #142, opened 2021-01-14
Rename Swap amounts to Swap req/res

D4nte:
As per #140

Question: Do we prefer `Negotiation Request/Response` to `Swap Request/Response`?


da-kami:
Oh my, I overlooked that one. I'll see to get the remainder of these changes in after the test/prod refactoring is done.
I remember using `SwapAmounts` on some occasions.

rishflab (approved):
Prefer Swap over Negotiation but I am not fussed either way.

PR #143, opened 2021-01-15
Upgrade to tokio 1.0

rishflab:
When we were refactoring tests we realised we probably want the ability to abort a `tokio::JoinHandle` to kill the `EventLoop` to simulate a real world crash. tokio 1.0 is needed for this. It is probably about time to upgrade tokio anyway. 

In order to upgrade to tokio 1.0 the following dependencies were also upgraded in the swap crate and monero-harness-rs
* backoff
* libp2p
* request

UPDATE: This should be merged until the following dependencies are uprgraded to Tokio 1.0 or Tokio compat  is used

- [x] bitcoin-harness-rs https://github.com/coblox/bitcoin-harness-rs/pull/20

### DIFF
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -10,10 +10,10 @@ anyhow = "1"
 async-recursion = "0.3.1"
 async-trait = "0.1"
 atty = "0.2"
-backoff = { version = "0.2", features = ["tokio"] }
+backoff = { git = "https://github.com/ihrwein/backoff", rev = "9d03992a83dfdc596be26276d4e5c5254a4b11a2", features = ["tokio"] }
 base64 = "0.12"
 bitcoin = { version = "0.25", features = ["rand", "use-serde"] }
-bitcoin-harness = { git = "https://github.com/coblox/bitcoin-harness-rs", rev = "864b55fcba2e770105f135781dd2e3002c503d12" }
+bitcoin-harness = { git = "https://github.com/coblox/bitcoin-harness-rs", rev = "ae2f6cd547496e680941c0910018bbe884128799" }
 conquer-once = "0.3"
 cross-curve-dleq = { git = "https://github.com/comit-network/cross-curve-dleq", rev = "eddcdea1d1f16fa33ef581d1744014ece535c920", features = ["serde"] }
 curve25519-dalek = "2"
@@ -21,8 +21,7 @@ derivative = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", rev = "cdfbc766045ea678a41780919d6228dd5acee3be", features = ["libsecp_compat", "serde"] }
 ed25519-dalek = { version = "1.0.0-pre.4", features = ["serde"] }# Cannot be 1 because they depend on curve25519-dalek version 3
 futures = { version = "0.3", default-features = false }
-libp2p = { version = "0.29", default-features = false, features = ["tcp-tokio", "yamux", "mplex", "dns", "noise", "request-response"] }
-libp2p-tokio-socks5 = "0.4"
+libp2p = { version = "0.34", default-features = false, features = ["tcp-tokio", "yamux", "mplex", "dns", "noise", "request-response"] }
 log = { version = "0.4", features = ["serde"] }
 miniscript = { version = "4", features = ["serde"] }
 monero = { version = "0.9", features = ["serde_support"] }
@@ -30,7 +29,7 @@ monero-harness = { path = "../monero-harness" }
 pem = "0.8"
 prettytable-rs = "0.8"
 rand = "0.7"
-reqwest = { version = "0.10", default-features = false, features = ["socks"] }
+reqwest = { version = "0.11", default-features = false }
 rust_decimal = "1.8"
 serde = { version = "1", features = ["derive"] }
 serde_cbor = "0.11"
@@ -43,7 +42,7 @@ strum = { version = "0.20", features = ["derive"] }
 tempfile = "3"
 thiserror = "1"
 time = "0.2"
-tokio = { version = "0.2", features = ["rt-threaded", "time", "macros", "sync"] }
+tokio = { version = "1.0", features = ["rt-multi-thread", "time", "macros", "sync"] }
 tracing = { version = "0.1", features = ["attributes"] }
 tracing-core = "0.1"
 tracing-futures = { version = "0.2", features = ["std-future", "futures-03"] }
@@ -55,7 +54,7 @@ void = "1"
 
 [dev-dependencies]
 get-port = "3"
-hyper = "0.13"
+hyper = "0.14"
 port_check = "0.1"
 serde_cbor = "0.11"
 spectral = "0.6"
--- a/swap/src/config.rs
+++ b/swap/src/config.rs
@@ -16,9 +16,23 @@ pub struct Config {
     pub monero_network: monero::Network,
 }
 
-impl Config {
-    pub fn mainnet() -> Self {
-        Self {
+// TODO: This trait is not needed
+pub trait GetConfig {
+    fn get_config() -> Config;
+}
+
+#[derive(Clone, Copy)]
+pub struct Mainnet;
+
+#[derive(Clone, Copy)]
+pub struct Testnet;
+
+#[derive(Clone, Copy)]
+pub struct Regtest;
+
+impl GetConfig for Mainnet {
+    fn get_config() -> Config {
+        Config {
             bob_time_to_act: *mainnet::BOB_TIME_TO_ACT,
             bitcoin_finality_confirmations: mainnet::BITCOIN_FINALITY_CONFIRMATIONS,
             bitcoin_avg_block_time: *mainnet::BITCOIN_AVG_BLOCK_TIME,
@@ -29,9 +43,11 @@ impl Config {
             monero_network: monero::Network::Mainnet,
         }
     }
+}
 
-    pub fn testnet() -> Self {
-        Self {
+impl GetConfig for Testnet {
+    fn get_config() -> Config {
+        Config {
             bob_time_to_act: *testnet::BOB_TIME_TO_ACT,
             bitcoin_finality_confirmations: testnet::BITCOIN_FINALITY_CONFIRMATIONS,
             bitcoin_avg_block_time: *testnet::BITCOIN_AVG_BLOCK_TIME,
@@ -42,9 +58,11 @@ impl Config {
             monero_network: monero::Network::Stagenet,
         }
     }
+}
 
-    pub fn regtest() -> Self {
-        Self {
+impl GetConfig for Regtest {
+    fn get_config() -> Config {
+        Config {
             bob_time_to_act: *regtest::BOB_TIME_TO_ACT,
             bitcoin_finality_confirmations: regtest::BITCOIN_FINALITY_CONFIRMATIONS,
             bitcoin_avg_block_time: *regtest::BITCOIN_AVG_BLOCK_TIME,
--- a/swap/src/config.rs
+++ b/swap/src/config.rs
@@ -104,7 +104,7 @@ mod regtest {
 
     pub static MONERO_FINALITY_CONFIRMATIONS: u32 = 1;
 
-    pub static BITCOIN_CANCEL_TIMELOCK: Timelock = Timelock::new(50);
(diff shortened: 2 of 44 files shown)


swap/src/protocol/alice.rs
```diff
@@ -168,27 +168,47 @@ impl Behaviour {
...
     /// Alice always sends her messages as a response to a request from Bob.
-    pub fn send_amounts(&mut self, channel: ResponseChannel<AliceToBob>, amounts: SwapAmounts) {
+    pub fn send_amounts(
+        &mut self,
+        channel: ResponseChannel<AliceToBob>,
+        amounts: SwapAmounts,
+    ) -> Result<()> {
         let msg = AliceToBob::Amounts(amounts);
-        self.amounts.send(channel, msg);
+        let _ = self.amounts.send(channel, msg);
```
D4nte, line 177:
```suggestion
        self.amounts.send(channel, msg)?;
```
Any reason why you silence the error?

swap/src/protocol/alice.rs
```diff
@@ -168,27 +168,47 @@ impl Behaviour {
...
     /// Send Message0 to Bob in response to receiving his Message0.
-    pub fn send_message0(&mut self, channel: ResponseChannel<AliceToBob>, msg: Message0) {
-        self.message0.send(channel, msg);
-        debug!("Sent Message0");
+    pub fn send_message0(
+        &mut self,
+        channel: ResponseChannel<AliceToBob>,
+        msg: Message0,
+    ) -> Result<()> {
+        let _ = self.message0.send(channel, msg);
```
D4nte, line 188:
```suggestion
        self.message0.send(channel, msg)?;
```

swap/src/protocol/alice.rs
```diff
@@ -168,27 +168,47 @@ impl Behaviour {
...
     /// Send Message1 to Bob in response to receiving his Message1.
-    pub fn send_message1(&mut self, channel: ResponseChannel<AliceToBob>, msg: Message1) {
-        self.message1.send(channel, msg);
-        debug!("Sent Message1");
+    pub fn send_message1(
+        &mut self,
+        channel: ResponseChannel<AliceToBob>,
+        msg: Message1,
+    ) -> Result<()> {
+        let _ = self.message1.send(channel, msg);
```
D4nte, line 199:
```suggestion
        self.message1.send(channel, msg)?;
```

swap/src/protocol/alice.rs
```diff
@@ -168,27 +168,47 @@ impl Behaviour {
...
     /// Send Message2 to Bob in response to receiving his Message2.
-    pub fn send_message2(&mut self, channel: ResponseChannel<AliceToBob>, msg: Message2) {
-        self.message2.send(channel, msg);
-        debug!("Sent Message2");
+    pub fn send_message2(
+        &mut self,
+        channel: ResponseChannel<AliceToBob>,
+        msg: Message2,
+    ) -> Result<()> {
+        let _ = self.message2.send(channel, msg);
```
D4nte, line 210:
```suggestion
        self.message2.send(channel, msg)?;
```

swap/src/protocol/bob.rs
```diff
@@ -149,33 +149,33 @@ impl Behaviour {
...
-        let _id = self.amounts.request_amounts(alice.clone(), btc);
+        let _id = self.amounts.request_amounts(alice, btc);
         info!("Requesting amounts from: {}", alice);
     }
 
     /// Sends Bob's first message to Alice.
     pub fn send_message0(&mut self, alice: PeerId, msg: bob::Message0) {
         self.message0.send(alice, msg);
-        debug!("Sent Message0");
+        info!("Sent Message0");
```
D4nte, line 159: Any reason to upgrade this? I think `debug` is appropriate, a user does not really care about that.

D4nte (changes requested):
You currently silence failure to send, please make it return an error to avoid the program just staling.

thomaseizinger:
Unless each commit compiles, I'd squash them all together.

swap/src/protocol/alice.rs
rishflab: yep this is incorrect, will fix

swap/src/protocol/bob.rs
rishflab: I will remove this commit. The amounts message is `info` for some reason, the rest are debug.

swap/src/monero/wallet.rs
```diff
@@ -117,46 +116,45 @@ impl WatchForTransfer for Wallet {
         let wallet = self.inner.clone();
 
         let confirmations = Arc::new(AtomicU32::new(0u32));
-        let res = (move || {
+
```
rishflab, line 119: I got this compiling but I am not sure if I changed the behaviour

swap/src/protocol/alice/message0.rs
```diff
@@ -49,9 +50,11 @@ pub struct Behaviour {
 }
 
 impl Behaviour {
-    pub fn send(&mut self, channel: ResponseChannel<AliceToBob>, msg: Message0) {
+    pub fn send(&mut self, channel: ResponseChannel<AliceToBob>, msg: Message0) -> Result<()> {
         let msg = AliceToBob::Message0(Box::new(msg));
-        self.rr.send_response(channel, msg);
+        self.rr
+            .send_response(channel, msg)
+            .map_err(|_| anyhow!("Sending Amounts response failed"))
```
thomaseizinger, line 57: Did you try using `.context` instead? That is usually preferable unless the error returned doesn't implement `std::error::Error`.

```suggestion
            .context("Sending Amounts response failed")
```

swap/src/monero/wallet.rs
thomaseizinger: Looks identical to me.

thomaseizinger: If this compiles, you can likely remove all the clones in the beginning now.

swap/src/protocol/alice.rs
```diff
@@ -308,26 +308,42 @@ impl Behaviour {
         &mut self,
         channel: ResponseChannel<AliceToBob>,
         swap_response: SwapResponse,
-    ) {
-        self.amounts.send(channel, swap_response);
+    ) -> Result<()> {
+        let _ = self.amounts.send(channel, swap_response)?;
```
thomaseizinger, line 312: I suspect the return value of `send` is `Result<(), _>`? Using `_` for `()` is a bit pointless :)

swap/src/protocol/alice/event_loop.rs
```diff
@@ -228,24 +227,24 @@ impl EventLoop {
                         }
                     }
                 },
-                swap_response = self.send_swap_response.next().fuse() => {
+                swap_response = self.send_swap_response.recv().fuse() => {
                     if let Some((channel, swap_response)) = swap_response  {
-                        self.swarm.send_swap_response(channel, swap_response);
+                        self.swarm.send_swap_response(channel, swap_response).expect("Could not send amounts message");
```
thomaseizinger, line 232: I doubt that panicking here is the desired long-term strategy. That will tear-down the whole event loop.

thomaseizinger: Even if tear-down is desired, you probably just want to return an error from the whole function :)

rishflab: This code is executed in a loop inside the event_loop.run() function. We spawn it into an async task. How does this interact with returning an error? Does it stop looping executing and return the error?

swap/src/protocol/alice.rs
rishflab: There is a log statement after amounts.send() which I will have to get rid of if I get rid of the pointless binding. I didn't want to make that change in an upgrade PR.

thomaseizinger: That is not what I meant. What I meant is keep the `?` and just remove the binding:

```suggestion
        self.amounts.send(channel, swap_response)?;
```

swap/src/protocol/alice/event_loop.rs
thomaseizinger: Yes, it will break the loop and end the task. If you are `await`ing the `JoinHandle` that you get returned from `spawn`ing the task somewhere, then that will resolve to the `Err`.

swap/src/protocol/alice/message0.rs
D4nte: > Did you try using `.context` instead? That is usually preferable unless the error returned doesn't implement `std::error::Error`.

I don't think you can use context here because the error is an `AliceToBob` enum (the message that failed to send).

swap/src/protocol/alice/event_loop.rs
D4nte: Not sure if we discussed that after or before this change but best for now to log any error you encounter and let the loop run.

da-kami:
@rishflab did you keep records of the CI runs by any chance? Is it always the same test that fails or is it different ones?

swap/tests/testutils/mod.rs
```diff
@@ -302,7 +304,7 @@ impl TestContext {
     }
 }
 
-pub async fn setup_test<T, F>(testfn: T)
+pub async fn setup_test<T, F>(override_config: Option<Config>, testfn: T)
```
thomaseizinger, line 307: If you change `override_config` to be of type `Option<O>` and add `where O: GetConfig` then you don't have to call `get_config()` on the outside but can just call it in here by doing:

```
let config = override_config.unwrap_or(config::Regtest)::get_config()
```

You might need to change the associated function to a method for this to work (i.e. make `get_config` take `self`)

D4nte: no need to make it a method, just need to call `O::get_config()` :)

swap/Cargo.toml
```diff
@@ -10,27 +10,26 @@ anyhow = "1"
 async-recursion = "0.3.1"
 async-trait = "0.1"
 atty = "0.2"
-backoff = { version = "0.2", features = ["tokio"] }
+backoff = { git = "https://github.com/ihrwein/backoff", rev = "9d03992a83dfdc596be26276d4e5c5254a4b11a2", features = ["tokio"] }
```
da-kami, line 13: There is no released version to depend on yet?
(Just asking to make sure we depend on the commit hash on purpose and it was not for testing.)

swap/src/config.rs
```diff
@@ -104,7 +122,7 @@ mod regtest {
 
     pub static MONERO_FINALITY_CONFIRMATIONS: u32 = 1;
 
-    pub static BITCOIN_CANCEL_TIMELOCK: Timelock = Timelock::new(50);
+    pub static BITCOIN_CANCEL_TIMELOCK: Timelock = Timelock::new(100);
```
da-kami, line 125: Do we know what this has to be higher now (compared to master, where the tests pass)?

rishflab:
Cant give a GitHub approval but keen to get this merge.

swap/src/config.rs
```diff
@@ -16,9 +16,23 @@ pub struct Config {
     pub monero_network: monero::Network,
 }
 
-impl Config {
-    pub fn mainnet() -> Self {
-        Self {
+// TODO: This trait is not needed
+pub trait GetConfig {
```
da-kami, line 20: Why are there traits for the config? 
Shouldn't it be enough to create static test configs that are passed in?

swap/src/config.rs
```diff
@@ -16,9 +16,23 @@ pub struct Config {
     pub monero_network: monero::Network,
 }
 
-impl Config {
-    pub fn mainnet() -> Self {
-        Self {
+// TODO: This trait is not needed
```
da-kami, line 19: Is this TODO correct? The test configuration seems to depend on the trait, e.g.:

```

impl GetConfig for SlowCancelConfig {
    fn get_config() -> Config {
        Config {
            bitcoin_cancel_timelock: Timelock::new(180),
            ..config::Regtest::get_config()
        }
    }
}
```

D4nte: Indeed, missed that.

swap/src/config.rs
D4nte: Because I don't want to introduced `Fast/Slow` config swap release code.

swap/src/config.rs
D4nte: It's not relevant anymore as we use custom times for each test.

swap/Cargo.toml
D4nte: @rishflab ?

PR #144, opened 2021-01-15
Test refactor

da-kami:
This PR is pure refactoring, keeping the logic of the tests we had before. No production code is touched besides re-exports in early commits (no logic changes).

In the follow ups improvements will be introduced, that touch the production code as well.

All remaining tasks actioned since Friday: 

- [x] `happy_path_bob _restart` (trivial)
- [x] add refund assertions to harnesses (trivial)
- [x] convert all refund scenarios currently being tested (trivial)
- [x] remove dead test init code once all old tests are converted
- [ ] ~~(optional) move alice and bob harness code into separate files~~ -> might action this once re-using test code in production.

Out of scope, follow up:
- [x] https://github.com/comit-network/xmr-btc-swap/pull/145 - We can do exact assertions for Bob's redeem as well, but have to store Bob's `tx_lock` id in the respective final state. Make `tx_lock` available in `BtcRedeemed` and `BtcPunished` to have better assertions / harmonize test behaviour. 
- [ ] update the production code to use the `Alice` and `Bob` structs to bundle the params - update tests to use the production struct.
- [ ] Re-use test swap setup in production (i.e. `Alice-/BobHarness::new`) to setup the swap.
- [ ] add additional tests
- [ ] re-try moving the tests from `test` to `src` (if the peer_id was the only problem this should be trivial now - but should be done after the refactor is finished)
- [ ] creating new wallets upon restart
- [ ] aborting the old event loop after restart

### DIFF
--- a/swap/tests/happy_path.rs
+++ b/swap/tests/happy_path.rs
@@ -1,21 +1,9 @@
-use crate::testutils::{init_alice, init_bob};
-use futures::{
-    future::{join, select},
-    FutureExt,
-};
-use get_port::get_port;
-use libp2p::Multiaddr;
 use rand::rngs::OsRng;
 use swap::{
     bitcoin,
-    config::Config,
-    monero,
-    protocol::{alice, bob},
-    seed::Seed,
+    protocol::{alice, alice::AliceState, bob, bob::BobState},
 };
-use testcontainers::clients::Cli;
-use testutils::init_tracing;
-use uuid::Uuid;
+use tokio::join;
 
 pub mod testutils;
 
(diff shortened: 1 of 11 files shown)


rishflab (approved):

swap/tests/happy_path.rs
```diff
@@ -1,130 +1,40 @@
...
+        let bob_swap = bob::swap(
+            bob.state,
+            bob.event_loop_handle,
+            bob.db,
+            bob.bitcoin_wallet.clone(),
+            bob.monero_wallet.clone(),
+            OsRng,
+            bob.swap_id,
+        );
+        let (alice_state, bob_state) = join!(alice_swap, bob_swap);
```
D4nte, line 34: nice! 🚀

D4nte (approved):

PR #145, opened 2021-01-18
Make lock-tx id available in redeem/punish state to be able to assert exact fees

da-kami:
We can do exact assertions for Bob's redeem as well, but have to store Bob's tx_lock id in the respective final state. Make tx_lock available in BtcRedeemed and BtcPunished to have better assertions / harmonize test behaviour.

Storing this information is strictly speaking not needed for the production environment. But it is static information that can be seen as additional information that can be handy for a user. We could potentially extract it inside the tests as well (for redeem without restart would be a bit tricky), but I think this solution is more elegant.

### DIFF
--- a/swap/src/protocol/bob/state.rs
+++ b/swap/src/protocol/bob/state.rs
@@ -35,8 +35,8 @@ pub enum BobState {
     CancelTimelockExpired(State4),
     BtcCancelled(State4),
     BtcRefunded(State4),
-    XmrRedeemed,
-    BtcPunished,
+    XmrRedeemed(State6),
+    BtcPunished(State6),
     SafelyAborted,
 }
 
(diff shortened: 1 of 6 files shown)


swap/src/protocol/bob/state.rs
```diff
@@ -35,8 +35,8 @@ pub enum BobState {
     CancelTimelockExpired(State4),
     BtcCancelled(State4),
     BtcRefunded(State4),
-    XmrRedeemed,
-    BtcPunished,
+    XmrRedeemed(State6),
```
D4nte, line 38: I don't think we should define a `State6`. Let me try something locally see if it looks better.

D4nte: @da-kami https://github.com/comit-network/xmr-btc-swap/pull/146

D4nte (approved):
Looks good but I think `State6` is an unnecessary abstraction https://github.com/comit-network/xmr-btc-swap/pull/146

Issue #131, opened 2021-01-11
[SP] Automated Swap Backend


D4nte:
The Service Provider (SP) also runs a backend that has the role to execute swaps. The backend...

- [x] https://github.com/comit-network/xmr-btc-swap/pull/174 WIP - automatically executes a swap when reached via the p2p interface.
- [x] gets any related configuration items for a config file, the config file has reasonable defaults.
- [ ] provides an API to allow the website to access the current rate.
- [ ] uses kraken mid-market rate as rate.
- [ ] uses an up-to-date rate when locking the Bitcoin asset.
- [ ] does not proceed with swap execution if predicted Bitcoin quantity to sell is above a configurable ratio of total balance.
- [ ] uses bitcoind wallet as Bitcoin Wallet.
- [x] uses monero-wallet-rpc as Monero Wallet.
- [ ] prints Bitcoin cancellation, Bitcoin punishment transactions when computed
- [ ] prints Monero (refund) private key when the protocol allows (Bitcoin refunded)
- [ ] can run several swaps concurrently.
- [ ] Bonus: bundles the swap website, allowing the SP operator to just proxy from their favourite webserver.

da-kami:
Notes on `can run swaps concurrently`, after working on https://github.com/comit-network/xmr-btc-swap/pull/147 I see the following issues to be resolved to make concurrent swaps happen:

* Separate `EventLoop` and `EventLoopHandle` - I think we will need this to make concurrent swaps possible. We should be able to drive a swap through a handle without creating a new `EventLoop`. I think the `EventLoop` should live in the factory, while the handles live in each swap. (Currently the `EventLoop` is returned upon creating a swap.)
* Don't just have one `swap_id` in the `Factory` but have a list of swaps.
* Make swaps identifiable in the network communication.

D4nte:
Regarding `can run swaps concurrently`, it also means concurrent access to the DB. I think we should do #133 to not end up fixing the db concurrent access twice.

thomaseizinger:
Closing because epic from previous project.

PR #147, opened 2021-01-18
Refactor prod code after test refactoring

da-kami:
Follow up of https://github.com/comit-network/xmr-btc-swap/pull/144

Took a bit more than expected, but this is really neat now!
The commits should be well-contained for reviewing, but https://github.com/comit-network/xmr-btc-swap/commit/00835baa154ddc278194fd9b98ffbfd8225e1c76 is quite big. 

I changed the abstraction on the way - out methods are finally named `run` and `run_until` which both take a swap - which makes way more sense I think :)

Also had to change the abstraction layers in `testutils` and introduced `Test` which specifies the swap amounts (that would usually come from the commandline and should not live in the factory as they are irrelevant for resumed swaps).

### DIFF
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -292,13 +292,16 @@ async fn alice_swap(
     db: Database,
     seed: Seed,
 ) -> Result<AliceState> {
-    let alice_behaviour = alice::Behaviour::new(network::Seed::new(seed));
-    let alice_peer_id = alice_behaviour.peer_id();
-    info!("Own Peer-ID: {}", alice_peer_id);
-    let alice_transport = build(alice_behaviour.identity())?;
+    let identity = network::Seed::new(seed).derive_libp2p_identity();
+
+    let peer_id = identity.public().into_peer_id();
+
+    let alice_behaviour = alice::Behaviour::default();
+    info!("Own Peer-ID: {}", peer_id);
+    let alice_transport = build(identity)?;
 
     let (mut event_loop, handle) =
-        alice::event_loop::EventLoop::new(alice_transport, alice_behaviour, listen_addr)?;
+        alice::event_loop::EventLoop::new(alice_transport, alice_behaviour, listen_addr, peer_id)?;
 
     let swap = alice::Swap {
         state,
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -23,7 +23,7 @@ use swap::{
     config::Config,
     database::Database,
     monero,
-    protocol::{alice, bob, bob::BobSwapFactory},
+    protocol::{alice, bob, bob::SwapFactory},
     trace::init_tracing,
     StartingBalances, SwapAmounts,
 };
@@ -82,7 +82,7 @@ async fn main() -> Result<()> {
                 send_monero, receive_bitcoin, swap_id
             );
 
-            let alice_factory = alice::AliceSwapFactory::new(
+            let alice_factory = alice::SwapFactory::new(
                 seed,
                 config,
                 swap_id,
@@ -127,7 +127,7 @@ async fn main() -> Result<()> {
                 send_bitcoin, receive_monero, swap_id
             );
 
-            let bob_factory = BobSwapFactory::new(
+            let bob_factory = SwapFactory::new(
                 seed,
                 db_path,
                 swap_id,
@@ -172,7 +172,7 @@ async fn main() -> Result<()> {
             )
             .await?;
 
-            let alice_factory = alice::AliceSwapFactory::new(
+            let alice_factory = alice::SwapFactory::new(
                 seed,
                 config,
                 swap_id,
@@ -204,7 +204,7 @@ async fn main() -> Result<()> {
             )
             .await?;
 
-            let bob_factory = BobSwapFactory::new(
+            let bob_factory = SwapFactory::new(
                 seed,
                 db_path,
                 swap_id,
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -14,20 +14,16 @@
 #![allow(non_snake_case)]
 
 use crate::cli::{Command, Options, Resume};
-use anyhow::{bail, Context, Result};
-use libp2p::{core::Multiaddr, PeerId};
+use anyhow::{Context, Result};
 use prettytable::{row, Table};
-use rand::rngs::OsRng;
 use std::sync::Arc;
 use structopt::StructOpt;
 use swap::{
     bitcoin,
     config::Config,
-    database::{Database, Swap},
-    monero, network,
-    network::transport::build,
-    protocol::{alice, bob, bob::BobState},
-    seed::Seed,
+    database::Database,
+    monero,
+    protocol::{alice, bob, bob::BobSwapFactory},
     trace::init_tracing,
     StartingBalances, SwapAmounts,
 };
@@ -52,8 +48,6 @@ async fn main() -> Result<()> {
     );
     let data_dir = std::path::Path::new(opt.data_dir.as_str()).to_path_buf();
     let db_path = data_dir.join("database");
-
-    let db = Database::open(db_path.as_path()).context("Could not open database")?;
 
     let seed = swap::config::seed::Seed::from_file_or_generate(&data_dir)
         .expect("Could not retrieve/initialize seed")
@@ -83,6 +77,11 @@ async fn main() -> Result<()> {
 
             let swap_id = Uuid::new_v4();
 
+            info!(
+                "Swap sending {} and receiving {} starting with ID {}",
+                send_monero, receive_bitcoin, swap_id
+            );
+
             let alice_factory = alice::AliceSwapFactory::new(
                 seed,
                 config,
@@ -108,7 +107,12 @@ async fn main() -> Result<()> {
             send_bitcoin,
             receive_monero,
         } => {
-            let (bitcoin_wallet, monero_wallet, _) = setup_wallets(
+            let swap_amounts = SwapAmounts {
+                btc: send_bitcoin,
+                xmr: receive_monero,
+            };
+
+            let (bitcoin_wallet, monero_wallet, starting_balances) = setup_wallets(
                 bitcoind_url,
                 bitcoin_wallet_name.as_str(),
                 monero_wallet_rpc_url,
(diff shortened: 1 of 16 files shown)


swap/src/protocol/alice/swap.rs
```diff
@@ -81,10 +64,31 @@ pub fn is_encsig_learned(state: &AliceState) -> bool {
...
+        swap.db,
+    )
+    .await
+}
+
 // State machine driver for swap execution
 #[async_recursion]
 #[allow(clippy::too_many_arguments)]
-pub async fn run_until(
+async fn do_run_until(
```
rishflab, line 91: NITPICK: why do_run_until instead of run_until? It seems like an unnecessary addition of a word

swap/src/protocol/alice.rs
```diff
@@ -39,29 +40,170 @@ pub mod state;
 mod steps;
 pub mod swap;
 
-pub type Swarm = libp2p::Swarm<Behaviour>;
+pub struct Swap {
```
rishflab, line 43: I wondering if it would be possible to change the DB to a reference here so we can support concurrent swaps (using the same DB).

rishflab (approved):
Approved but I think we should of attempted #131 first to gain some context on what the production abstractions should look like. I think they will change again once we try to support concurrent swaps.

swap/src/protocol/alice.rs
da-kami: I would record that as 

> Approved but I think we should of attempted #131 first to gain some context on what the production abstractions should look like. I think they will change again once we try to support concurrent swaps.

I think the changes will fit on top of this just fine. I did think about concurrent swaps on the way, but I think we have to do several iterations anyway to level in on a suitable design. 
That said: I would merge this and then do changes on top. To my mind this is definitely an improvement to what was there before :)

da-kami:
> Approved but I think we should of attempted #131 first to gain some context on what the production abstractions should look like. I think they will change again once we try to support concurrent swaps.

I think the changes will fit on top of this just fine. I did think about concurrent swaps on the way, but I think we have to do several iterations anyway to level in on a suitable design. 
That said: I would merge this and then do changes on top. To my mind this is definitely an improvement to what was there before :)

swap/src/protocol/alice/swap.rs
da-kami: Because I added another `run_until` function that is exposed and takes the `Swap` as an argument. That way we have the internal function and a wrapper function that is exposed. I was not able to give it the same name because Rust does not support function overloading where the function would have the same name but different parameters :)

swap/src/protocol/alice.rs
da-kami: Fair point, I'd action this once we go for concurrent swaps.

swap/src/protocol/alice/swap.rs
da-kami: Note: I did try to achieve this by using the object, but ran into lifetime issues (notably mutably borrowing the event loop handle) that could not be resolved easily. I am not sure this can be easily solved due to the recursive function calls.

da-kami:
@thomaseizinger would be cool if you could have a look at this one as well. This is the basis of the improvements of https://github.com/comit-network/xmr-btc-swap/pull/150

swap/src/protocol/alice.rs
```diff
@@ -46,6 +50,162 @@ pub struct Swap {
...
+    }
+
+    pub async fn recover_alice_from_db(&self) -> Result<(Swap, EventLoop)> {
+        // reopen the existing database
+        let db = Database::open(self.db_path.clone().as_path())?;
+
+        let resume_state = if let database::Swap::Alice(state) = db.get_state(self.swap_id)? {
+            state.into()
+        } else {
+            unreachable!()
```
thomaseizinger, line 134: Why not return an error here?

swap/tests/testutils/mod.rs
```diff
@@ -320,12 +330,31 @@ where
         xmr: swap_amounts.xmr * 10,
         btc: bitcoin::Amount::ZERO,
     };
+
+    let port = get_port().expect("Failed to find a free port");
```
thomaseizinger, line 327: You could just let the OS assign you a port by passing 0 to the multi-addr. You'll need to wait for the `NewListenAddr` event then but you will probably need to do that anyway to make sure the Swarm is ready :)

swap/src/main.rs
```diff
@@ -83,6 +77,11 @@ async fn main() -> Result<()> {
 
             let swap_id = Uuid::new_v4();
 
+            info!(
+                "Swap sending {} and receiving {} starting with ID {}",
```
thomaseizinger, line 81: Nit: This message reads like the swap "starts with" the ID the same way a sentence starts with a specific word.

You might want to start thinking about tracing's [spans](https://docs.rs/tracing/0.1.22/tracing/span/index.html) and attach the ID as a context to all log messages.

Make sure to read the linked docs in detail regarding async code, there are some subtleties to it on how to use it correctly.

thomaseizinger (commented):
Looks like a good improvement to me!

Left some comments that you might find interesting :)
Thanks for splitting the PR into distinct patches, was fairly easy to review!

swap/tests/testutils/mod.rs
da-kami: Ah, so the way to do it is to listen to `NewListenAddr` - we should give this a try @rishflab ! - This might solve the dialing test failure when concurrently executing tests.

swap/src/main.rs
da-kami: Do I understand this correctly, that the span would represent the swap (i.e. the period of time one specific swap is executed.)?

thomaseizinger: Yes. You probably want to attach the span using `.instrument` to the `Future` that represents the swap execution and include the ID within that swarm. That will automatically tag all tracing events with this ID if they are logged within that `Future`.

swap/tests/testutils/mod.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/issues/151

swap/src/main.rs
da-kami: I'd keep this in mind until we actually implement concurrent swap.

swap/src/protocol/alice.rs
D4nte: > I wondering if it would be possible to change the DB to a reference here so we can support concurrent swaps (using the same DB).

At this stage we still need to understand how the db can support concurrent swaps. No need to do this work with sled. Better migrate to sqlite first.

swap/src/protocol/alice/swap.rs
```diff
@@ -81,10 +64,31 @@ pub fn is_encsig_learned(state: &AliceState) -> bool {
     )
 }
 
+pub async fn run(swap: alice::Swap) -> Result<AliceState> {
+    run_until(swap, is_complete).await
+}
+
+pub async fn run_until(
```
D4nte, line 71:
```suggestion
pub async fn run_swap_until(
```

swap/src/protocol/alice/swap.rs
```diff
@@ -81,10 +64,31 @@ pub fn is_encsig_learned(state: &AliceState) -> bool {
...
+        swap.db,
+    )
+    .await
+}
+
 // State machine driver for swap execution
 #[async_recursion]
 #[allow(clippy::too_many_arguments)]
-pub async fn run_until(
+async fn run_until_internal(
```
D4nte, line 91:
```suggestion
async fn run_until(
```
I understand why you want to use `internal` but it's not a commonly used terminology. Maybe for for the 2 name changes I propose or `_run_until`.

swap/src/protocol/alice/swap.rs
```diff
@@ -110,9 +114,9 @@ pub async fn run_until(
                 };
 
                 let db_state = (&state).into();
-                db.insert_latest_state(swap_id, Swap::Alice(db_state))
+                db.insert_latest_state(swap_id, database::Swap::Alice(db_state))
                     .await?;
-                run_until(
+                run_until_internal(
```
D4nte, line 119: Is this massive rename necessary? I don't understand what is the added value.

D4nte (approved):
Looks good.

As discussed, there is a number of improvements that can be done as shown in #150.

Also, the title does not represent the added-value of the PR `Refactor prod code after test refactoring`. It would be good to put in the title what improvement you are doing with the refactor.
_Refactor 'cause I refactored_ sounds like mindless code churn 🙃

swap/src/protocol/alice/swap.rs
da-kami: there are two `run_until` functions now. One is exposed and the other one isn't. The one exposed is:

```
pub async fn run_until(
    swap: alice::Swap,
    is_target_state: fn(&AliceState) -> bool,
)
```

it acts as a wrapper for the internal one is: 

```
async fn run_until_internal(
    state: AliceState,
    is_target_state: fn(&AliceState) -> bool,
    mut event_loop_handle: EventLoopHandle,
    bitcoin_wallet: Arc<bitcoin::Wallet>,
    monero_wallet: Arc<monero::Wallet>,
    config: Config,
    swap_id: Uuid,
    db: Database,
)
```

@rishflab and @thomaseizinger did not like the name `do_run_until` for the internal one so I changed it to `run_until_internal`. I don't think it is very important to agree on the perfect name for it. Happy for follow up PRs on that one :)

swap/src/protocol/alice/swap.rs
da-kami: I think `run` and `run_until` are fine. The swap is the first parameter. That implies that we are running the swap.

swap/src/protocol/bob.rs
```diff
@@ -38,20 +38,166 @@ mod message3;
...
+            alice_connect_address,
+            alice_connect_peer_id,
+            bitcoin_wallet,
+            monero_wallet,
+            starting_balances,
+        }
+    }
+
+    pub async fn new_swap_as_bob(
+        &self,
```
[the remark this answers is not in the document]
da-kami: This change is not easily possible in the current version, because the `SwapFactory` is actually reused. The boundaries of the factory are not defined well (it should not be used for assertions in `Test` but `Test` should store the necessary information itself). I'd say we solve this in #150

swap/src/protocol/alice.rs
```diff
@@ -1,22 +1,18 @@
...
 use anyhow::Result;
-use libp2p::{
-    core::{identity::Keypair, Multiaddr},
-    request_response::ResponseChannel,
-    NetworkBehaviour, PeerId,
-};
+use libp2p::{request_response::ResponseChannel, NetworkBehaviour, PeerId};
 use tracing::{debug, info};
 
 use crate::{
```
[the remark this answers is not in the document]
da-kami: The problem of having a newline between `use` and `use crate` statements is all over the codebase. I don't think I introduced this problem. Happy to fix this with a follow up globally.

PR #148, opened 2021-01-19
Wait for Bob's refund finality

da-kami:
For Alice we ensure to wait for redeem/punish finality, so it should be the same for Bob on Bitcoin.

Notes:
Arguably we could also skip this and transition to `Redeemed` once we have the tx in the mempool. (so the user does not have to wait for the confirmations but monitors the tx finality himself.) 
But if we opt for that resume won't trigger the transaction again. Additionally we should not have different behavior for Alice and Bob I feel.

### DIFF
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -135,11 +135,9 @@ async fn main() -> Result<()> {
                 Arc::new(monero_wallet),
                 alice_addr,
                 alice_peer_id,
+                config,
             );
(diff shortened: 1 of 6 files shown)


da-kami:
After taking another look at Alice's code I noticed that we actually wait for transaction finality in production for her.
If we opt to do that we should do the same for Bob! 
I will adapt this accordingly.

rishflab (approved):

Issue #137, opened 2021-01-12
Problems to be fixed after manual tests on test-/stagenet


da-kami:
The following issues have to be fixed by adding `e2e` tests for edge cases:

- [x] Bob does not record Alice's lock proof (message 2) by saving a separate state after receiving Message2. So, when restarted Bob waits for the lock proof again, which never arrives. The time-window of this failure is closed once XMR lock transactions has the necessary confirmations and Bob transitions to XmrLocked.
- [x] https://github.com/comit-network/xmr-btc-swap/pull/149 - After EncSigLearned Alice may fail when trying to redeem. There is a bug where we always publish the redeem tx on her side in that state without taking timelocks in account (Bob already moved and cancelled, Alice redeem always fails, she never transitions and never refunds / punishes). 

Once those two issues above are done we should add more `e2e` tests for remaining edge cases (stop application after each state and see if the application behaves as expected after resuming).

Details to failure scenarios can be found in https://github.com/comit-network/xmr-btc-swap/issues/135

Note: Scenarios involving the `cancel` path should not be actioned because for the MVP we will focus on manual recovery.

da-kami:
I will start with working on 

* After EncSigLearned Alice may fail when trying to redeem. There is a bug where we always publish the redeem tx on her side in that state without taking timelocks in account (Bob already moved and cancelled, Alice redeem always fails, she never transitions and never refunds / punishes).

PR #149, opened 2021-01-19
Fix Alice redeem scenario

da-kami:
Follow up of #144, partial fix of https://github.com/comit-network/xmr-btc-swap/issues/137

 Fix Alice redeem scenario

- Properly check the timelocks before trying to redeem
- Distinguish different failure scenarios and reactions to it.
    - if we fail to construct the redeem transaction: wait for cancel.
    - if we fail to publish the redeem transaction: wait for cancel but let the user know that restarting the application will result in retrying to publish the tx.
    - if we succeed to publish the tx but then fail when waiting for finality, print error to the user (secreat already leaked, the user has to check manually if the tx was included)

### DIFF
--- a/swap/src/protocol/alice/swap.rs
+++ b/swap/src/protocol/alice/swap.rs
@@ -1,6 +1,6 @@
 //! Run an XMR/BTC swap in the role of Alice.
 //! Alice holds XMR and wishes receive BTC.
-use anyhow::Result;
+use anyhow::{bail, Result};
 use async_recursion::async_recursion;
 use futures::{
     future::{select, Either},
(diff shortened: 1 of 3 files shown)


swap/src/protocol/alice/swap.rs
```diff
@@ -1,19 +1,19 @@
...
     future::{select, Either},
     pin_mut,
 };
 use rand::{CryptoRng, RngCore};
 use std::sync::Arc;
-use tracing::info;
+use tracing::{error, info};
 use uuid::Uuid;
 
 use crate::{
```
D4nte, line 14:
```suggestion
use uuid::Uuid;
use crate::{
```

This newline is problematic for IDE. Let's follow `rustfmt`'s output (which does **not** separate the imports).

D4nte (approved):

rishflab (approved):

swap/src/protocol/alice/swap.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/153

da-kami:
> Timed out.

Huh? The build actually succeeded:https://app.bors.tech/batches/119595 
Will re-trigger.

da-kami:
bors retry

PR #150, opened 2021-01-19
Few improvements

D4nte:


D4nte:
Curious to know what @thomaseizinger thinks about 82e4c50

da-kami:
I am not sure about https://github.com/comit-network/xmr-btc-swap/pull/150/commits/82e4c506bcadb6da37904f2d74564f416a6de291 and the "factory" being consumed. In the context of concurrent swaps I would have seen the factory in the role of producing multiple swap instances. Mentioned ideas around that here: https://github.com/comit-network/xmr-btc-swap/issues/131#issuecomment-762502947

It might be too early for designing for that. I think we can roll with the changes concerning the consumption and the single-time use. Is there a bigger motivation behind those changes @D4nte ?

swap/src/protocol/bob.rs
```diff
@@ -154,31 +150,31 @@ impl SwapFactory {
...
-        btc_to_swap,
-        xmr_to_swap,
-        config.bitcoin_cancel_timelock,
-        config.bitcoin_punish_timelock,
-        refund_address,
-        config.monero_finality_confirmations,
-    );
-
-    Ok(BobState::Started { state0, amounts })
+    async fn init_state(
```
thomaseizinger, line 176: I would call this `initial_state` (or maybe even `make_initial_state`) because that is what it returns. If something is called `init_state`, I would expect it to take `&mut self` so it can actually initialize something on `self`.

swap/src/main.rs
```diff
@@ -138,7 +138,7 @@ async fn main() -> Result<()> {
                 alice_addr,
                 alice_peer_id,
             );
-            let (swap, event_loop) = bob_factory.new_swap(swap_amounts).await?;
+            let (swap, event_loop) = bob_factory.new_swap_as_bob(swap_amounts).await?;
```
thomaseizinger: This was just renamed in the previous patch, probably something to `fixup!` here :)

swap/src/main.rs
```diff
@@ -136,7 +136,7 @@ async fn main() -> Result<()> {
                 alice_addr,
                 alice_peer_id,
             );
-            let (swap, event_loop) = bob_factory.new_swap_as_bob(swap_amounts, config).await?;
+            let (swap, event_loop) = bob_factory.new_swap(swap_amounts, config).await?;
```
thomaseizinger: And it is renamed back again :D

Also a `fixup!` I'd assume.

swap/src/main.rs
```diff
@@ -249,10 +243,5 @@ async fn setup_wallets(
     );
     let monero_wallet = Arc::new(monero_wallet);
```
thomaseizinger, line 248: I would defer `Arc`ing things up until the point where you need it (i.e. to the first call of `clone`). This function seems so know too much about what is happening afterwards :)

D4nte:
> I am not sure about https://github.com/comit-network/xmr-btc-swap/pull/150/commits/82e4c506bcadb6da37904f2d74564f416a6de291 and the "factory" being consumed. In the context of concurrent swaps I would have seen the factory in the role of producing multiple swap instances. Mentioned ideas around that here: https://github.com/comit-network/xmr-btc-swap/issues/131#issuecomment-762502947
> 
> It might be too early for designing for that. I think we can roll with the changes concerning the consumption and the single-time use. Is there a bigger motivation behind those changes @D4nte ?

The current design is not appropriate for multiple swaps as the factory has a swap id as field.
Alice swap builder could become a factory in the future for nectar but further changes would be needed. 

I believe the current appropriate design is builder.

swap/tests/testutils/mod.rs
```diff
@@ -28,12 +28,12 @@ pub struct StartingBalances {
 pub struct Test {
     swap_amounts: SwapAmounts,
 
-    alice_swap_factory: Option<alice::SwapFactory>,
+    alice_swap_factory: Option<alice::Builder>,
```
thomaseizinger, line 31: Agree with the rename if it is only useful once. However, having an `Option` so you can `take` it is seems kind of weird.

You could implement `Clone` on the builder to make new instances without having to `take` it.

Also, this `Test` struct seems to be doing way too many things. Why not just create more entrypoins on `testutils` for the different kinds of tests there are? You can then customize the parameters that are passed to the closure based on the different test scenarios (happy path, resumable, etc). That would drive down the complexity of this struct and save you LoC in the test because you directly receive the arguments you need.

I prefer duplication in tests if it helps readability. A test is worth nothing if I cannot judge whether it is testing the right thing :)

da-kami (approved):
Thanks for this follow up! 
It seems that https://github.com/comit-network/xmr-btc-swap/pull/150/commits/bc051c202f9d1980d083190014c4c43eb9ccbae6 is introducing an problem that makes certain e2e tests fail - it looks like it is related to the optional factory in `Test`.

I generally like the changes proposed, but am not totally sure about the `Builder` the way it is introduced. I think we should bring these changes in as a follow up though.
The objective was to be able to write tests in a fast way so we are able to add more tests easily. Additionally we said we want to use the setup code in production as well. Both of these objectives are achieved with https://github.com/comit-network/xmr-btc-swap/pull/147

I know that https://github.com/comit-network/xmr-btc-swap/pull/147 is not perfect, but I think it is an improvement and can be merged before changing things on top. 
I will focus on adding the additional tests on top of https://github.com/comit-network/xmr-btc-swap/pull/147 and move on to working on the features.

da-kami:
@D4nte if you think that that #147 is not mergable in it's current form please ping me. I don't want to be annoying, but I think we can do further improvements as follow ups. I don't want to be blocked on finishing the fixes and move on to features :)

D4nte:
> It seems that [bc051c2](https://github.com/comit-network/xmr-btc-swap/commit/bc051c202f9d1980d083190014c4c43eb9ccbae6) is introducing an problem that makes certain e2e tests fail - it looks like it is related to the optional factory in `Test`.

I'll investigate.

> I generally like the changes proposed, but am not totally sure about the `Builder` the way it is introduced. I think we should bring these changes in as a follow up though.

What do you mean by "the way it is introduced"?

> The objective was to be able to write tests in a fast way so we are able to add more tests easily. Additionally we said we want to use the setup code in production as well. Both of these objectives are achieved with #147
> 
> I know that #147 is not perfect, but I think it is an improvement and can be merged before changing things on top.
> I will focus on adding the additional tests on top of #147 and move on to working on the features.

I have approved #147. Indeed, the functionality is achieved. However, the PR introduced terminology which is incorrect (the `SwapFactory` is not defined nor used as per the factory design pattern).

PR #153, opened 2021-01-21
Remove newlines from import statements to avoid problems

da-kami:
Rust fmt automatically groups the imports (from top to bottom) as `pub use` `use crate` and `use`.
There is no need to introduce sections which cause annoyance when auto importing using the IDE.

### DIFF
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -1,7 +1,15 @@
+pub mod timelocks;
+pub mod transactions;
+pub mod wallet;
+
 pub use crate::bitcoin::{
     timelocks::Timelock,
     transactions::{TxCancel, TxLock, TxPunish, TxRedeem, TxRefund},
 };
+pub use ::bitcoin::{util::amount::Amount, Address, Network, Transaction, Txid};
+pub use ecdsa_fun::{adaptor::EncryptedSignature, fun::Scalar, Signature};
+pub use wallet::Wallet;
+
 use crate::{bitcoin::timelocks::BlockHeight, config::Config, ExpiredTimelocks};
 use ::bitcoin::{
     hashes::{hex::ToHex, Hash},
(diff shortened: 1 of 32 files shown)


swap/src/bitcoin.rs
```diff
@@ -1,26 +1,24 @@
...
     hashes::{hex::ToHex, Hash},
     secp256k1,
     util::psbt::PartiallySignedTransaction,
     SigHash,
 };
+pub use ::bitcoin::{util::amount::Amount, Address, Network, Transaction, Txid};
 use anyhow::{anyhow, bail, Result};
 use async_trait::async_trait;
 use ecdsa_fun::{adaptor::Adaptor, fun::Point, nonce::Deterministic, ECDSA};
+pub use ecdsa_fun::{adaptor::EncryptedSignature, fun::Scalar, Signature};
```
D4nte, line 16: I would recommend to keep the `pub use` separated as they often have a different function from `use`.
The `pub use` are about re-exporting other modules/crates items to have them on the API of the given module. Whereas `use` are about importing what is needed inside the module.

swap/src/config.rs
```diff
@@ -1,5 +1,4 @@
 pub mod seed;
-
 use crate::bitcoin::Timelock;
```
D4nte, line 2: A line separating the `pub mod` and `use` is good as they are different items.

swap/src/database.rs
```diff
@@ -5,7 +5,6 @@ use uuid::Uuid;
 
 mod alice;
 mod bob;
```
D4nte, line 7: Keep the separation.

swap/src/main.rs
```diff
@@ -12,7 +12,6 @@
 )]
 #![forbid(unsafe_code)]
 #![allow(non_snake_case)]
```
D4nte, line 14: Keep this newline

D4nte (commented):
IMHO this change is too zealous.

swap/src/config.rs
da-kami: Unfortunately it's not that simple... I played around a bit and keeping the newline between `use` and `pub use` imports can lead to the same problem we had with `use::crate`.

If we have `use` and `use::crate` \n` `pub use` then newly added `use` statements might be added to the `pub use` section by Clion.

That can be overcome by putting all `pub use` statements on the top. Would that be acceptable? 

Note: The problem only occurs with multiple `pub use` statements, I tried some auto importing in `bitcoin.rs` to explore the behavior.

da-kami:
Gave this another shot with a more structured approach https://github.com/comit-network/xmr-btc-swap/pull/153/commits/925ab01d45f68fbc064c8bf44eb697ccbec4beaa

D4nte (approved):

rishflab:
> Gave this another shot with a more structured approach, using this order with newlines in between:
> 
> ```
> pub mod ...
> 
> mod ...
> 
> pub use ...
> 
> use / use crate / use self
> ```
> 
> [925ab01](https://github.com/comit-network/xmr-btc-swap/commit/925ab01d45f68fbc064c8bf44eb697ccbec4beaa)

Won't this cause problem's when you let the IDE auto-import? It will try to but the `use::*` above the pub mode? If this compatible with IDE/formatter/IDE-auto-import then I am ok with it.

We don't have a best practice for this afaik. `comit-rs` imports are all over the place.

PR #154, opened 2021-01-21
Clean up `lib.rs`

D4nte:
Do not remove it just yet as there are still issues with the CI but proceed with some clean up.

Merging this now to avoid further merge conflicts.


da-kami (approved):
Ha! nice that it work! Happy to finally see this being resolved.

Note:
I recently re-tried pudding several tests into the same file, but then one of the two would always fail with a dialing error. I think that is what Rishab encountered when moving the tests to `src`. Seems that problem was unrelated though.

PR #155, opened 2021-01-21
Bob saves lock proof after received so he can resume swap

da-kami:
Introducing a new state was too much so I added the additional required information to the enum state rather than adding another struct. 

As discussed earlier we should cleanup the state machine (enum states vs data structs) at some point :)

### DIFF
--- a/swap/src/protocol/bob/swap.rs
+++ b/swap/src/protocol/bob/swap.rs
@@ -11,7 +11,7 @@ use async_recursion::async_recursion;
 use rand::{rngs::OsRng, CryptoRng, RngCore};
 use std::sync::Arc;
 use tokio::select;
-use tracing::{debug, info};
+use tracing::info;
 use uuid::Uuid;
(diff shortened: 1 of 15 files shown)


swap/src/protocol/bob/swap.rs
```diff
@@ -29,6 +29,10 @@ pub fn is_btc_locked(state: &BobState) -> bool {
     matches!(state, BobState::BtcLocked(..))
 }
 
+pub fn is_lock_proof_received(state: &BobState) -> bool {
```
D4nte, line 32: This is only used in tests, right? Please don't add test code to release code :)

D4nte (approved):

swap/src/protocol/bob/swap.rs
da-kami: I moved all of those `is_xxx_state` funtcions into testutils

PR #158, opened 2021-01-22
restart event loop in tests

rishflab:
Alice was attempting to create a new event loop using the same listen addr as the old one which was still running. This commit aborts the event loop before creating a new one.

### DIFF
--- a/swap/tests/testutils/mod.rs
+++ b/swap/tests/testutils/mod.rs
@@ -14,6 +14,7 @@ use swap::{
 };
 use tempfile::tempdir;
 use testcontainers::{clients::Cli, Container};
+use tokio::task::JoinHandle;
 use tracing_core::dispatcher::DefaultGuard;
 use tracing_log::LogTracer;
 use uuid::Uuid;
@@ -35,7 +36,7 @@ struct AliceParams {
 }
 
 impl AliceParams {
-    pub async fn builder(&self) -> alice::Builder {
+    pub fn builder(&self) -> alice::Builder {
         alice::Builder::new(
             self.seed,
             self.config,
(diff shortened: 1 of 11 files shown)


swap/tests/testutils/mod.rs
```diff
@@ -118,20 +122,27 @@ impl TestContext {
...
-        tokio::spawn(async move { event_loop.run().await });
+        let join_handle = tokio::spawn(async move { event_loop.run().await });
 
-        swap
+        (swap, BobEventLoopJoinHandle(join_handle))
     }
 
-    pub async fn recover_alice_from_db(&mut self) -> alice::Swap {
-        let (swap, mut event_loop) = self.alice_params.builder().await.build().await.unwrap();
+    pub async fn recover_alice_from_db(
```
D4nte, line 130:
```suggestion
    pub async fn stop_and_resume_alice_from_db(
```
1. We use `resume` terminology everywhere else, let's remove a synonym (`recover`)
2. Express the fact that we simulate stopping the software (abort event loop) as part of this function

swap/tests/testutils/mod.rs
```diff
@@ -118,20 +122,27 @@ impl TestContext {
...
+
+        let (swap, mut event_loop) = self.alice_params.builder().build().await.unwrap();
 
         tokio::spawn(async move { event_loop.run().await });
 
         swap
     }
 
-    pub async fn recover_bob_from_db(&mut self) -> bob::Swap {
+    pub async fn recover_bob_from_db(&mut self, join_handle: BobEventLoopJoinHandle) -> bob::Swap {
```
D4nte, line 143:
```suggestion
    pub async fn stop_and_resume_bob_from_db(&mut self, join_handle: BobEventLoopJoinHandle) -> bob::Swap {
```
See previous comment for explanation.

D4nte (approved):
Well done!

PR #159, opened 2021-01-22
Introduce "one shots"

D4nte:
Instead of Alice sending messages to Bob by replying on a `RequestResponse` channel, she just acks reception on the channel and open her own channel to send messages to Bob.

This means we do not need to carry the channel around anymore and instead carry Bob's peer id.

We take this opportunity to rename messages as per the sequence diagram.

At this stage we only touch message 4 & 5 as they are the one shots messages of the protocol.

Decoupling message 4 from message 3 (still current called 2) will allow the introduction of an await/async behaviour that handles message 1, 2, 3.

To introduce message 1, 2, 3 we will introduced the async/await behaviour using https://github.com/comit-network/rust-libp2p-async-await

First, we need to upgrade tokio and libp2p.

Please note that thanks to this change, Alice now uses Bob's peer id to send message meaning that recovery is improved (as before we could not save the channel to the db).

### DIFF
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -25,15 +25,17 @@ use tracing::{info, log::LevelFilter};
 use uuid::Uuid;
 
 pub mod bitcoin;
-mod cli;
 pub mod config;
 pub mod database;
-mod fs;
 pub mod monero;
 pub mod network;
 pub mod protocol;
 pub mod seed;
 pub mod trace;
+
+mod cli;
+mod fs;
+mod serde_peer_id;
 
 #[macro_use]
 extern crate prettytable;
--- a/swap/src/network/request_response.rs
+++ b/swap/src/network/request_response.rs
@@ -19,13 +19,14 @@ const BUF_SIZE: usize = 1024 * 1024;
 // Codec for each Message and a macro that implements them.
 
 /// Messages Bob sends to Alice.
+// TODO: Remove this once network changes are done
+#[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum BobToAlice {
     SwapRequest(bob::SwapRequest),
     Message0(Box<bob::Message0>),
     Message1(bob::Message1),
     Message2(bob::Message2),
-    Message5(Message5),
 }
 
 /// Messages Alice sends to Bob.
@@ -35,7 +36,19 @@ pub enum AliceToBob {
     Message0(Box<alice::Message0>),
     Message1(Box<alice::Message1>),
     Message2(alice::Message2),
-    Message5, // empty response
+}
+
+/// Messages sent from one party to the other.
+/// All responses are empty
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Request {
+    Message5(Message5),
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+/// Response are only used for acknowledgement purposes.
+pub enum Response {
+    Message5,
 }
 
 #[derive(Debug, Clone, Copy, Default)]
@@ -128,6 +141,95 @@ where
             .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
         let mut de = serde_cbor::Deserializer::from_slice(&message);
         let msg = AliceToBob::deserialize(&mut de).map_err(|e| {
+            tracing::debug!("serde read_response error: {:?}", e);
+            io::Error::new(io::ErrorKind::InvalidData, e)
+        })?;
+
+        Ok(msg)
+    }
+
+    async fn write_request<T>(
+        &mut self,
+        _: &Self::Protocol,
+        io: &mut T,
+        req: Self::Request,
+    ) -> io::Result<()>
+    where
+        T: AsyncWrite + Unpin + Send,
+    {
+        let bytes =
+            serde_cbor::to_vec(&req).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+
+        upgrade::write_one(io, &bytes).await?;
+
+        Ok(())
+    }
+
+    async fn write_response<T>(
+        &mut self,
+        _: &Self::Protocol,
+        io: &mut T,
+        res: Self::Response,
+    ) -> io::Result<()>
+    where
+        T: AsyncWrite + Unpin + Send,
+    {
+        debug!("enter write_response");
+        let bytes = serde_cbor::to_vec(&res).map_err(|e| {
+            tracing::debug!("serde write_reponse error: {:?}", e);
+            io::Error::new(io::ErrorKind::InvalidData, e)
+        })?;
+        upgrade::write_one(io, &bytes).await?;
+
+        Ok(())
+    }
+}
+
+#[derive(Clone, Copy, Debug, Default)]
+pub struct OneShotCodec<P> {
+    phantom: PhantomData<P>,
+}
+
+#[async_trait]
+impl<P> RequestResponseCodec for OneShotCodec<P>
+where
+    P: Send + Sync + Clone + ProtocolName,
+{
+    type Protocol = P;
+    type Request = Request;
+    type Response = Response;
+
+    async fn read_request<T>(&mut self, _: &Self::Protocol, io: &mut T) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        debug!("enter read_request");
+        let message = upgrade::read_one(io, BUF_SIZE)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+        let mut de = serde_cbor::Deserializer::from_slice(&message);
+        let msg = Request::deserialize(&mut de).map_err(|e| {
+            tracing::debug!("serde read_request error: {:?}", e);
+            io::Error::new(io::ErrorKind::Other, e)
+        })?;
+
+        Ok(msg)
+    }
+
+    async fn read_response<T>(
+        &mut self,
+        _: &Self::Protocol,
+        io: &mut T,
+    ) -> io::Result<Self::Response>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        debug!("enter read_response");
+        let message = upgrade::read_one(io, BUF_SIZE)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+        let mut de = serde_cbor::Deserializer::from_slice(&message);
+        let msg = Response::deserialize(&mut de).map_err(|e| {
             tracing::debug!("serde read_response error: {:?}", e);
             io::Error::new(io::ErrorKind::InvalidData, e)
         })?;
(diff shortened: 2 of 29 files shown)


swap/src/network/request_response.rs
```diff
@@ -21,21 +21,34 @@ const BUF_SIZE: usize = 1024 * 1024;
 /// Messages Bob sends to Alice.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum BobToAlice {
-    SwapRequest(bob::SwapRequest),
+    SwapRequest(Box<bob::SwapRequest>),
```
da-kami, line 24: ❓ Why is there a `Box` needed for the `SwapRequest` now?

swap/src/protocol/alice/message4.rs
```diff
@@ -0,0 +1,101 @@
...
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::VecDeque,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tracing::error;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Message4 {
```
da-kami, line 22: 🙃 What do you think about naming these new messages in a more semantic way? 
I would find that way better for understanding. 
e.g.:

`LockTransferProofMessage` (or similar) for Alice `Message4`
`EncryptionSignatureMessage` (or similar) for Bob `Message5`

(If we do it would apply to the corresponding protocol names - e.g. `Message4Protocol` as well)

swap/src/network/request_response.rs
```diff
@@ -171,3 +193,92 @@ where
...
     }
 }
+
+#[derive(Clone, Copy, Debug, Default)]
+pub struct OneShotCodec<P> {
+    phantom: PhantomData<P>,
+}
+
+#[async_trait]
+impl<P> RequestResponseCodec for OneShotCodec<P>
```
da-kami, line 203: ❓We did the same in `comit-rs` - I am not sure why this is needed though. Is there a short explanation why we have to implement `RequestResponseCodec` for the `OneShotCodec`?

(Is this needed because under the hood the oneshot is still a request-response protocol?)

swap/src/protocol/alice/steps.rs
```diff
@@ -35,11 +34,11 @@ pub async fn negotiate(
     xmr_amount: monero::Amount,
     event_loop_handle: &mut EventLoopHandle,
     config: Config,
-) -> Result<(ResponseChannel<AliceToBob>, alice::State3)> {
+) -> Result<(PeerId, alice::State3)> {
     trace!("Starting negotiate");
 
     // todo: we can move this out, we dont need to timeout here
```
da-kami, line 40: 🙃 Is the `ToDo` still valid? Could we action it as part of this PR?

da-kami (approved):
Nice work! I added a few questions and suggestions. Not blockers for merging :)

swap/src/protocol/bob/message5.rs
```diff
@@ -28,27 +28,27 @@ pub enum OutEvent {
     Msg,
 }
 
-/// A `NetworkBehaviour` that represents sending message 3 to Alice.
+/// A `NetworkBehaviour` that represents sending message 5 to Alice.
```
thomaseizinger, line 31:
```suggestion
/// A `NetworkBehaviour` that represents sending `Message5` to Alice.
```

swap/src/protocol/bob/message5.rs
```diff
@@ -66,7 +66,7 @@ impl Default for Behaviour {
         Self {
             rr: RequestResponse::new(
                 Codec::default(),
-                vec![(Message3Protocol, ProtocolSupport::Full)],
+                vec![(Message5Protocol, ProtocolSupport::Full)],
```
thomaseizinger, line 70: Is `Full` actually correct here? Given that you are structuring it into Alice and Bob behaviours, this is likely either only inbound or outbound.

swap/src/network/request_response.rs
thomaseizinger: > (Is this needed because under the hood the oneshot is still a request-response protocol?)

Yes, it is using `libp2p-request-response`.
See the `rr` field within `Behaviour` in the next file in this diff.

thomaseizinger: You either have to implement all libp2p abstraction layers yourself, that is:

`ProtocolInfo`, `InboundUpgrade`, `OutboundUpgrade`, `ProtocolsHandler` and `NetworkBehaviour`.

Or you use a more high-level abstraction like `RequestResponse` that takes care of this.

Regardless of which one you are using, you will need to specify somehow, how a message gets encoded and decoded into bytes. Usually you'd do that in either `InboundUpgrade`/`OutboundUpgrade` or in the `ProtocolsHandler` but you don't get in touch with these abstractions when you use `RequestResponse` hence you will need to specify that in a different way: the `RequestResponseCodec` abstraction of `libp2p-request-response`.

I can recommending spending some time trying to implement your own protocol where you touch all the layers (or carefully read the existing code). It will make a lot more sense once you know what is going on behind the scenes.

swap/src/network/request_response.rs
```diff
@@ -171,3 +184,92 @@ where
...
+    type Response = Response;
+
+    async fn read_request<T>(&mut self, _: &Self::Protocol, io: &mut T) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        debug!("enter read_request");
+        let message = upgrade::read_one(io, BUF_SIZE)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
```
thomaseizinger, line 218: The `ReadOneError` returned here actually [has an `Io` variant](https://docs.rs/libp2p/0.34.0/libp2p/core/upgrade/enum.ReadOneError.html) that you could just return. Only the `TooLarge` variant would need to be mapped and that one probably should be an `io::ErrorKind::Other`.

swap/src/protocol/alice/message5.rs
```diff
@@ -73,12 +74,11 @@ impl NetworkBehaviourEventProcess<RequestResponseEvent<BobToAlice, AliceToBob>>
...
-                    debug!("Received message 5");
-                    self.events.push_back(OutEvent::Msg(msg));
-                    // Send back empty response so that the request/response protocol completes.
-                    self.rr.send_response(channel, AliceToBob::Message5);
-                }
+                let Request::Message5(msg) = request;
+                debug!("Received message 5");
+                self.events.push_back(OutEvent::Msg(msg));
+                // Send back empty response so that the request/response protocol completes.
+                let _ = self.rr.send_response(channel, Response::Message5);
```
thomaseizinger: I think you shouldn't need to send back an empty response, you can just close the substream after writing / reading to it (should probably happen in the codec).

swap/src/network/request_response.rs
```diff
@@ -90,6 +95,12 @@ impl ProtocolName for Message2Protocol {
     }
 }
 
+impl ProtocolName for Message4Protocol {
+    fn protocol_name(&self) -> &[u8] {
+        b"/xmr/btc/message4/1.0.0"
```
thomaseizinger, line 98: Eventually we might want to namespace these under `/comit` or something.

swap/src/protocol/alice/message4.rs
thomaseizinger: My 2 sats:

Numbering them scales better because not all messages have a single purpose. In these protocols, you often put whatever you can send at once into a single message which makes it hard to name.

I'd prefer consistency in naming over descriptive names that you can only use sometimes.

However, if you can come up with a descriptive name for all of them, go for it :)

swap/src/network/request_response.rs
D4nte: a910bc204636cf5c85f520168aab3c5061ba31d7

swap/src/network/request_response.rs
D4nte: We need a codec for `RequestResponse`.
Yes, the "one shot" is a request (the one shot) and a response (empty, act as ACK) behaviour.

swap/src/protocol/bob.rs
```diff
@@ -210,7 +212,8 @@ pub enum OutEvent {
     SwapResponse(alice::SwapResponse),
     Message0(Box<alice::Message0>),
     Message1(Box<alice::Message1>),
-    Message2(alice::Message2),
+    Message2,
```
thomaseizinger, line 215: What is the "Event" of `Message2`? Given that it doesn't have any content, I assume it is `Message2Sent`? These variants could use better names :)

swap/src/protocol/alice/steps.rs
D4nte: I tried to action it as part of #157. It will be taken care of when starting to build nectar.

swap/src/protocol/bob/message4.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Default for Behaviour {
+    fn default() -> Self {
+        let timeout = Duration::from_secs(TIMEOUT);
+        let mut config = RequestResponseConfig::default();
+        config.set_request_timeout(timeout);
+
+        Self {
+            rr: RequestResponse::new(
+                OneShotCodec::default(),
+                vec![(Message4Protocol, ProtocolSupport::Full)],
```
thomaseizinger, line 59: Again, `Full` is likely to be incorrect.

swap/src/protocol/bob.rs
```diff
@@ -302,8 +314,8 @@ impl Behaviour {
 
     /// Sends Bob's fourth message to Alice.
     pub fn send_message3(&mut self, alice: PeerId, tx_redeem_encsig: EncryptedSignature) {
```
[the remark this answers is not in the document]
D4nte: As explained during the call. I am starting to bring the terminology of the sequence diagram but I am not touching the "swap setup" phase.

You must look at the sequence diagram to understand the numbers in this transition phase.

swap/src/protocol/alice/message2.rs
```diff
@@ -79,7 +79,7 @@ impl NetworkBehaviourEventProcess<RequestResponseEvent<BobToAlice, AliceToBob>>
                 if let BobToAlice::Message2(msg) = request {
```
thomaseizinger, line 79: Could avoid a level of nesting by including this pattern in the outer match. Not sure if it worth it though.

swap/src/main.rs
```diff
@@ -25,16 +25,18 @@ use tracing::{info, log::LevelFilter};
...
-mod fs;
 pub mod monero;
 pub mod network;
 pub mod protocol;
 pub mod seed;
 pub mod trace;
 
+mod cli;
+mod fs;
+mod serde_peer_id;
```
thomaseizinger, line 38: We should consider upstreaming a `serde` feature to `libp2p`.

swap/src/protocol/bob.rs
D4nte: These will be scraped when I use `libp2p-async-await` :)

swap/src/main.rs
D4nte: yes :)

D4nte:
bors retry

swap/src/protocol/bob.rs
thomaseizinger: Fair enough, still some feedback to consider for the future: Naming any event should likely include a verb in simple past tense.

D4nte:
Bors retry

D4nte:
bors retry

PR #165, opened 2021-01-28
Add reset-config command that allows removing config file

da-kami:
Note: This PR causes side effects with the data-dir (because the data-dir is defined globally and the seed is always generated), this was fixed in follow up: https://github.com/comit-network/xmr-btc-swap/pull/166

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -47,6 +47,10 @@ pub enum Command {
     },
     History,
     Resume(Resume),
+    ResetConfig {
+        #[structopt(flatten)]
+        config: Config,
+    },
 }
 
 #[derive(structopt::StructOpt, Debug)]
--- a/swap/src/config.rs
+++ b/swap/src/config.rs
@@ -1,4 +1,4 @@
-use crate::fs::ensure_directory_exists;
+use crate::fs::{default_config_path, ensure_directory_exists};
 use anyhow::{Context, Result};
 use config::{Config, ConfigError};
 use dialoguer::{theme::ColorfulTheme, Input};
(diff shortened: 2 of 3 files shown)


D4nte:
> Note: This PR causes side effects with the data-dir (because the data-dir is defined globally and the seed is always generated), this was fixed in follow up: #166

What are you trying to do ? Why providing this feature? This seems unlikely to be used by anyone. On the contrary, I see it as a danger for a user that do it by mistake and has to reconfigure things.

da-kami:
> > Note: This PR causes side effects with the data-dir (because the data-dir is defined globally and the seed is always generated), this was fixed in follow up: #166
> 
> What are you trying to do ? Why providing this feature? This seems unlikely to be used by anyone. On the contrary, I see it as a danger for a user that do it by mistake and has to reconfigure things.

This is a useful feature. What if you enter something wrong when setting up your config? 
I don't understand the concern that a user would "trigger this by accident". I think it is more likely that a user enters something wrong and then would look for such a feature.
We provided similar functionality with Ambrosia.

I don't feel strong about it and can remove it if you think we should not provide it.

D4nte:
> > > Note: This PR causes side effects with the data-dir (because the data-dir is defined globally and the seed is always generated), this was fixed in follow up: #166
> > 
> > 
> > What are you trying to do ? Why providing this feature? This seems unlikely to be used by anyone. On the contrary, I see it as a danger for a user that do it by mistake and has to reconfigure things.
> 
> This is a useful feature. What if you enter something wrong when setting up your config?
> I don't understand the concern that a user would "trigger this by accident". I think it is more likely that a user enters something wrong and then would look for such a feature.
> We provided similar functionality with Ambrosia.
> 
> I don't feel strong about it and can remove it if you think we should not provide it.

When you reset the config, do you get the wizard to re-enter the config on the go or does it just delete the file?

If you enter something wrong you go correct it in the config file or you manually delete the file.

This is an added feature which is not part of the initial scope. As you mentioned, we are tight on time. For us to achieve our goals we need to ensure we don't get distracted.

da-kami:
> > > > Note: This PR causes side effects with the data-dir (because the data-dir is defined globally and the seed is always generated), this was fixed in follow up: #166
> > > 
> > > 
> > > What are you trying to do ? Why providing this feature? This seems unlikely to be used by anyone. On the contrary, I see it as a danger for a user that do it by mistake and has to reconfigure things.
> > 
> > 
> > This is a useful feature. What if you enter something wrong when setting up your config?
> > I don't understand the concern that a user would "trigger this by accident". I think it is more likely that a user enters something wrong and then would look for such a feature.
> > We provided similar functionality with Ambrosia.
> > I don't feel strong about it and can remove it if you think we should not provide it.
> 
> When you reset the config, do you get the wizard to re-enter the config on the go or does it just delete the file?
> 
> If you enter something wrong you go correct it in the config file or you manually delete the file.
> 
> This is an added feature which is not part of the initial scope. As you mentioned, we are tight on time. For us to achieve our goals we need to ensure we don't get distracted.

OK, will remove it then and try to not get distracted next time ;)

PR #170, opened 2021-02-01
Cancel and refund commands

da-kami:
I plugged the cancel and refund logic into the current state/state-machine logic of the swap.

## Follow ups  (out of scope)

We might want to record issues to be tackled later, since we are on a tight time budget :)
Please let me know what you think @D4nte @rishflab 

### Problems with `ack` after sending a message

Alice was waiting forever when awaiting the `ack` from bob when sending the lock proof in case she runs into a dial error. It seems the `acks` can cause the program to hang. This is a severe problem that we most probably will encountered in production at some point. For this PR I wrapped the `ack` of Alice upon sending the `encsig` in a timeout to work around this problem, see https://github.com/comit-network/xmr-btc-swap/pull/170/commits/7463081f88c24af6496c6743d50766db24b31112 - but **we might want to consider to remove all `ack` message. I don't see much value in them if we don't have a resilient retry strategy.**

### Do not require Monero wallet for cancel/refund

The cancel/refund commands don't require a monero wallet.
In this PR we re-uses the builder which requires the monero wallet as well - and we check for the monero balance upon wallet initialization, so the command will fail if no monero wallet is started.

### Save Alice connection info in Bob DB

Save Alice's peer-id/address in DB: It's cumbersome for the user to lookup those details again.

### DIFF
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -29,7 +29,7 @@ use protocol::{alice, bob, bob::Builder, SwapAmounts};
 use std::{path::PathBuf, sync::Arc};
 use structopt::StructOpt;
 use trace::init_tracing;
-use tracing::info;
+use tracing::{error, info, warn};
 use uuid::Uuid;
 
 pub mod bitcoin;
@@ -247,11 +247,12 @@ async fn main() -> Result<()> {
                 Ok((txid, _)) => {
                     info!("Cancel transaction successfully published with id {}", txid)
                 }
-                Err(CancelError::CancelTimelockNotExpiredYet) => {
-                    info!("The Cancel Transaction cannot be published yet, because the timelock has not expired. Please try again later.")
-                }
+                Err(CancelError::CancelTimelockNotExpiredYet) => error!(
+                    "The Cancel Transaction cannot be published yet, \
+                    because the timelock has not expired. Please try again later."
+                ),
                 Err(CancelError::CancelTxAlreadyPublished) => {
-                    info!("The Cancel Transaction has already been published.")
+                    warn!("The Cancel Transaction has already been published.")
                 }
             }
         }
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -13,11 +13,12 @@
 #![allow(non_snake_case)]
 
 use crate::{
-    cli::{Command, Options, Resume},
+    cli::{Cancel, Command, Options, Resume},
     config::{
         initial_setup, query_user_for_initial_testnet_config, read_config, ConfigNotInitialized,
     },
     execution_params::GetExecutionParams,
+    protocol::bob::cancel::CancelError,
 };
 use anyhow::{Context, Result};
 use database::Database;
@@ -208,6 +209,36 @@ async fn main() -> Result<()> {
 
             tokio::spawn(async move { event_loop.run().await });
             bob::run(swap).await?;
+        }
+        Command::Cancel(Cancel::BuyXmr {
+            swap_id,
+            alice_peer_id,
+            alice_addr,
+            config,
+        }) => {
+            // TODO: Optimization: Only init the Bitcoin wallet, Monero wallet unnecessary
+            let (bitcoin_wallet, monero_wallet) =
+                init_wallets(config.path, bitcoin_network, monero_network).await?;
+
+            let bob_factory = Builder::new(
+                seed,
+                db_path,
+                swap_id,
+                Arc::new(bitcoin_wallet),
+                Arc::new(monero_wallet),
+                alice_addr,
+                alice_peer_id,
+                execution_params,
+            );
+            let (swap, event_loop) = bob_factory.build().await?;
+
+            tokio::spawn(async move { event_loop.run().await });
+
+            match bob::cancel(swap.swap_id, swap.state, swap.bitcoin_wallet, swap.db).await? {
+                Ok((txid, _)) => { info!("Cancel transaction successfully published with id {}", txid)},
+                Err(CancelError::CancelTimelockNotExpiredYet) => {info!("The Cancel Transaction cannot be published yet, because the timelock has not expired. Please try again later.")},
+                Err(CancelError::CancelTxAlreadyPublished) => {info!("The Cancel Transaction has already been published.")}
+            }
         }
     };
 
--- a/swap/src/main.rs
+++ b/swap/src/main.rs
@@ -215,6 +215,7 @@ async fn main() -> Result<()> {
             alice_peer_id,
             alice_addr,
             config,
+            force,
         }) => {
             // TODO: Optimization: Only init the Bitcoin wallet, Monero wallet unnecessary
             let (bitcoin_wallet, monero_wallet) =
@@ -234,7 +235,15 @@ async fn main() -> Result<()> {
 
             tokio::spawn(async move { event_loop.run().await });
 
-            match bob::cancel(swap.swap_id, swap.state, swap.bitcoin_wallet, swap.db).await? {
+            match bob::cancel(
+                swap.swap_id,
+                swap.state,
+                swap.bitcoin_wallet,
+                swap.db,
+                force,
+            )
+            .await?
+            {
                 Ok((txid, _)) => {
                     info!("Cancel transaction successfully published with id {}", txid)
                 }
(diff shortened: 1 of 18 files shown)


da-kami:
Just noticed that `--force` override is missing, I'll add that tomorrow in a follow up PR so the review of this PR is not blocked.

D4nte:
> we might want to consider to remove all ack message. I don't see much value in them if we don't have a resilient retry strategy.

In the scenario where Alice sends a message to Bob, Bob replies with a ack.

If we remove the ack then Alice might send the message to Bob, save in DB that message is sent and shutdown whereas actually, Bob did not yet receive the message (because of network issue or because the event loop did not get to it).

Hence, we need to wait for Bob to ack a message before considering it as "sent" and saving in DB.

swap/src/main.rs
```diff
@@ -209,6 +210,71 @@ async fn main() -> Result<()> {
...
+                alice_peer_id,
+                execution_params,
+            );
+            let (swap, event_loop) = bob_factory.build().await?;
+
+            tokio::spawn(async move { event_loop.run().await });
+
+            match bob::cancel(swap.swap_id, swap.state, swap.bitcoin_wallet, swap.db, force).await? {
+                Ok((txid, _)) => { info!("Cancel transaction successfully published with id {}", txid)},
+                Err(CancelError::CancelTimelockNotExpiredYet) => {info!("The Cancel Transaction cannot be published yet, because the timelock has not expired. Please try again later.")},
```
D4nte, line 240:
```suggestion
                Err(CancelError::CancelTimelockNotExpiredYet) => {error!("The Cancel Transaction cannot be published yet, because the timelock has not expired. Please try again later.")},
```
Did you run the formatter?

swap/src/main.rs
```diff
@@ -209,6 +210,71 @@ async fn main() -> Result<()> {
...
+                Err(CancelError::CancelTxAlreadyPublished) => {info!("The Cancel Transaction has already been published.")}
```
D4nte, line 241:
```suggestion
                Err(CancelError::CancelTxAlreadyPublished) => {warn!("The Cancel Transaction has already been published.")}
```

swap/src/protocol/alice/event_loop.rs
```diff
@@ -123,13 +127,27 @@ impl EventLoopHandle {
...
+        msg: TransferProof,
+        execution_params: ExecutionParams,
+    ) -> Result<()> {
         let _ = self.send_transfer_proof.send((bob, msg)).await?;
 
-        self.recv_transfer_proof_ack
-            .recv()
-            .await
-            .ok_or_else(|| anyhow!("Failed to receive transfer proof ack from Bob"))?;
+        // TODO: Re-evaluate if these acknowledges are necessary at all.
```
D4nte, line 138: I commented on the review re-ack. I think they are necessary to avoid scenarios where one party restart and wrongly assumes that the previous message they sent did send.

Are you saying that if Alice did not dial Bob then the `send_transfer_proof.send` does **not** fail?

swap/tests/bob_refunds_using_cancel_and_refund_command_timelock_not_exired.rs
```diff
@@ -0,0 +1,58 @@
...
+        let alice_handle = alice::run(alice_swap);
+        tokio::spawn(alice_handle);
+
+        let bob_state = bob::run_until(bob_swap, is_btc_locked).await.unwrap();
+        assert!(matches!(bob_state, BobState::BtcLocked {..}));
+
+        let (bob_swap, bob_join_handle) = ctx.stop_and_resume_bob_from_db(bob_join_handle).await;
+        assert!(matches!(bob_swap.state, BobState::BtcLocked {..}));
+
+        // Bob manually cancels
```
D4nte, line 22:
```suggestion
        // Bob tries but fails to manually cancel
```

swap/tests/bob_refunds_using_cancel_and_refund_command_timelock_not_exired.rs
```diff
@@ -0,0 +1,58 @@
...
+        .unwrap()
+        .err()
+        .unwrap();
+
+        assert!(matches!(result, CancelError::CancelTimelockNotExpiredYet));
+
+        let (bob_swap, bob_join_handle) = ctx.stop_and_resume_bob_from_db(bob_join_handle).await;
+        assert!(matches!(bob_swap.state, BobState::BtcLocked {..}));
+
+        // Bob manually refunds
```
D4nte, line 40:
```suggestion
        // Bob tries but fails to manually refund
```

D4nte (approved):

swap/src/protocol/alice/event_loop.rs
da-kami: That is exactly what I am saying. Without the timeout the program hangs forever on awaiting the ack.

swap/src/main.rs
da-kami: Yes, I did run the formatter (otherwise CI `static_analysis` check would fail). - it did not pick the long lines up. Only after adding a linebreak manually in one of the statements and running the formatter again it picked it up and formatted the whole block. Strange :]

da-kami:
> > we might want to consider to remove all ack message. I don't see much value in them if we don't have a resilient retry strategy.
> 
> In the scenario where Alice sends a message to Bob, Bob replies with a ack.
> 
> If we remove the ack then Alice might send the message to Bob, save in DB that message is sent and shutdown whereas actually, Bob did not yet receive the message (because of network issue or because the event loop did not get to it).
> 
> Hence, we need to wait for Bob to ack a message before considering it as "sent" and saving in DB.

I am still not convinced that the `ack` brings much value. I think it is totally fair to not deal with such kind of network issues - i.e. if Alice sends but Bob does not receive (and hence does not act) she will just refund. I would see how often these network problems occurs and then decide if more an elaborate retry strategy is worth the effort.
At the moment the "retry" strategy is failing the swap and restarting the program. To my mind that is not resilient at all and not a good strategy so I would remove the `acks`, create an issue and re-visit this. 

Besides that it is strange that sending the message does not fail. Theoretically it could happen, that the message is sent with Bob still being available, but Bob never is able to send the ack because at that point he is already shut down (theoretically it could be that split of a second that makes the difference...). But it is hard to say if that is the actual problem. 

I suppose for now the timeout is a solution we can agree on to make it more resilient. I won't insist on removing the acks, we can re-visit this problem later. One thing we should do again is trying to swap together on test-/stagenet to see if the previous network issues are actually resolved :)

D4nte:
@da-kami more investigation is needed. If sending a message fails than the Behaviour emit an event. We process these events in the implementation or NetworkBehaviourEvent... I can check after the retro

da-kami:
> @da-kami more investigation is needed. If sending a message fails than the Behaviour emit an event. We process these events in the implementation or NetworkBehaviourEvent... I can check after the retro

Should I hold off merging, because more investigation is needed @D4nte ? I am not sure investigation would lead to much at this point.

I think it might be coincidental to the test that sending does not fail but we don't receive the ack. 
In the real world this is potentially not likely to happen - but possible. For me the program should just be resilient enough to deal with such a scenario. (That's why I think the timeout is fine.)

Note: The `bors` build failure is due to a weird error when building (checksum of a library was not accessible) - this is not related to the changes made.

D4nte:
PRoceed with mergining. non blocking comment.

da-kami:
@thomaseizinger I pushed #53fcd92 as discussed, to give Alice's `refund` priority over `punish` - if she sees the `refund` tx before punish was finalized she should extract Bob's Monero key and refund no matter what. 

Here is the original CI build that shows the problem of Bob refunding, but Alice transitioning to a punish scenario: https://github.com/comit-network/xmr-btc-swap/runs/1858987512

I am still baffled by the fact that Alice publication of the punish transaction did not fail. 
Note: I was not able to reproduce this problem locally.

swap/tests/bob_refunds_using_cancel_and_refund_command.rs
```diff
@@ -0,0 +1,66 @@
...
+            bob_swap.bitcoin_wallet,
+            bob_swap.db,
+            false,
+        )
+        .await
+        .unwrap()
+        .unwrap();
+        assert!(matches!(state, BobState::BtcCancelled { .. }));
+
+        let (bob_swap, _) = ctx.stop_and_resume_bob_from_db(bob_join_handle).await;
```
D4nte, line 44: @da-kami So the user needs to run the `cancel` command, and then run `resume` and then run the `refund` command? Seems like poor UX.

PR #172, opened 2021-02-03
Remove dead code

bonomat:
We found some dead code.

Should this be used somewhere?


luckysori (approved):
This is very old :older_adult:

PR #173, opened 2021-02-04
Use libp2p-async-await to improve API of execution setup phase

D4nte:


swap/src/protocol/alice/state.rs
```diff
@@ -122,17 +129,14 @@ impl State0 {
         }
     }
 
-    pub fn next_message<R: RngCore + CryptoRng>(&self, rng: &mut R) -> alice::Message0 {
-        info!("Producing first message");
-        let dleq_proof_s_a = cross_curve_dleq::Proof::new(rng, &self.s_a);
-
+    pub fn next_message(&self) -> alice::Message0 {
```
thomaseizinger: In this case, I think this is a good refactoring because you can just pre-compute the dleq proof.

Generally though, ["call site dependency injection"](https://matklad.github.io/2020/12/28/csdi.html) plays very nicely with Rust's lifetimes.

I know we've done the opposite in the past in the comit-rs repository (with `Dependencies` etc) but the blogpost makes a few good arguments why you probably don't want to carry around those long-lived dependencies in a struct and rather limit your struct to data.

monero-harness/rustfmt.toml
```diff
@@ -1,7 +1,7 @@
 edition = "2018"
 condense_wildcard_suffixes = true
 format_macro_matchers = true
-merge_imports = true
+imports_granularity = "Crate"
```
thomaseizinger, line 4: You may enjoy the new `module` setting, it is much better than `crate`.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-08-13
```
thomaseizinger, line 1: Why is this project using nightly? Can we not compile on stable?

swap/src/protocol/alice/execution_setup.rs
```diff
@@ -0,0 +1,91 @@
...
+            inner: libp2p_async_await::Behaviour::new(b"/execution_setup/1.0.0"),
+        }
+    }
+}
+
+impl Behaviour {
+    fn run(&mut self, bob: PeerId, state0: State0) {
+        self.inner
+            .do_protocol_listener(bob, move |mut substream| async move {
+                let alice_message0 = state0.next_message();
```
thomaseizinger: If this message is not used until you transition to `State1`, why is it produced by `State0` and not `State1`?

swap/src/protocol/alice/execution_setup.rs
```diff
@@ -0,0 +1,91 @@
...
+
+                {
+                    let alice_message2 = state2.next_message();
+                    substream
+                        .write_message(
+                            &serde_cbor::to_vec(&alice_message2)
+                                .context("failed to serialize Message2")?,
+                        )
+                        .await?;
+                }
```
thomaseizinger: This block seems to be inconsistent and not necessary.

You didn't create one for writing message0 and also you are not taking advantage of the limited scope that `alice_message2` is in. Unless you want to enforce a specific `Drop` behaviour or need to avoid shadowing a variable, I'd recommend to not put extra scopes as it can be confusing why it is there precisely because sometimes you need an extra scope and then you start searching for the reason :)

swap/src/protocol/bob/swap.rs
```diff
@@ -377,29 +377,35 @@ async fn run_until_internal(
 pub async fn negotiate(
     state0: crate::protocol::bob::state::State0,
     amounts: SwapAmounts,
-    swarm: &mut EventLoopHandle,
+    event_loop_handle: &mut EventLoopHandle,
```
thomaseizinger, line 374: Minor nit: I would order the commits as such that the feature is in the end.

swap/src/protocol/alice/execution_setup.rs
```diff
@@ -39,7 +39,7 @@ impl Default for Behaviour {
 }
 
 impl Behaviour {
-    fn run(&mut self, bob: PeerId, state0: State0) {
+    pub fn run(&mut self, bob: PeerId, state0: State0) {
```
thomaseizinger, line 63: I had a comment in the previous commit about how you are getting away with this being private :grin: 

It is easier to review if code is used right away and you are not committing dead code separately (does the compiler not emit a dead-code warning for the other commit?). I would probably squash this one together with the one where you introduce it, even if it makes it slightly bigger.

swap/src/protocol/bob/swap_request.rs
```diff
@@ -35,14 +35,14 @@ pub struct OutEvent {
 #[behaviour(out_event = "OutEvent", poll_method = "poll")]
```
thomaseizinger, line 35: Why is this not using `event_process = false`?

swap/src/network/request_response.rs
```diff
@@ -19,54 +17,23 @@ pub const TIMEOUT: u64 = 3600; // One hour.
...
 }
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-/// Response are only used for acknowledgement purposes.
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum Response {
+    SwapResponse(Box<SwapResponse>),
     TransferProof,
     EncryptedSignature,
 }
```
thomaseizinger, line 32: This model looks like it can be improved. I doubt that a `TransferProof` request can be answered with an `EncryptedSignature` response. Why are these bundled together into one enum?

swap/src/protocol/alice/swap_response.rs
```diff
@@ -37,15 +37,15 @@ pub struct SwapResponse {
 #[behaviour(out_event = "OutEvent", poll_method = "poll")]
```
thomaseizinger, line 37: Same here, this can likely use `event_process = false`.

thomaseizinger (commented):
Definitely a feels-good PR with all the deleted code :)

swap/src/protocol/alice/state.rs
D4nte: Ah yeah, a revamp of all states is needed at some point.

I just want to have to carry the `rng` around. However, at the moment it feels silly to pass the `rng` when we just use `OsRng` anyway. Some thoughts are needed on this aspect.

monero-harness/rustfmt.toml
D4nte: As far as I could see I am happy with `"Crate"`

swap/src/protocol/alice/execution_setup.rs
D4nte: Because it can be produced by `state0`.

swap/src/protocol/bob/swap.rs
D4nte: > Minor nit: I would order the commits as such that the feature is in the end.

I tried but it involved merging work that I was not willing to do.

swap/src/protocol/alice/execution_setup.rs
D4nte: Both commits are huge in this case. I thought it would be better for the reviewers to have them separate.

swap/src/network/request_response.rs
D4nte: > Why are these bundled together into one enum?

Inspired for existing design, I haven't yet dug enough in the code to understand if there was any reason for this design and whether it can be split.

swap/src/protocol/bob/swap_request.rs
D4nte: Because we are not all as verse as you with libp2p :)

rust-toolchain
D4nte: Because of the secp25kfun crate:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> /home/froyer/.cargo/git/checkouts/secp256kfun-fb74af16957b2754/cdfbc76/secp256kfun/src/lib.rs:1:1
  |
1 | #![feature(rustc_attrs, min_specialization, external_doc)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

monero-harness/rustfmt.toml
thomaseizinger: Module has the advantage of causing a lot less merge conflicts because it breaks it up into multiple lines, thereby causing less whitespace change on adding/removing imports.

rust-toolchain
thomaseizinger: Oh didn't know we were using secpfun.

swap/src/protocol/alice/execution_setup.rs
thomaseizinger: Well, just because it can doesn't mean it should.
Like we see in this code, it requires us to produce the message unnecessarily early.

swap/src/protocol/alice/execution_setup.rs
thomaseizinger: I'd take a bigger but coherent diff over one that introduces code that is not used. It is harder to give feedback if I have to navigate back and forth between separate commits to understand the bigger picture.

swap/src/protocol/alice/execution_setup.rs
D4nte: I agree with you.

swap/src/protocol/bob/swap_request.rs
thomaseizinger: Time to change that! :D

swap/src/protocol/alice/state.rs
D4nte: This is very interesting. I think it could be a good team exercise to improve the current codebase as we very often used this pattern `State0 > send message0 > recv message1 > State1` in our PoC. What do you think?

monero-harness/rustfmt.toml
D4nte: Hum, we could consider it in a follow-up PR.

swap/src/protocol/alice/encrypted_signature.rs
```diff
@@ -59,48 +48,42 @@ impl Default for Behaviour {
...
             } => {
-                if let Request::EncryptedSignature(msg) = request {
-                    debug!("Received encrypted signature");
-                    self.events.push_back(OutEvent::Msg(*msg));
-                    // Send back empty response so that the request/response protocol completes.
-                    if let Err(error) = self.rr.send_response(channel, Response::EncryptedSignature)
-                    {
-                        error!("Failed to send Encrypted Signature ack: {:?}", error);
-                    }
+                debug!("Received encrypted signature from {}", peer);
```
da-kami, line 70: ❓ We don't have to use `push_back` to bubble the event up anymore? Is this because of using `event_process = false` ?

da-kami: I confused myself - the event is returned now.

thomaseizinger (commented):
Definitely a feels-good PR! :blush: 

A minor nit: Some of your commit messages start with a lowercase letter.

Some food for thought: I am not convinced that the separation into Alice and Bob Behaviours makes sense on _all_ levels. The mid-level behaviours for transfer proof look like a lot of duplication that could just live in one implementation and just needs to be called accordingly. I get that the idea is to make it impossible for Bob to send a specific message but if it amounts in this much code, this guarantee may just not be worth it.

D4nte:
> A minor nit: Some of your commit messages start with a lowercase letter.

1 commit message 🙃

swap/src/protocol/alice/state.rs
thomaseizinger: I never saw it the way that `State0` needs to produce `Message0` just because they are both `0`. The whole point of these states is that it gives you a type-safe interface of what you can do in a certain state. If a state can produce a message, you should produce it and send it to the other party.

If the message counter is global, then the first message will be produced by _some_ party's `State0` but all the other parties will only have a `receive` method in their `State0`. `a2l` is a good example here because it actually has three parties. That leads to situations where the first thing a certain party does is receive the 4th message of a specific protocol and produce the 1st message of the next one: https://github.com/comit-network/a2l-poc/blob/a261027f4bdb2384f715efa3f634d5796fe95019/src/receiver.rs#L123-L155

> What do you think?

I am of the strong opinion that if this is not how the protocol is currently implemented, then we should change it because it otherwise defeats the point of type-level programming.

PR #177, opened 2021-02-08
Nectar

D4nte:
Re-work Alice do be the Service Provider (nectar).
This means that Alice's event loop runs all the time and processes swap requests.

- We currently use a hardcoded fixed rate
- The binary for nectar is to be done in follow-up PR
- We removed the message acknowledgement feature as it created issues with other work and some work would have been needed to link a ack to the correct swap on Alice side.
- Because we removed the acks, we had to remove a number of restart tests. This can be re-introduced at a later stage once a decision on how to best test the restarts can be done
- We remove the punish test as it is a nectar (alice) feature and current focus is on the CLI (bob).


swap/src/protocol/alice/event_loop.rs
```diff
@@ -118,6 +146,14 @@ impl EventLoop {
         Ok((driver, handle))
     }
 
+    pub fn clone_handle(&self) -> EventLoopHandle {
+        EventLoopHandle {
+            recv_encrypted_signature: self.recv_encrypted_signature.subscribe(),
+            send_transfer_proof: self.handle.send_transfer_proof.clone(),
+            recv_transfer_proof_ack: self.recv_transfer_proof_ack.subscribe(),
+        }
+    }
```
thomaseizinger: Why not just implement `Clone`? You don't need to derive it, you can just implement it yourself.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -174,11 +203,76 @@ impl EventLoop {
...
+        // 2. Send response
+
+        let btc_amount = swap_request.btc_amount;
+        let xmr_amount = btc_amount.as_btc() * RATE as f64;
+        let xmr_amount = monero::Amount::from_monero(xmr_amount)?;
+        let swap_response = SwapResponse { xmr_amount };
+
+        self.swarm
+            .send_swap_response(channel, swap_response)
+            .context("Failed to send swap response")?;
```
thomaseizinger, line 213: We were talking about names of this at some point. Really what we are doing here is requesting a quote :)

swap/tests/happy_path_restart_bob_before_comm.rs
```diff
@@ -1,17 +1,13 @@
 pub mod testutils;
 
-use swap::protocol::{alice, bob, bob::BobState};
+use swap::protocol::{bob, bob::BobState};
 use testutils::{bob_run_until::is_xmr_locked, SlowCancelConfig};
 
 #[tokio::test]
 async fn given_bob_restarts_after_xmr_is_locked_resume_swap() {
     testutils::setup_test(SlowCancelConfig, |mut ctx| async move {
-        let (alice_swap, _) = ctx.new_swap_as_alice().await;
```
thomaseizinger, line 9: It appears that this commit should be part of the one where you adapt the test framework.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -103,59 +98,52 @@ pub struct EventLoop {
...
+            peer_id,
+            execution_params,
+            bitcoin_wallet,
+            monero_wallet,
+            db,
+            listen_address,
+            swap_amounts: Default::default(),
+            recv_encrypted_signature: recv_encrypted_signature.sender,
+            send_transfer_proof: send_transfer_proof.receiver,
+            handle,
```
thomaseizinger, line 143: It feels weird that the `EventLoop` has a handle to itself.

bors.toml
```diff
@@ -4,4 +4,7 @@ status = [
     "build_test (x86_64-apple-darwin)",
     "docker_tests (happy_path)",
     "docker_tests (happy_path_restart_bob_before_comm)",
+    "docker_tests (bob_refunds_using_cancel_and_refund_command.rs)",
```
thomaseizinger: Don't think you want the `.rs` extension here.

swap/tests/testutils/mod.rs
```diff
@@ -62,7 +62,7 @@ impl BobParams {
     }
 }
 
-pub struct BobEventLoopJoinHandle(JoinHandle<()>);
+pub struct BobEventLoopJoinHandle(pub JoinHandle<()>);
```
thomaseizinger, line 65: Instead of making this public, why no forward the `abort` call? newtypes are pretty useless if their inner fields are public.

thomaseizinger (approved):
Thanks for structuring the commits, was good to review!

Left some comments along the way. The only thing that is a bit bothering is the removal of the `punish` test but I also don't see an easy way of adding that back for now.

thomaseizinger:
A comment regarding GitHub: If you stop marking checks as "required" in Github but only make "bors" mandatory to merge, you can change the merge requirements atomically by editing `bors.toml` and you don't have to touch GitHub config.

swap/Cargo.toml
```diff
@@ -5,6 +5,12 @@ authors = ["CoBloX developers <team@coblox.tech>"]
 edition = "2018"
 description = "XMR/BTC trustless atomic swaps."
 
+[[bin]]
+name = "cli"
```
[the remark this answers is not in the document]
D4nte: huum, good to know!

swap/src/protocol/alice/behaviour.rs
```diff
@@ -53,8 +53,10 @@ impl From<alice::OutEvent> for OutEvent {
 
 impl From<execution_setup::OutEvent> for OutEvent {
     fn from(event: execution_setup::OutEvent) -> Self {
+        use crate::protocol::alice::execution_setup::OutEvent::*;
         match event {
-            execution_setup::OutEvent::Done(res) => OutEvent::ExecutionSetupDone(res.map(Box::new)),
+            Done(state3) => OutEvent::ExecutionSetupDone(Box::new(state3)),
+            Failure(err) => OutEvent::Failure(err),
```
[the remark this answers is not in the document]
D4nte: Thank for pointing this out.

Originally, this change was needed because `anyhow::Error` is not `Clone` and hence cannot be in a `broadcast::Channel`. However, I removed the channel that carries `anyhow::Result<State3>` so I could have removed this refactoring.

I kept the refactoring because I just wanted to align the result processing with the other behaviours by gathering all failures in the `OutEvent` of the "root" `Behaviour`. IMHO, having this odd case of the event returning a `Result` without a proper justification for it would be confusing for future readers.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -79,54 +85,77 @@ impl EventLoopHandle {
...
     swarm: libp2p::Swarm<Behaviour>,
+    peer_id: PeerId,
+    execution_params: ExecutionParams,
+    bitcoin_wallet: Arc<bitcoin::Wallet>,
+    monero_wallet: Arc<monero::Wallet>,
+    db: Arc<Database>,
+    listen_address: Multiaddr,
+
+    // Amounts agreed upon for swaps currently in the setup execution phase
+    // Note: We can do a setup execution with only one peer at a time.
```
da-kami, line 91: ❓ Why is the setup execution protocol currently restricted to one peer at the time?
As far as I could see the setup is triggered upon swap request. I multiple peers trigger it wouldn't it work? 
I would like to understand better what is missing to allow concurrent swaps (setup).
The hashmap indicates that only one swap per peer is allowed concurrently.

da-kami (approved):

swap/src/protocol/alice/event_loop.rs
D4nte: Hum, good point, the thing is that to call the `receiver` end of a broadcast channel you need the `sender` end.
I could keep both ends in an `EventLoopHandle` to allow the cloning.

But in terms of API, it does not really matter because it is used in the `EventLoop` to produce a new handle. So maybe I should change the name of the function instead to `new_handle`.

swap/Cargo.toml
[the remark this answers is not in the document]
D4nte: As discussed, I don't think it's possible as I previously tried your strategy with @da-kami during pair programming. I'll leave it for now.

swap/src/protocol/alice.rs
```diff
@@ -175,30 +175,20 @@ impl Builder {
...
         let amounts = SwapAmounts {
             btc: btc_to_swap,
             xmr: xmr_to_swap,
         };
 
-        let a = bitcoin::SecretKey::new_random(rng);
-        let s_a = cross_curve_dleq::Scalar::random(rng);
-        let v_a = monero::PrivateViewKey::new_random(rng);
-        let redeem_address = self.bitcoin_wallet.new_address().await?;
-        let punish_address = redeem_address.clone();
```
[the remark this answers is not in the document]
D4nte: The reason why (Just added that to the commit, will push soon) is because the `EventLoop` is now handling the instantiation of `State0` using `State0::new` so I just wanted to avoid having to instantiate too many variable to use this function within the event loop. In short, I just want the API to be easier to use so I can then use it :)

swap/src/protocol/alice.rs
```diff
@@ -56,7 +56,7 @@ pub struct Builder {
     swap_id: Uuid,
     identity: Keypair,
     peer_id: PeerId,
-    db_path: PathBuf,
+    db: Database,
```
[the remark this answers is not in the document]
D4nte: Updated https://github.com/comit-network/xmr-btc-swap/commit/212c0a88850dabaf492daed377f7362ac310a201

swap/src/protocol/alice/state.rs
```diff
@@ -128,8 +128,8 @@ impl State0 {
             punish_address,
             btc,
             xmr,
-            cancel_timelock,
-            punish_timelock,
+            cancel_timelock: execution_params.bitcoin_cancel_timelock,
```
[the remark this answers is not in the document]
D4nte: Next time :)

swap/src/protocol/alice/event_loop.rs
D4nte: True, let's do that in follow-up

swap/src/protocol/alice/event_loop.rs
D4nte: Actually we don't need the full handle but we do have to keep the `sender` end of the `mpsc::Channel` channels to be able to clone them when producing a new handle.

I'll remove the handle but it means the event loop will hold both ends of some channels.

bors.toml
D4nte: Yes, typo already corrected.

swap/src/protocol/alice/event_loop.rs
D4nte: One execution setup per peer at a given time. will correct comment to make it clearer.

swap/src/protocol/alice/event_loop.rs
thomaseizinger: Ah, sorry. I didn't realize that this function is on `EventLoop` and not `EventLoopHandle`. Makes total sense!

D4nte:
All comments have been addressed.

bors r+

PR #178, opened 2021-02-08
Replace bitcoind wallet with bdk wallet

da-kami:
Create and pair bitcoind and electrs testcontainers using the same docker network and volume.

Follow-up to be fixed on top: 

- [ ] Remove bitcoin-harness dependency and directly depend on https://github.com/thomaseizinger/rust-jsonrpc-client for the RPC client.

### DIFF
--- a/swap/src/bitcoin/timelocks.rs
+++ b/swap/src/bitcoin/timelocks.rs
@@ -18,11 +18,16 @@ impl BlockHeight {
     pub const fn new(block_height: u32) -> Self {
         Self(block_height)
     }
+    pub const fn checked_sub(self, rhs: Self) -> Option<Self> {
+        match self.0.checked_sub(rhs.0) {
+            Some(result) => Some(BlockHeight(result)),
+            None => None,
+        }
+    }
 }
 
 impl Add<u32> for BlockHeight {
     type Output = BlockHeight;
-
     fn add(self, rhs: u32) -> Self::Output {
         BlockHeight(self.0 + rhs)
     }
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -23,6 +23,14 @@ use tokio::{sync::Mutex, time::interval};
 
 const SLED_TREE_NAME: &str = "default_tree";
 
+#[derive(Debug)]
+enum Error {
+    Io(reqwest::Error),
+    Parse(std::num::ParseIntError),
+    NotYetMined,
+    JsonDeserialisation(reqwest::Error),
+}
+
 pub struct Wallet {
     pub inner: Arc<Mutex<bdk::Wallet<ElectrumBlockchain, bdk::sled::Tree>>>,
     pub network: bitcoin::Network,
@@ -150,7 +158,7 @@ impl BroadcastSignedTransaction for Wallet {
 
 #[async_trait]
 impl WatchForRawTransaction for Wallet {
-    async fn watch_for_raw_transaction(&self, txid: Txid) -> Transaction {
+    async fn watch_for_raw_transaction(&self, txid: Txid) -> Result<Transaction> {
         tracing::debug!("watching for tx: {}", txid);
         retry(ConstantBackoff::new(Duration::from_secs(1)), || async {
             let client = Client::new(self.rpc_url.as_ref())?;
@@ -159,7 +167,7 @@ impl WatchForRawTransaction for Wallet {
             Ok(tx)
         })
         .await
-        .expect("transient errors to be retried")
+        .map_err(|err| anyhow!("transient errors to be retried: {:?}", err))
     }
 }
 
@@ -174,19 +182,11 @@ impl GetRawTransaction for Wallet {
 
 #[async_trait]
 impl GetBlockHeight for Wallet {
-    async fn get_block_height(&self) -> BlockHeight {
-        // todo: create this url using the join() api in the Url type
-        let url = format!("{}{}", self.http_url.as_str(), "blocks/tip/height");
-        #[derive(Debug)]
-        enum Error {
-            Io(reqwest::Error),
-            Parse(std::num::ParseIntError),
-        }
+    async fn get_block_height(&self) -> Result<BlockHeight> {
+        let url = self.http_url.join("blocks/tip/height")?;
         let height = retry(ConstantBackoff::new(Duration::from_secs(1)), || async {
-            // todo: We may want to return early if we cannot connect to the electrum node
-            // rather than retrying
             let height = reqwest::Client::new()
-                .request(Method::GET, &url)
+                .request(Method::GET, url.clone())
                 .send()
                 .await
                 .map_err(Error::Io)?
@@ -194,37 +194,29 @@ impl GetBlockHeight for Wallet {
                 .await
                 .map_err(Error::Io)?
                 .parse::<u32>()
-                .map_err(Error::Parse)?;
+                .map_err(|err| backoff::Error::Permanent(Error::Parse(err)))?;
             Result::<_, backoff::Error<Error>>::Ok(height)
         })
         .await
-        .expect("transient errors to be retried");
+        .map_err(|err| anyhow!("transient errors to be retried: {:?}", err))?;
 
-        BlockHeight::new(height)
+        Ok(BlockHeight::new(height))
     }
 }
 
 #[async_trait]
 impl TransactionBlockHeight for Wallet {
-    async fn transaction_block_height(&self, txid: Txid) -> BlockHeight {
-        // todo: create this url using the join() api in the Url type
-        let url = format!("{}tx/{}/status", self.http_url, txid);
+    async fn transaction_block_height(&self, txid: Txid) -> Result<BlockHeight> {
+        let url = self.http_url.join(&format!("tx/{}/status", txid))?;
+
         #[derive(Serialize, Deserialize, Debug, Clone)]
         struct TransactionStatus {
             block_height: Option<u32>,
             confirmed: bool,
         }
-        // todo: See if we can make this error handling more elegant
-        // errors
-        #[derive(Debug)]
-        enum Error {
-            Io(reqwest::Error),
-            NotYetMined,
-            JsonDeserialisation(reqwest::Error),
-        }
         let height = retry(ConstantBackoff::new(Duration::from_secs(1)), || async {
             let resp = reqwest::Client::new()
-                .request(Method::GET, &url)
+                .request(Method::GET, url.clone())
                 .send()
                 .await
                 .map_err(|err| backoff::Error::Transient(Error::Io(err)))?;
@@ -241,9 +233,9 @@ impl TransactionBlockHeight for Wallet {
             Result::<_, backoff::Error<Error>>::Ok(block_height)
         })
         .await
-        .expect("transient errors to be retried");
+        .map_err(|err| anyhow!("transient errors to be retried: {:?}", err))?;
 
-        BlockHeight::new(height)
+        Ok(BlockHeight::new(height))
     }
 }
 
@@ -260,9 +252,9 @@ impl WaitForTransactionFinality for Wallet {
         let mut interval = interval(execution_params.bitcoin_avg_block_time / 4);
 
         loop {
-            let tx_block_height = self.transaction_block_height(txid).await;
+            let tx_block_height = self.transaction_block_height(txid).await?;
             tracing::debug!("tx_block_height: {:?}", tx_block_height);
-            let block_height = self.get_block_height().await;
+            let block_height = self.get_block_height().await?;
             tracing::debug!("latest_block_height: {:?}", block_height);
             if let Some(confirmations) = block_height.checked_sub(tx_block_height) {
                 tracing::debug!("confirmations: {:?}", confirmations);
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use backoff::{backoff::Constant as ConstantBackoff, tokio::retry};
 use bdk::{
     blockchain::{noop_progress, Blockchain, ElectrumBlockchain},
-    electrum_client::{Client, ElectrumApi},
+    electrum_client::{self, Client, ElectrumApi},
     keys::GeneratableDefaultOptions,
     FeeRate,
 };
@@ -45,7 +45,10 @@ impl Wallet {
         network: bitcoin::Network,
         waller_dir: &Path,
     ) -> Result<Self> {
-        let client = Client::new(electrum_rpc_url.as_str())
+        // Workaround for https://github.com/bitcoindevkit/rust-electrum-client/issues/47.
+        let config = electrum_client::ConfigBuilder::default().retry(2).build();
+
+        let client = Client::from_config(electrum_rpc_url.as_str(), config)
             .map_err(|e| anyhow!("Failed to init electrum rpc client: {:?}", e))?;
 
         let db = bdk::sled::open(waller_dir)?.open_tree(SLED_TREE_NAME)?;
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -186,7 +186,7 @@ impl GetRawTransaction for Wallet {
 #[async_trait]
 impl GetBlockHeight for Wallet {
     async fn get_block_height(&self) -> Result<BlockHeight> {
-        let url = self.http_url.join("blocks/tip/height")?;
+        let url = blocks_tip_height_url(&self.http_url)?;
         let height = retry(ConstantBackoff::new(Duration::from_secs(1)), || async {
             let height = reqwest::Client::new()
                 .request(Method::GET, url.clone())
@@ -210,8 +210,7 @@ impl GetBlockHeight for Wallet {
 #[async_trait]
 impl TransactionBlockHeight for Wallet {
     async fn transaction_block_height(&self, txid: Txid) -> Result<BlockHeight> {
-        let url = self.http_url.join(&format!("tx/{}/status", txid))?;
-
+        let url = tx_status_url(txid, &self.http_url)?;
         #[derive(Serialize, Deserialize, Debug, Clone)]
         struct TransactionStatus {
             block_height: Option<u32>,
@@ -276,5 +275,44 @@ impl WaitForTransactionFinality for Wallet {
 impl GetNetwork for Wallet {
     async fn get_network(&self) -> bitcoin::Network {
         self.inner.lock().await.network()
+    }
+}
+
+fn tx_status_url(txid: Txid, base_url: &Url) -> Result<Url> {
+    let url = base_url.join(&format!("tx/{}/status", txid))?;
+    Ok(url)
+}
+
+fn blocks_tip_height_url(base_url: &Url) -> Result<Url> {
+    let url = base_url.join("blocks/tip/height")?;
+    Ok(url)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        bitcoin::{
+            wallet::{blocks_tip_height_url, tx_status_url},
+            Txid,
+        },
+        cli::config::DEFAULT_ELECTRUM_HTTP_URL,
+    };
+    use reqwest::Url;
+
+    #[test]
+    fn create_tx_status_url_from_default_base_url_success() {
+        let txid: Txid = Txid::default();
+        let base_url = Url::parse(DEFAULT_ELECTRUM_HTTP_URL).expect("Could not parse url");
+        let url = tx_status_url(txid, &base_url).expect("Could not create url");
+        let expected = format!("https://blockstream.info/testnet/api/tx/{}/status", txid);
+        assert_eq!(url.as_str(), expected);
+    }
+
+    #[test]
+    fn create_block_tip_height_url_from_default_base_url_success() {
+        let base_url = Url::parse(DEFAULT_ELECTRUM_HTTP_URL).expect("Could not parse url");
+        let url = blocks_tip_height_url(&base_url).expect("Could not create url");
+        let expected = "https://blockstream.info/testnet/api/blocks/tip/height";
+        assert_eq!(url.as_str(), expected);
     }
 }
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -13,7 +13,7 @@ use backoff::{backoff::Constant as ConstantBackoff, tokio::retry};
 use bdk::{
     blockchain::{noop_progress, Blockchain, ElectrumBlockchain},
     electrum_client::{self, Client, ElectrumApi},
-    keys::GeneratableDefaultOptions,
+    miniscript::bitcoin::PrivateKey,
     FeeRate,
 };
 use reqwest::{Method, Url};
@@ -43,7 +43,8 @@ impl Wallet {
         electrum_rpc_url: Url,
         electrum_http_url: Url,
         network: bitcoin::Network,
-        waller_dir: &Path,
+        wallet_dir: &Path,
+        private_key: PrivateKey,
     ) -> Result<Self> {
         // Workaround for https://github.com/bitcoindevkit/rust-electrum-client/issues/47.
         let config = electrum_client::ConfigBuilder::default().retry(2).build();
@@ -51,11 +52,10 @@ impl Wallet {
         let client = Client::from_config(electrum_rpc_url.as_str(), config)
             .map_err(|e| anyhow!("Failed to init electrum rpc client: {:?}", e))?;
 
-        let db = bdk::sled::open(waller_dir)?.open_tree(SLED_TREE_NAME)?;
+        let db = bdk::sled::open(wallet_dir)?.open_tree(SLED_TREE_NAME)?;
 
-        let p_key = ::bitcoin::PrivateKey::generate_default()?;
         let bdk_wallet = bdk::Wallet::new(
-            bdk::template::P2WPKH(p_key),
+            bdk::template::P2WPKH(private_key),
             None,
             network,
             db,
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -7,55 +7,96 @@ use crate::{
     execution_params::ExecutionParams,
 };
 use ::bitcoin::{util::psbt::PartiallySignedTransaction, Txid};
-use anyhow::{Context, Result};
+use anyhow::{anyhow, bail, Result};
 use async_trait::async_trait;
 use backoff::{backoff::Constant as ConstantBackoff, tokio::retry};
-use bitcoin_harness::{bitcoind_rpc::PsbtBase64, BitcoindRpcApi};
-use reqwest::Url;
-use std::time::Duration;
-use tokio::time::interval;
+use bdk::{
+    blockchain::{noop_progress, Blockchain, ElectrumBlockchain},
+    electrum_client::{Client, ElectrumApi},
+    keys::GeneratableDefaultOptions,
+    FeeRate,
+};
+use reqwest::{Method, Url};
+use serde::{Deserialize, Serialize};
+use std::{path::Path, sync::Arc, time::Duration};
+use tokio::{sync::Mutex, time::interval};
 
-#[derive(Debug)]
+const SLED_TREE_NAME: &str = "default_tree";
+
 pub struct Wallet {
-    pub inner: bitcoin_harness::Wallet,
+    pub inner: Arc<Mutex<bdk::Wallet<ElectrumBlockchain, bdk::sled::Tree>>>,
     pub network: bitcoin::Network,
+    pub http_url: Url,
+    pub rpc_url: Url,
 }
 
 impl Wallet {
-    pub async fn new(name: &str, url: Url, network: bitcoin::Network) -> Result<Self> {
-        let wallet = bitcoin_harness::Wallet::new(name, url).await?;
+    pub async fn new(
+        electrum_rpc_url: Url,
+        electrum_http_url: Url,
+        network: bitcoin::Network,
+        waller_dir: &Path,
+    ) -> Result<Self> {
+        let client = Client::new(electrum_rpc_url.as_str())
+            .map_err(|e| anyhow!("Failed to init electrum rpc client: {:?}", e))?;
+
+        let db = bdk::sled::open(waller_dir)?.open_tree(SLED_TREE_NAME)?;
+
+        let p_key = ::bitcoin::PrivateKey::generate_default()?;
+        let bdk_wallet = bdk::Wallet::new(
+            bdk::template::P2WPKH(p_key),
+            None,
+            network,
+            db,
+            ElectrumBlockchain::from(client),
+        )?;
 
         Ok(Self {
-            inner: wallet,
+            inner: Arc::new(Mutex::new(bdk_wallet)),
             network,
+            http_url: electrum_http_url,
+            rpc_url: electrum_rpc_url,
         })
     }
 
     pub async fn balance(&self) -> Result<Amount> {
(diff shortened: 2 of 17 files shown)


swap/src/config.rs
```diff
@@ -13,7 +13,9 @@ use url::Url;
 
 pub mod seed;
 
-const DEFAULT_BITCOIND_TESTNET_URL: &str = "http://127.0.0.1:18332";
+const DEFAULT_ELECTRUM_HTTP_URL: &str = "https://blockstream.info/testnet/api/";
+// TODO: Not a sensible default, copied from https://github.com/bitcoindevkit/bdk
```
da-kami, line 17: Note: @rishflab tested using this URL in a test-project and it actually worked. Might want to remove the comment?

bonomat (commented):
We should not forget that we are using the sync version right now but actually want to switch to async. 
When doing so, don't forget that the default features do not work with the async features: 
`features = [ "esplora", "async-interface"], default-features=false`

swap/src/bitcoin/wallet.rs
```diff
@@ -195,7 +262,10 @@ impl WaitForTransactionFinality for Wallet {
     }
 }
 
+#[async_trait]
 impl GetNetwork for Wallet {
+    // todo: Get this from the electrum and instead of storing it in the Wallet
```
bonomat, line 267: Unless you really want to go all the way to electrum, the bdk-wallet itself has a function to get the network.  
```
self.inner.lock().await.network()
```

swap/tests/testutils/bitcoind.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl Default for Bitcoind {
+    fn default() -> Self {
+        Bitcoind {
+            tag: "v0.19.1".into(),
+            args: BitcoindArgs::default(),
+            ports: None,
+            entrypoint: Some("/usr/bin/bitcoind".into()),
+            volume: uuid::Uuid::new_v4().to_string(),
```
bonomat, line 75: Note: your bitcoind image will always mount a volume with this approach even if `with_volume` is not being called. 
It would be better to use `Option::None`, set it to `Some(xxx)` in `with_volume` and check in the function `volumes()` if it is set.

swap/tests/testutils/bitcoind.rs
```diff
@@ -0,0 +1,137 @@
...
+            "-listen=1".to_string(),
+            "-prune=0".to_string(),
+            "-rpcallowip=0.0.0.0/0".to_string(),
+            "-rpcbind=0.0.0.0".to_string(),
+            "-rpcuser=admin".to_string(),
+            "-rpcpassword=123".to_string(),
+            "-printtoconsole".to_string(),
+            "-rest".to_string(),
+            "-fallbackfee=0.0002".to_string(),
+            "-datadir=/home/bdk-test".to_string(),
```
bonomat, line 129: To simplify the setup you could fall back to the default bitcoin path in the image which is in `/root/.bitcoin`. This would allow you to use the image without a mounted volume.

swap/tests/testutils/electrs.rs
```diff
@@ -0,0 +1,164 @@
+use bitcoin::Network;
+use std::collections::HashMap;
+use testcontainers::{
+    core::{Container, Docker, Port, WaitForMessage},
+    Image,
+};
+// pub const ELECTRS_CONTAINER_NAME: &str = "electrs";
+// pub const ELECTRS_DEFAULT_NETWORK: &str = "electrs_network";
+// pub const ELECTRS_RPC_PORT: u16 = 48081;
+// pub const ELECTRS_HTTP_PORT: u16 = 48083;
```
bonomat, line 10: Can probably be removed.

swap/tests/testutils/mod.rs
```diff
@@ -391,21 +404,163 @@ where
...
+    cli: &Cli,
+    volume: String,
+    name: String,
+    network: String,
+) -> Result<(Container<'_, Cli, bitcoind::Bitcoind>, Url)> {
+    let image = bitcoind::Bitcoind::default()
+        .with_mapped_port(Port {
+            local: 7041,
+            internal: 18443,
+        })
```
bonomat, line 449: Afaik our bitcoind image exports the default ports and you could rely on docker to provide you random ports. That way you would not run into random `address in use` issues. 
See https://github.com/testcontainers/testcontainers-rs/blob/fe51e06fb2c44b1d3e3fd627cedea3fd582487ef/src/core/container.rs#L81

swap/tests/testutils/bitcoind.rs
```diff
@@ -0,0 +1,137 @@
...
+        let args = vec![
+            "-server".to_string(),
+            "-regtest".to_string(),
+            "-txindex=1".to_string(),
+            "-listen=1".to_string(),
+            "-prune=0".to_string(),
+            "-rpcallowip=0.0.0.0/0".to_string(),
+            "-rpcbind=0.0.0.0".to_string(),
+            "-rpcuser=admin".to_string(),
+            "-rpcpassword=123".to_string(),
```
bonomat, line 125: We should use constants for username and admin or let it be autogenerated as in our coblox-bitcoin-core image: 

https://github.com/testcontainers/testcontainers-rs/blob/fe51e06fb2c44b1d3e3fd627cedea3fd582487ef/src/images/coblox_bitcoincore.rs#L51-L96

Otherwise we run in the danger of random errors because things get changed in different places.

swap/tests/testutils/electrs.rs
```diff
@@ -0,0 +1,164 @@
...
+impl Default for ElectrsArgs {
+    fn default() -> Self {
+        ElectrsArgs {
+            network: Network::Regtest,
+            daemon_dir: "/home/bdk-test/".to_string(),
+            daemon_rpc_addr: "0.0.0.0:18443".to_string(),
+            cookie: "admin:123".to_string(),
+            http_addr: "0.0.0.0:3002".to_string(),
+            electrum_rpc_addr: "0.0.0.0:60401".to_string(),
+            cors: "*".to_string(),
```
bonomat, line 136: Most of the defaults don't make sense as electrum-rs won't work then.
Afaik, electrum-rs has some smart defaults which would just connect to public bitcoin nodes.

swap/tests/testutils/mod.rs
```diff
@@ -391,21 +404,163 @@ where
...
+        .with_volume(volume)
+        .with_tag("0.19.1");
+
+    let run_args = RunArgs::default().with_name(name).with_network(network);
+
+    let docker = cli.run_with_args(image, run_args);
+
+    let bitcoind_url = {
+        let input = format!("http://{}:{}@localhost:{}", "admin", "123", 7041);
```
bonomat, line 462: See comment on the image in regards to username and password.

bonomat (changes requested):
Nice to see the test pass 👏👏.

I think there are heaps of minor things which have to be ironed out before this is can be merged.

swap/tests/testutils/mod.rs
```diff
@@ -391,21 +404,163 @@ where
...
+            internal: 60401,
+        })
+        .with_volume(volume)
+        .with_daemon_rpc_addr(bitcoind_rpc_addr)
+        .with_tag("latest");
+
+    let run_args = RunArgs::default().with_network(network);
+
+    cli.run_with_args(image, run_args)
+}
```
thomaseizinger, line 494: Passing the `Cli` into this function is weird. You are only using it in the last line of the function which means this function might as well just "create" the `Image` instance and return it, something like:

```rust
pub async fn new_electrs_image(
    volume: String,
    bitcoind_container_name: String,
) -> electrs::Electrs {
    let bitcoind_rpc_addr = format!("{}:18443", bitcoind_container_name);
    
    electrs::Electrs::default()
        .with_mapped_port(Port {
            local: 3012,
            internal: 3002,
        })
        .with_mapped_port(Port {
            local: 60401,
            internal: 60401,
        })
        .with_volume(volume)
        .with_daemon_rpc_addr(bitcoind_rpc_addr)
        .with_tag("latest")
}
```

That would simplify the signature quite a bit without majorly impacting the call-site.

thomaseizinger (approved):
Left some feedback, nothing is immediately pressing I think. Just some food for thought :)

swap/tests/testutils/electrs.rs
rishflab: Are suggesting that ElectrsArgs should be an empty struct? This image is only used by the tests in these projects. Shouldn't the defaults reflect this?

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,93 @@ use crate::{
...
             network,
+            http_url: electrum_http_url,
+            rpc_url: electrum_rpc_url,
         })
     }
 
     pub async fn balance(&self) -> Result<Amount> {
-        let balance = self.inner.balance().await?;
-        Ok(balance)
+        self.inner.lock().await.sync(noop_progress(), None)?;
```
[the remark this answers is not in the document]
rishflab: Yeh this has been noted in the list of issues. There was mention in the bdk docs of moving sync into a seperate thread and calling it in a loop.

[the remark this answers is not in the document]
da-kami: I propose the following change, instead of doing this in the wallet's balance function, where it clearly does not belong and most likely will cause blocking trouble.I would suggest to move it to the caller as @thomaseizinger suggested.

For the tests I see this right before we assert on balances. For production we might want to think about it a bit more, but as a start it might be sufficient to sync before and after a swap.

swap/src/bitcoin/wallet.rs
```diff
@@ -175,18 +239,21 @@ impl WaitForTransactionFinality for Wallet {
         txid: Txid,
         execution_params: ExecutionParams,
     ) -> Result<()> {
+        debug!("waiting for tx finality: {}", txid);
         // TODO(Franck): This assumes that bitcoind runs with txindex=1
```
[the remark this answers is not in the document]
rishflab: nope, txindex no longer needed

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,93 @@ use crate::{
...
+        electrum_rpc_url: Url,
+        electrum_http_url: Url,
+        network: bitcoin::Network,
+        datadir: &Path,
+    ) -> Result<Self> {
+        let client =
+            Client::new(electrum_rpc_url.as_str()).expect("Failed to init electrum rpc client");
+
+        let db = bdk::sled::open(datadir).expect("could not open sled db");
+        let db = db.open_tree("default_tree").expect("could not open tree");
```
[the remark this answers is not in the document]
rishflab: Need to implement conversion to anyhow::Error. Putting a todo: to do it later.

swap/tests/testutils/bitcoind.rs
rishflab: switched to constants

swap/src/bitcoin/wallet.rs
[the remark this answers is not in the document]
thomaseizinger: Are you saying sled's errors don't implement `std::error::Error`? Because anything implementing that can be converted to `anyhow::Error`.

swap/src/bitcoin/transactions.rs
```diff
@@ -83,7 +81,7 @@ impl TxLock {
         };
 
         let tx_out = TxOut {
-            value: self.inner.output[self.lock_output_vout()].value - TX_FEE,
+            value: self.inner.clone().extract_tx().output[self.lock_output_vout()].value - TX_FEE,
```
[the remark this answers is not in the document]
rishflab: ill leave a todo

swap/tests/testutils/electrs.rs
```diff
@@ -0,0 +1,164 @@
...
+        volumes
+    }
+
+    fn env_vars(&self) -> Self::EnvVars {
+        HashMap::new()
+    }
+
+    fn ports(&self) -> Option<Vec<Port>> {
+        self.ports.clone()
+    }
```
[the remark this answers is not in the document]
rishflab: ill add it to the follow up tasks

swap/src/bitcoin/transactions.rs
```diff
@@ -31,35 +31,33 @@ impl TxLock {
...
     }
 
     pub fn lock_amount(&self) -> Amount {
-        Amount::from_sat(self.inner.output[self.lock_output_vout()].value)
+        Amount::from_sat(self.inner.clone().extract_tx().output[self.lock_output_vout()].value)
     }
 
     pub fn txid(&self) -> Txid {
-        self.inner.txid()
+        self.inner.clone().extract_tx().txid()
```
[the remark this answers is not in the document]
rishflab: I left a todo to review this

swap/src/bitcoin/wallet.rs
```diff
@@ -122,17 +166,18 @@ impl WatchForRawTransaction for Wallet {
 
 #[async_trait]
 impl GetRawTransaction for Wallet {
-    // todo: potentially replace with option
     async fn get_raw_transaction(&self, txid: Txid) -> Result<Transaction> {
-        Ok(self.inner.get_raw_transaction(txid).await?)
+        self.get_tx(txid)
+            .await?
+            .ok_or_else(|| anyhow!("Could not get raw tx with id: {}", txid))
```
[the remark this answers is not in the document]
rishflab: This expression ```
   self.get_tx(txid)
            .await
            .with_context(|| format!("Could not get raw tx with id: {}", txid))?
``` still returns an option

File: swap/src/protocol/alice.rs [hunk 18]
```diff
@@ -215,6 +215,7 @@ impl Builder {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
rishflab: I am leaving this for now because it stems from sending a psbt instead of a transaction in message0.

swap/src/bitcoin/wallet.rs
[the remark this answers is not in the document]
rishflab: Sled's errors are fine. The bdk electrum client error cant be handled with ?. It doesn't seem to implement `std::error::Error`

swap/tests/testutils/mod.rs
rishflab: We let docker map the ports now. To do that we need to create the `Container` which requires the `&Cli`.

swap/tests/testutils/electrs.rs
rishflab: Used constants where I can added a todo to fix this later

rishflab (approved):

swap/src/bitcoin/timelocks.rs
```diff
@@ -49,6 +49,14 @@ impl Add<Timelock> for BlockHeight {
...
 }
 
+impl Sub<BlockHeight> for BlockHeight {
+    type Output = BlockHeight;
+
+    fn sub(self, rhs: BlockHeight) -> Self::Output {
+        BlockHeight(self.0 - rhs.0)
+    }
+}
+
```
D4nte, line 59: Please clarify why you would substract two absolute block heights?

Also, implementing `sub_checked` would be safer.

swap/src/bitcoin/transactions.rs
```diff
@@ -14,7 +14,11 @@ use std::collections::HashMap;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct TxLock {
-    inner: Transaction,
+    // todo: We only need to send the transaction here but bdk_wallet.sign() wants a psbt.
+    // Constructing the psbt from the tx was failing after the introduction of psbt so we send the
+    // full psbt over. This is not ideal and has resulted in several clones to get the txid in the
+    // rest of this file.
+    inner: PartiallySignedTransaction,
```
D4nte, line 21: Using PSBT internally is for the best. I don't see the clone as an issue. I'd recommend to remove this TODO.

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,100 @@ use crate::{
...
+use serde::{Deserialize, Serialize};
+use std::{path::Path, sync::Arc, time::Duration};
+use tokio::{sync::Mutex, time::interval};
+
+const SLED_TREE_NAME: &str = "default_tree";
 
-#[derive(Debug)]
 pub struct Wallet {
-    pub inner: bitcoin_harness::Wallet,
+    pub inner: Arc<Mutex<bdk::Wallet<ElectrumBlockchain, bdk::sled::Tree>>>,
```
D4nte, line 27: The intent was for Bob to still use a `bitcoind`. It seems we are using BDK for both Alice and Bob. Is this intentional?

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,100 @@ use crate::{
...
-        let wallet = bitcoin_harness::Wallet::new(name, url).await?;
+    pub async fn new(
+        electrum_rpc_url: Url,
+        electrum_http_url: Url,
+        network: bitcoin::Network,
+        datadir: &Path,
+    ) -> Result<Self> {
+        // todo: Implement conversion to anyhow::error so we can use ?
+        let client =
+            Client::new(electrum_rpc_url.as_str()).expect("Failed to init electrum rpc client");
```
D4nte, line 42:
```suggestion
        let client =
            Client::new(electrum_rpc_url.as_str()).map_err(|e| anyhow!("Failed to init electrum rpc client: {:?}", e))?;
```
In the mean time you do not need to use `expect`.

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,100 @@ use crate::{
...
+
+        let db = bdk::sled::open(datadir)?.open_tree(SLED_TREE_NAME)?;
```
D4nte, line 44: is `datadir`:
- The `data-dir` of the CLI? If so, we should but BDK related items in a subfolder (e.g. `bitcoin_wallet`.
- A subfolder of the CLI's `data-dir`? If so, then maybe best to call it `wallet-dir` so it's clear.

swap/src/bitcoin/wallet.rs
```diff
@@ -7,55 +7,100 @@ use crate::{
...
+
+        // todo: make key generation configurable using a descriptor
```
D4nte, line 46: Why would that be needed?

swap/src/bitcoin/wallet.rs
```diff
@@ -66,54 +111,56 @@ impl BuildTxLockPsbt for Wallet {
...
-
+        self.sync_wallet().await?;
+        tracing::debug!("building tx lock");
+        let (psbt, _details) = self.inner.lock().await.create_tx(
+            bdk::TxBuilder::with_recipients(vec![(
+                output_address.script_pubkey(),
+                output_amount.as_sat(),
+            )])
+            // todo: get actual fee
+            .fee_rate(FeeRate::from_sat_per_vb(5.0)),
```
D4nte, line 122: Move this value in a const so it's easier to remove all static fee values at once.

swap/src/bitcoin/wallet.rs
```diff
@@ -66,54 +111,56 @@ impl BuildTxLockPsbt for Wallet {
...
-        let psbt: PartiallySignedTransaction = bitcoin::consensus::deserialize(&as_hex)?;
-
-        let tx = psbt.extract_tx();
-
+        let (signed_psbt, finalized) = self.inner.lock().await.sign(psbt, None)?;
+        if !finalized {
+            bail!("Could not finalize TxLock psbt")
+        }
+        let tx = signed_psbt.extract_tx();
+        tracing::debug!("signed tx lock");
```
D4nte, line 140: This log message does not seem very useful.

swap/src/bitcoin/wallet.rs
```diff
@@ -122,17 +169,37 @@ impl WatchForRawTransaction for Wallet {
...
+        self.get_tx(txid)
+            .await?
+            .ok_or_else(|| anyhow!("Could not get raw tx with id: {}", txid))
     }
 }
 
 #[async_trait]
 impl GetBlockHeight for Wallet {
     async fn get_block_height(&self) -> BlockHeight {
+        // todo: create this url using the join() api in the Url type
```
D4nte, line 182: Just do it?

swap/src/bitcoin/wallet.rs
```diff
@@ -144,20 +211,36 @@ impl GetBlockHeight for Wallet {
...
+        let url = format!("{}tx/{}/status", self.http_url, txid);
+        #[derive(Serialize, Deserialize, Debug, Clone)]
+        struct TransactionStatus {
+            block_height: Option<u32>,
+            confirmed: bool,
+        }
+        // todo: See if we can make this error handling more elegant
+        // errors
         #[derive(Debug)]
         enum Error {
```
D4nte, line 224: Sharing this error between the traits implementation would make it more readable.

swap/src/protocol/alice.rs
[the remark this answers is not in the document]
D4nte: > I am leaving this for now because it stems from sending a psbt instead of a transaction in message0.

I do not understand your justification. Boxing it is very easy.

swap/tests/happy_path.rs
```diff
@@ -1,6 +1,7 @@
 pub mod testutils;
 
 use swap::protocol::{alice, bob};
+
 use testutils::SlowCancelConfig;
```
D4nte, line 5:
```suggestion
use swap::protocol::{alice, bob};
use testutils::SlowCancelConfig;
```

swap/src/bitcoin/wallet.rs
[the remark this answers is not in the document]
thomaseizinger: Fair, if you have to use `format!` it is probably equally noisy.

swap/src/protocol/alice.rs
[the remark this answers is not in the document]
thomaseizinger: To be fair, a PSBT is probably the nicer solution anyway because that is the format designed for collaboratively signing a transaction.

thomaseizinger (approved):
Let's do it!

swap/src/bitcoin/wallet.rs
da-kami: Clarification: We intend for Alice to use `bitcoind` - as she is the role within Nectar. 

Since Nectar was not introduced yet we opted for changing both parties and then potentially bring back `bitcoind` for Alice again. (we discussed that in the chat)
I am not sure this strategy is optimal, as it might block merging https://github.com/comit-network/xmr-btc-swap/pull/177

@thomaseizinger what do you think? Should we just bring back `bitcoind` and maintain both wallets (one for CLI = Bob, one for Nectar = Alice)?

swap/src/bitcoin/wallet.rs
rishflab: Doable but not that straightforward. join can error and this function is not meant to fail.

swap/src/bitcoin/timelocks.rs
rishflab: Yep switched to checked-sub this morning. Haven't pushed it.

We are subtracting the transaction block height with the latest block height to get the number of confirmations. BDK wallet does not have a API for this.

swap/src/bitcoin/wallet.rs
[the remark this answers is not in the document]
rishflab: mapping the error and propagating it now

swap/src/bitcoin/wallet.rs
thomaseizinger: Like we discussed on Friday, I wouldn't bother with bitcoind. Makes deploying Alice easier.

thomaseizinger:
I think 88dcfaf should be squashed into f830e30.

Also, ee7496d seems to be generally useful and we can probably get that merged before we get the whole BDK stuff in. If that patch applies cleanly onto current master, it might be worth opening a separate PR for that.

rishflab:
> I think [88dcfaf](https://github.com/comit-network/xmr-btc-swap/commit/88dcfaf266f9df65192e475844f4fcd65b21f3fa) should be squashed into [f830e30](https://github.com/comit-network/xmr-btc-swap/commit/f830e309d7558f795ab024248eaf704a661c6a81).
> 
> Also, [ee7496d](https://github.com/comit-network/xmr-btc-swap/commit/ee7496d794c8baea056d056115f15da25f614a08) seems to be generally useful and we can probably get that merged before we get the whole BDK stuff in. If that patch applies cleanly onto current master, it might be worth opening a separate PR for that.

Can do the squash. I don't think the last commit or equivalent commit can be merged or will make sense on its own.

thomaseizinger:
> > I think [88dcfaf](https://github.com/comit-network/xmr-btc-swap/commit/88dcfaf266f9df65192e475844f4fcd65b21f3fa) should be squashed into [f830e30](https://github.com/comit-network/xmr-btc-swap/commit/f830e309d7558f795ab024248eaf704a661c6a81).
> > Also, [ee7496d](https://github.com/comit-network/xmr-btc-swap/commit/ee7496d794c8baea056d056115f15da25f614a08) seems to be generally useful and we can probably get that merged before we get the whole BDK stuff in. If that patch applies cleanly onto current master, it might be worth opening a separate PR for that.
> 
> Can do the squash. I don't think the last commit or equivalent commit can be merged or will make sense on its own.

Thank you! Did you intend to have the 2nd commit message as the "primary" one? The current commit message is just the concatenation of several commits. Do you mind re-wording that to a nice summary of what this patch actually does?

rishflab:
> > > I think [88dcfaf](https://github.com/comit-network/xmr-btc-swap/commit/88dcfaf266f9df65192e475844f4fcd65b21f3fa) should be squashed into [f830e30](https://github.com/comit-network/xmr-btc-swap/commit/f830e309d7558f795ab024248eaf704a661c6a81).
> > > Also, [ee7496d](https://github.com/comit-network/xmr-btc-swap/commit/ee7496d794c8baea056d056115f15da25f614a08) seems to be generally useful and we can probably get that merged before we get the whole BDK stuff in. If that patch applies cleanly onto current master, it might be worth opening a separate PR for that.
> > 
> > 
> > Can do the squash. I don't think the last commit or equivalent commit can be merged or will make sense on its own.
> 
> Thank you! Did you intend to have the 2nd commit message as the "primary" one? The current commit message is just the concatenation of several commits. Do you mind re-wording that to a nice summary of what this patch actually does?

crap my bad, will fix

da-kami:
I ran into this error on Bob's (CLI) side when swapping on test/stagenet with the current branch:

```
2021-02-15T11:43:07.926089Z  INFO swap::protocol::bob: Start execution setup with 12D3KooWCdMKjesXMJz1SiZ7HgotrxuqhQJbP5sgBm2BwP1cqThi
2021-02-15T11:43:19.720454Z DEBUG swap::bitcoin::wallet: building tx lock
2021-02-15T11:43:19.720502Z DEBUG swap::bitcoin::wallet: syncing wallet
2021-02-15T11:43:19.923615Z DEBUG swap::bitcoin::wallet: tx lock built
2021-02-15T11:43:20.001058Z  INFO swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::protocol::bob::swap: Current state: negotiated
2021-02-15T11:43:20.001193Z DEBUG swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::protocol::bob::event_loop: Attempt to dial Alice
2021-02-15T11:43:20.002030Z DEBUG swap::protocol::bob::event_loop: Already connected to Alice: 12D3KooWCdMKjesXMJz1SiZ7HgotrxuqhQJbP5sgBm2BwP1cqThi
2021-02-15T11:43:20.004695Z DEBUG swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::bitcoin::wallet: signing tx lock: db990a50a375a0460e39f3e4e71ac454ab675a4e57f5fce5423ee12488bbdd2c
2021-02-15T11:43:20.008502Z DEBUG swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::bitcoin::wallet: signed tx lock: db990a50a375a0460e39f3e4e71ac454ab675a4e57f5fce5423ee12488bbdd2c
2021-02-15T11:43:20.008862Z  INFO swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::protocol::bob::state: db990a50a375a0460e39f3e4e71ac454ab675a4e57f5fce5423ee12488bbdd2c
2021-02-15T11:43:20.009080Z DEBUG swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::bitcoin::wallet: attempting to broadcast tx: db990a50a375a0460e39f3e4e71ac454ab675a4e57f5fce5423ee12488bbdd2c
2021-02-15T11:43:20.206863Z  INFO swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::bitcoin::wallet: Bitcoin tx broadcasted! TXID = db990a50a375a0460e39f3e4e71ac454ab675a4e57f5fce5423ee12488bbdd2c
2021-02-15T11:43:20.217932Z  INFO swap{id=ae5a5a6b-7f25-4532-b64a-dbc4497679f9}: swap::protocol::bob::swap: Current state: btc is locked
thread 'main' panicked at 'transient errors to be retried: JsonDeserialisation(reqwest::Error { kind: Decode, source: Error("expected value", line: 1, column: 1) })', swap/src/bitcoin/wallet.rs:238:10
stack backtrace:
```

It seems there can be other errors than the one we got with `confirmed` being missing.

da-kami:
Additionally I ran into the problem, that the automated swap backend gets stuck when watching for the Bitcoin transaction:

```
2021-02-15T12:37:57.658433Z DEBUG swap::protocol::alice::event_loop: Connection Established with 12D3KooWFiXr3QNrAMFpQHkPVFDFJog6ULtAmbtzykkRo265Jotm
2021-02-15T12:37:57.660929Z DEBUG swap::protocol::alice::quote_response: Received quote request from 12D3KooWFiXr3QNrAMFpQHkPVFDFJog6ULtAmbtzykkRo265Jotm
2021-02-15T12:37:57.661137Z  INFO swap::protocol::alice::behaviour: Sent quote response
2021-02-15T12:38:01.060079Z  INFO swap::protocol::alice::behaviour: Start execution setup with 12D3KooWFiXr3QNrAMFpQHkPVFDFJog6ULtAmbtzykkRo265Jotm
2021-02-15T12:38:09.348981Z  WARN swap::protocol::alice::event_loop: Could not send swap handle over channel: SendError(RemoteHandle { rx: Receiver { inner: Inner { complete: false, data: Lock { locked: false, data: UnsafeCell }, rx_task: Lock { locked: false, data: UnsafeCell }, tx_task: Lock { locked: false, data: UnsafeCell } } }, keep_running: false })
2021-02-15T12:38:09.349221Z  INFO swap{id=c0a58da1-f672-4ed2-971a-03c5f2444e6d}: swap::protocol::alice::swap: Current state:started
2021-02-15T12:38:09.349387Z DEBUG swap{id=c0a58da1-f672-4ed2-971a-03c5f2444e6d}: swap::bitcoin::wallet: watching for tx: 921df66407eb6e29852570efaf3407693a0a3693224a97e3b989e7ed03d35d39
```

Before starting to watch there is a warning printed. I was not able to investigate if that is related to the program hanging.

thomaseizinger:
> Before starting to watch there is a warning printed. I was not able to investigate if that is related to the program hanging.

I assume sure that error happens because we drop the receiving end of the swap handle in production because we only need that in the tests. @D4nte?

Should probably remove the warning because it is confusing.

da-kami:
> > Before starting to watch there is a warning printed. I was not able to investigate if that is related to the program hanging.
> 
> I assume sure that error happens because we drop the receiving end of the swap handle in production because we only need that in the tests. @D4nte?
> 
> Should probably remove the warning because it is confusing.

I talked to @D4nte about it and yes, this code piece is only concerning the tests and in production the handle is dropped. I downgraded the message to `debug` level, but would push that on a separate branch once I have a fix for Alice's watching problem.

PR #179, opened 2021-02-09
Watch only tmp monero wallet

da-kami:
When initializing Monero wallet, try to open temporary wallet, if opening errors then try to create it. If creating errors then fail with message asking the user to configure the wallet RPC.
If opening / creation succeeds get the block_height to verify that connection to wallet works, then proceed to swap. 

Note: RPC error handling is hacky, but I don't see the value of investing time into that at this point. The problem is, that we don't deal with the json rpc errors properly for calls were we don't serialize the result into an object (because the response is empty). Eventually we should switch to using https://github.com/thomaseizinger/rust-jsonrpc-client but that does not support parameters `by-name` yet, see: https://github.com/thomaseizinger/rust-jsonrpc-client/issues/20

### DIFF
--- a/monero-harness/src/rpc/wallet.rs
+++ b/monero-harness/src/rpc/wallet.rs
@@ -1,6 +1,6 @@
 use crate::rpc::{Request, Response};
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use reqwest::Url;
 use serde::{Deserialize, Serialize};
 use tracing::debug;
@@ -139,6 +139,10 @@ impl Client {
             .await?;
 
         debug!("create wallet RPC response: {}", response);
+
+        if response.contains("error") {
+            bail!("Failed to create wallet")
+        }
 
         Ok(())
     }
--- a/monero-harness/src/rpc/wallet.rs
+++ b/monero-harness/src/rpc/wallet.rs
@@ -121,6 +121,33 @@ impl Client {
         Ok(r.result)
     }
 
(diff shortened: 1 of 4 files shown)


rishflab (approved):

monero-harness/src/rpc/wallet.rs
```diff
@@ -1,6 +1,6 @@
 use crate::rpc::{Request, Response};
 
-use anyhow::Result;
+use anyhow::{bail, Result};
```
D4nte, line 3: Remove newline at line 2.

swap/src/main.rs
```diff
@@ -49,9 +50,11 @@ mod serde_peer_id;
 #[macro_use]
 extern crate prettytable;
 
+const MONERO_WATCH_ONLY_TEMP_WALLET_NAME: &str = "swap-tool-watch-only-tmp-wallet";
```
D4nte, line 53: The "temp" qualifier does not tell much about what this wallet is for.
Instead I'd suggest to name it "Watch Only Wallet" or "Blockchain Read Wallet" or "Watch Blockchain WAllet"

thomaseizinger (approved):
Agree with @D4nte's proposal of replacing "temporary" with "monitoring".

swap/src/main.rs
da-kami: I went for `swap-tool-blockchain-monitoring-wallet` because I feel it transports the meaning. Bit long but I don't mind about that :)

PR #180, opened 2021-02-10
Deposit command

da-kami:


swap/src/main.rs
```diff
@@ -297,6 +303,7 @@ async fn main() -> Result<()> {
                 bitcoin_network,
                 &wallet_data_dir,
                 monero_network,
+                seed,
```
D4nte, line 306: This is a symptom that maybe `init_wallet` is trying to do too much.

swap/src/seed.rs
```diff
@@ -18,6 +22,29 @@ impl Seed {
         Ok(Seed(bytes))
     }
 
+    /// Return the private key and chain code to be used as root extended
+    /// private key for a BIP32 wallet.
+    pub fn root_private_key(&self, network: bitcoin::Network) -> PrivateKey {
```
D4nte, line 27: We should use BDK here. Why are we calculating this manually?

D4nte (dismissed):
This PR does not make sense without BDK.

swap/src/bitcoin/wallet.rs
```diff
@@ -36,15 +35,14 @@ impl Wallet {
...
         // todo: Implement conversion to anyhow::error so we can use ?
         let client =
             Client::new(electrum_rpc_url.as_str()).expect("Failed to init electrum rpc client");
 
         let db = bdk::sled::open(datadir)?.open_tree(SLED_TREE_NAME)?;
 
-        // todo: make key generation configurable using a descriptor
-        let p_key = ::bitcoin::PrivateKey::generate_default()?;
         let bdk_wallet = bdk::Wallet::new(
             bdk::template::P2WPKH(p_key),
```
thomaseizinger, line 47: This TODO should be solved by using a descriptor that supports derivation. Most likely BIP84 so you are on native-segwit: https://docs.rs/bdk/0.3.0/bdk/descriptor/template/struct.BIP84.html

thomaseizinger: You will need an `ExtendedPrivKey` as an argument then instead of a normal `PrivateKey`.

swap/src/seed.rs
thomaseizinger: This is the right approach, it is just not fully complete:

1. We might want to create _sub-seed_ from where we derive the root private key, i.e.:
`[root-seed]` -> `[bitcoin-root-key-seed]` -> `b"Bitcoin seed"[bitcoin-root-key-seed]`
2. The return value of this function should be an `ExtendedPrivKey`.

An `ExtendedPrivKey` already does the last step of the derivation for you, see here: https://docs.rs/bitcoin/0.26.0/src/bitcoin/util/bip32.rs.html#491-510

As such, all you need to do is `[root-seed]` -> `[bitcoin-root-key-seed]` and then call `ExtendedPrivKey::new(network, [bitcoin-root-key-seed])`. (Pseudo-code, "array notation" is meant to be placeholder for binary data.)

swap/src/main.rs
```diff
@@ -334,6 +365,7 @@ async fn init_wallets(
     bitcoin_network: bitcoin::Network,
     bitcoin_wallet_data_dir: &Path,
     monero_network: monero::Network,
+    seed: Seed,
```
thomaseizinger, line 368: Isn't the `seed` loaded from the `config_path`? We are we passing it in separately?

thomaseizinger (changes requested):

swap/src/main.rs
da-kami: The seed is stored in the data-dir, that can be configured with `--data-dir` and is different than `--config` which specifies the path to the configuration file. Note that on Linux the default paths are also not the same :)

Additionally, we use the seed for two things, for deriving the peer-id and the wallet private key. The seed is read at the very top of main.

swap/src/bitcoin/wallet.rs
da-kami: Sweet, then it is exactly like the Nectar setup :)
We did use Nectar as an inspiration yesterday, but did not fully think it through and only used the private key, not the extended one in the end. I will see to fix it.

swap/src/main.rs
da-kami: Agreed. This needs some untangling that I am planning to tackle in cooldown: https://github.com/comit-network/xmr-btc-swap/issues/181

swap/src/seed.rs
D4nte: Can you clarify the relation with BDK? Do you derive manually and pass the result to BDK?

swap/src/main.rs
thomaseizinger: It is still inconsistent then. I would expect all handling of config / data dir _paths_ to be on the same abstraction level. If this component receives the _config_ path as a parameter, it should also take the _data-dir_ path as a parameter.

Alternatively, if you are reading the seed outside and passing it in, I would also expect whatever the function does with the config path to do that outside and pass it in.

swap/src/seed.rs
thomaseizinger: BDK is descriptor-based wallet and as such is initialized with a descriptor. To support derivation, we need to give it a descriptor that supports derivation, probably BIP84 to be native-segwit.

A BIP84 descriptor still needs an extended private key (this patch calls it "root" key) that needs to come from somewhere. The only thing that is missing here is that we should be deriving an `ExtenedPrivKey` from our seed and not a single `PrivateKey`.

BDK will not tell you how to create an `ExtendedPrivKey`, it just requires it as a parameter. As such, the code that we see here cannot be offloaded to BDK because it is out-of-scope. It can be partly offloaded to `rust-bitcoin` (see the link in my last comment).

See the other comment for more context + links.

swap/src/main.rs
da-kami: I totally agree. We should not pass a config path here, but should the the config!
Loading the config inside `init_wallets` was a mistake (that I introduced...) because at the moment only the wallet URLs are stored in the config. So I handled the initial setup (as in querying the user for said URLs) in `init_wallets`. 
But that is generally not a good abstraction. It causes troubles with being able to create only one wallet for specific commands as well. I tried to sum this up in https://github.com/comit-network/xmr-btc-swap/issues/181

thomaseizinger: Thank you for elaborating and creating the ticket!

rishflab:
Given that we are no longer going to have a deposit command, this PR is no longer need. This commit [f2a8e70](https://github.com/comit-network/xmr-btc-swap/pull/178/commits/f2a8e70295eb6a54d81355e6838d93779c836ed6) should allow us to fund the swaps for testing (@da-kami).

PR #182, opened 2021-02-10
Segregate CLI specific modules

D4nte:
To prepare for the introduction of nectar modules.


swap/src/seed.rs
```diff
@@ -1,6 +1,14 @@
...
+use pem::{encode, Pem};
 use rand::prelude::*;
-use std::fmt;
+use std::{
+    ffi::OsStr,
+    fmt,
+    fs::{self, File},
+    io::{self, Write},
+    path::{Path, PathBuf},
+};
```
thomaseizinger, line 11: Looking at the dependencies and the functionality, this module could be a candidate to be its own crate (for now, just in the workspace) but maybe later on also on crates.io.

- We've basically imported that thing from comit-rs and it hasn't changed much since then
- Minimal dependencies
- Provides a clear value

thomaseizinger (approved):

swap/src/seed.rs
thomaseizinger: `seedfile` is available on crates.io :)

D4nte: Good point.

da-kami (approved):

swap/src/seed.rs
D4nte: > `seedfile` is available on crates.io :)

Can't find it...

PR #190, opened 2021-02-15
Do not pass Monero amount to the CLI

D4nte:
The CLI user only pass the Bitcoin amount they want to sell.

The CLI then do a quote request to nectar which provides the Monero amount the taker can get.


swap/src/bitcoin/timelocks.rs
```diff
@@ -55,3 +55,44 @@ pub enum ExpiredTimelocks {
...
+            pub const fn new(number_of_blocks: u32) -> Self {
+                Self(Timelock::new(number_of_blocks))
+            }
+        }
+
+        impl From<$type> for u32 {
+            fn from(timelock: $type) -> Self {
+                timelock.0 .0
+            }
+        }
```
thomaseizinger, line 87: Such conversions on newtypes take some of their power away.

One way of dealing with this is to try and define the newtype in the same module as where the internals are accessed. In this case, I think only `TxCancel` needs access to the internals. If you would define `CancelTimelock` next to `TxCancel`, you could naturally access the private field without having to expose it publicly.

That would slightly increase the guarantees you get out of the newtype because you can isolate the code that gets "superpowers" and can access internals.

thomaseizinger: A newtype should always enforce its invariants across its full public API. If we can get the internals out and re-construct the newtype from its internals, there is no way we can enforce our invariants.

swap/src/protocol/bob/swap.rs
```diff
@@ -368,21 +374,27 @@ async fn run_until_internal(
...
 ) -> Result<bob::state::State2> {
-    tracing::trace!("Starting negotiate");
     event_loop_handle
-        .send_quote_request(QuoteRequest {
-            btc_amount: amounts.btc,
-        })
+        .send_quote_request(QuoteRequest { btc_amount })
         .await?;
 
-    // TODO: Use this once Bob's CLI is modified to only pass xmr amount in
```
thomaseizinger, line 383: :partying_face:

thomaseizinger (approved):

PR #191, opened 2021-02-15
Use Remote handle to access ongoing swaps on Alice

D4nte:
This helps stabilizing the test as sometimes we were checking Alice state (in the DB) too early.

### DIFF
--- a/swap/src/protocol/alice/event_loop.rs
+++ b/swap/src/protocol/alice/event_loop.rs
@@ -6,20 +6,23 @@ use crate::{
     network::{transport, TokioExecutor},
     protocol::{
         alice,
-        alice::{Behaviour, Builder, OutEvent, QuoteResponse, State0, State3, TransferProof},
+        alice::{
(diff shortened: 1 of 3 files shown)


rishflab (approved):

swap/src/protocol/alice/event_loop.rs
```diff
@@ -257,7 +265,14 @@ impl EventLoop {
         .build()
         .await?;
 
-        tokio::spawn(async move { alice::run(swap).await });
+        let (remote, remote_handle) = alice::run(swap).remote_handle();
+        tokio::spawn(async move { remote.await });
```
thomaseizinger, line 269: That should also do the trick, no?

```suggestion
        tokio::spawn(remote);
```

thomaseizinger (approved):
Well done!

As stated, slight increase of technical debt because we had to alter release code for the test. I'll take a look at this once I get a better overview of the codebase.

rishflab (approved):

PR #200, opened 2021-02-16
Wait for refund if insufficient Monero is locked up

da-kami:
In a scenario where Alice does not lock up sufficient funds Bob should properly transition to refunds. At the moment the CLI just panics. 
I noticed this when Alice accidentally had a different amount set than Bob. In the future this should not happen, because Alice provides the amount for Bob. However, in case Alice is malicious Bob should still transition correctly.

### DIFF
--- a/swap/src/protocol/bob/state.rs
+++ b/swap/src/protocol/bob/state.rs
@@ -7,7 +7,7 @@ use crate::{
     },
     execution_params::ExecutionParams,
     monero,
-    monero::{monero_private_key, TransferProof},
+    monero::{monero_private_key, InsufficientFunds, TransferProof},
     protocol::{
         alice::{Message1, Message3},
         bob::{EncryptedSignature, Message0, Message2, Message4},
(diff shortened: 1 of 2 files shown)


thomaseizinger (commented):
Is this really the whole change? I don't quite understand where `InsufficientFunds` is created and how it panicked before.

da-kami:
> Is this really the whole change? I don't quite understand where `InsufficientFunds` is created and how it panicked before.

Insufficient funds was triggered in a function inside, and we just questionmarked it. But I think the error handling should be in `run_until_internal`, because there we have to take the decision.
So I used a two-fold Result again to be able to go on in case the problem was related to insufficient funds.

thomaseizinger:
> > Is this really the whole change? I don't quite understand where `InsufficientFunds` is created and how it panicked before.
> 
> Insufficient funds was triggered in a function inside, and we just questionmarked it in `run_until_internal`. But I think the error handling should be in `run_until_internal`, because there we have to take the decision.
> So I used a two-fold Result to be able to go on in case the problem was related to insufficient funds.

But you are still questionmarking the error from `watch_for_transfer`?

thomaseizinger:
@da-kami fixed it for you and rebased onto latest master to resolve the conflict.

da-kami:
> @da-kami fixed it for you and rebased onto latest master to resolve the conflict.

The rebase alters `ci.yml` - is that anticipated?

PR #203, opened 2021-02-16
Introduce dynamic rates

D4nte:


swap/src/nectar/fixed_rate.rs
```diff
@@ -6,8 +6,10 @@ pub const RATE: f64 = 0.01;
 pub struct RateService(Rate);
 
 impl LatestRate for RateService {
-    fn latest_rate(&mut self) -> Rate {
-        self.0
+    type Error = anyhow::Error;
```
thomaseizinger, line 9: I think this should be `Infallible`.

thomaseizinger (approved):

swap/src/nectar/fixed_rate.rs
thomaseizinger: Posted a follow-up here: https://github.com/comit-network/xmr-btc-swap/pull/212

da-kami (approved):

da-kami:
With https://github.com/comit-network/xmr-btc-swap/pull/203/commits/9496dce91708f6196074097320387c221f48ad10 we skip heartbeat messages that have a certain format. 

Since we are only interested in the ticker data and know the exact heartbeat message I felt this was the solution to go for (ignoring the specific heartbeat message).

swap/src/asb/kraken.rs
```diff
@@ -87,6 +87,11 @@ impl RateService {
                     }
                 };
 
+                // If we encounter a heartbeat we skip it and iterate again
+                if msg.eq(r#"{"event":"heartbeat"}"#) {
```
thomaseizinger, line 91: How shit is this API? :sweat_smile: 

Do I see it correctly that heartbeat messages have an `event` field and regular responses don't?
Otherwise we could have just made an enum where we deserialize the response and `match` on it in a strongly typed fashion instead of this string comparison here.

thomaseizinger (approved):
Thanks @da-kami !

thomaseizinger:
> With [9496dce](https://github.com/comit-network/xmr-btc-swap/commit/9496dce91708f6196074097320387c221f48ad10) we skip heartbeat messages that have a certain format in production code.
> 
> Since we are only interested in the ticker data and know the exact heartbeat message I felt this was the solution to go for (jumping back to the start of the loop upon a specific heartbeat message).
> We could add a mock service to mimic the behavior for testing, but I am not sure that's worth the effort - what do you think @thomaseizinger ?

No need for the mock-service IMO.

What might be useful is a small CLI that uses this code and simply prints out what we get on the other end of the channel. That could be useful in debugging and or extending this piece of code.

swap/src/asb/kraken.rs
da-kami: yes :(

> > With [9496dce](https://github.com/comit-network/xmr-btc-swap/commit/9496dce91708f6196074097320387c221f48ad10) we skip heartbeat messages that have a certain format in production code.
> > Since we are only interested in the ticker data and know the exact heartbeat message I felt this was the solution to go for (jumping back to the start of the loop upon a specific heartbeat message).
> > We could add a mock service to mimic the behavior for testing, but I am not sure that's worth the effort - what do you think @thomaseizinger ?
> 
> No need for the mock-service IMO.
> 
> What might be useful is a small CLI that uses this code and simply prints out what we get on the other end of the channel. That could be useful in debugging and or extending this piece of code.

Agreed - do you see this as e.g. a separate crate at some point so we can use it as a lib in code but also provide the simple CLI?

da-kami:
> > With [9496dce](https://github.com/comit-network/xmr-btc-swap/commit/9496dce91708f6196074097320387c221f48ad10) we skip heartbeat messages that have a certain format in production code.
> > Since we are only interested in the ticker data and know the exact heartbeat message I felt this was the solution to go for (jumping back to the start of the loop upon a specific heartbeat message).
> > We could add a mock service to mimic the behavior for testing, but I am not sure that's worth the effort - what do you think @thomaseizinger ?
> 
> No need for the mock-service IMO.
> 
> What might be useful is a small CLI that uses this code and simply prints out what we get on the other end of the channel. That could be useful in debugging and or extending this piece of code.

Agreed - do you see this as e.g. a separate crate at some point so we can use it as a lib in code but also provide the simple CLI?

thomaseizinger:
> > > With [9496dce](https://github.com/comit-network/xmr-btc-swap/commit/9496dce91708f6196074097320387c221f48ad10) we skip heartbeat messages that have a certain format in production code.
> > > Since we are only interested in the ticker data and know the exact heartbeat message I felt this was the solution to go for (jumping back to the start of the loop upon a specific heartbeat message).
> > > We could add a mock service to mimic the behavior for testing, but I am not sure that's worth the effort - what do you think @thomaseizinger ?
> > 
> > 
> > No need for the mock-service IMO.
> > What might be useful is a small CLI that uses this code and simply prints out what we get on the other end of the channel. That could be useful in debugging and or extending this piece of code.
> 
> Agreed - do you see this as e.g. a separate crate at some point so we can use it as a lib in code but also provide the simple CLI?

I was thinking of a CLI next to `asb` and `swap_cli` that just acts as another entry point. I don't think it is generally useful enough to be in its own crate that gets released.

PR #207, opened 2021-02-17
Make bitcoin tx fee configurable

rishflab:
NOT READY FOR REVIEW:

Alice specifies the fee rate in the quote response to Bob. Bob's CLI tool should calculate and show the fees (along with the rate) that he will pay for a swap. Bob can decide whether he wishes to proceed with the swap. Currently the fee rate is hardcoded for Alice and the tests are failing because the balance assertions assume fixed tx fees. Remove fee rate from BuildTxLockPsbt as we manually specify the fees in build_spend_transaction().

### DIFF
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -35,14 +35,7 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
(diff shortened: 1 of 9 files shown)


D4nte:
Have you considered using RBF or CPFP?

rishflab:
> Have you considered using RBF or CPFP?

I don't think it is necessary at the moment and may introduce additional complexity

PR #213, opened 2021-02-19
Properly calculate the confirmations for Bitcoin tx

da-kami:
Once the transaction was included into a block it has one confirmation - before inclusion it has zero.
current-block-height - transaction-block-height = zero; but that means one confirmation.
Hence, the confirmation calculation was adapted to: Current-block-height - (transaction-block-height - 1).

I could not think of a more elegant solution - if you have one please comment :)

### DIFF
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -273,7 +273,11 @@ impl WaitForTransactionFinality for Wallet {
             tracing::debug!("tx_block_height: {:?}", tx_block_height);
             let block_height = self.get_block_height().await?;
             tracing::debug!("latest_block_height: {:?}", block_height);
-            if let Some(confirmations) = block_height.checked_sub(tx_block_height) {
+            if let Some(confirmations) = block_height.checked_sub(
+                tx_block_height
+                    .checked_sub(BlockHeight::new(1))
+                    .expect("transaction must be included in block with height >= 1"),
+            ) {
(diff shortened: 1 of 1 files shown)


rishflab (approved):

swap/src/bitcoin/wallet.rs
```diff
@@ -273,7 +273,11 @@ impl WaitForTransactionFinality for Wallet {
             tracing::debug!("tx_block_height: {:?}", tx_block_height);
             let block_height = self.get_block_height().await?;
             tracing::debug!("latest_block_height: {:?}", block_height);
-            if let Some(confirmations) = block_height.checked_sub(tx_block_height) {
+            if let Some(confirmations) = block_height.checked_sub(
+                tx_block_height
+                    .checked_sub(BlockHeight::new(1))
+                    .expect("transaction must be included in block with height >= 1"),
```
thomaseizinger, line 279: Not sure if it is more _elegant_ but it you chose to panic anyway, you could also use the non-checked API with `-`.

thomaseizinger (approved):

swap/src/bitcoin/wallet.rs
da-kami: I was more thinking in terms of changing the logic/naming on a larger scale to avoid this calculation. I like the `checked_sub` because it provides more context on what is going on with this calculation. Does not add anything to the runtime constraints though :)

PR #216, opened 2021-02-22
To avoid CI failure wait for the balance instead of sleep

da-kami:


thomaseizinger (approved):
As discussed, let's create an issue to track the removal of this code once we want to have this in the prod code as part of the whole flow.

PR #220, opened 2021-02-22
asb setup - systemd service files and readme

da-kami:
With this draft PR I wanted to show the systemd services in a draft-PR for discussion on improvements and sharing knowledge on the current setup.

Not sure if we actually want to merge this in it's current form as it is quite specific to the setup on the raspberry.
Depending on how the project evolves this we can iterate on making these services more independent of the underlying setup so that they could be reused by other users for setup.


docs/setup/asb/systemd/pull-and-build-asb.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Pull and build ASB from source
```
thomaseizinger, line 2: I think that is too specific to our setup.

docs/setup/asb/systemd/restart-asb.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Restart ASB service
+After=watch-asb-binary-change.path
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl restart start-asb.service
+ExecStopPost=/bin/sh -c 'if [ "$$EXIT_STATUS" -ne 0 ]; then /usr/local/bin/systemd-email.sh daniel@coblox.tech %n; fi'
```
thomaseizinger, line 8: That should just be an `OnFailure` no?

docs/setup/asb/systemd/start-asb.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Run the automated swap backend
+After=start-monero-wallet-rpc.service
+
+[Service]
+WorkingDirectory=/var/external/asb/xmr-btc-swap
+ExecStart=/home/asb/.cargo/bin/cargo run --release --bin asb -- --config /var/external/asb/run/asb-config/config.toml start
```
thomaseizinger, line 7: Why do `cargo run` and not just run the built binary?

docs/setup/asb/systemd/start-asb.service
```diff
@@ -0,0 +1,16 @@
...
+ExecStopPost=/bin/sh -c 'if [ "$$EXIT_STATUS" -ne 0 ]; then /usr/local/bin/systemd-email.sh daniel@coblox.tech %n; fi'
```
thomaseizinger, line 8: Again, should be an `OnFailure` I think.

docs/setup/asb/systemd/start-asb.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Run the automated swap backend
+After=start-monero-wallet-rpc.service
```
thomaseizinger, line 3: Probably want a `Requires` as well to enforce that this doesn't start if the wallet fails.

docs/setup/asb/systemd/start-asb.service
da-kami: I just did it like this because that way we don't have to have an explicit initial build. It does not really make a difference because Cargo will only build it new if there are changes.

docs/setup/asb/systemd/restart-asb.service
da-kami: Yeah, seems I should have digged a little deeper. 

`OnFailure` is an attribute of the `Unit` section, not the `Service` section. `OnFailure` can only trigger another service, so in order to achieve this behavior I'd have to configure a notification service similar to what is described here: https://serverfault.com/questions/694818/get-notification-when-systemd-monitored-service-enters-failed-state

I will do that.

docs/setup/asb/systemd/pull-and-build-asb.service
da-kami: That is what the service does at the moment. It's true that a service provider would not need that. 
It is not only this file that is concerned by this task, but also the timer, path-watcher and the restart service. Should I put those files into a separate folder?
It's still nice to have them in GitHub if only for us to track what is changing.

docs/setup/asb/systemd/restart-asb.service
da-kami: Added an additional templated service `unit-status-email@.service` that can be triggered on `OnFailure` by any other service like this: `OnFailure=unit-status-email@%n.service`

Where `%n` is the full service name of the caller (passed as parameter to `unit-status-email@.service`).

da-kami:
@thomaseizinger do you think it's worth merging these service files?  Should I separate the service files related to building and restarting the asb periodically into a separate folder (or remove them)?

I personally like to have these files tracked here, but in the end I am not sure anyone besides us would use them at the moment.

da-kami:
The current setup is too specific for our cause. Does not add value to share this in the project yet.

PR #225, opened 2021-02-24
Asb refund reload main wallet

### LINKED ISSUE #211: Monero wallet management for `refund` scenarios
## Problem

The ASB can currently not avoid refund scenarios. If Bob does not go through with the swap and refunds (for whatever reason) Alice will refund as well.
When Alice refunds she uses `generate_from_keys` to claim the XMR - which unloads her `default` wallet. There is currently no logic to reload the `default` wallet - the wallet created by `refund` will stay loaded which can affect other swaps currently ongoing concurrently. 

We don't expect these scenarios to happen very often, but nevertheless we have to fix Alice to reload her `default` wallet so she has funds for swapping after a `refund` scenario. 

## Potential Solution

As soon as `generate_from_keys` finishes the `default` wallet is reloaded (using `open_wallet`).
Additionally the wallet should be wrapped in a Mutex to avoid unguarded concurrent access. 
Note: Make sure that `generate_from_keys` is called with an appropriate `rescan_height` for refund scenarios - otherwise creating the refund wallet can take significant amounts of time on stage/mainnet!

da-kami:
Fixes #211 

Please review by commit and read the commit messages.

### DIFF
--- a/monero-rpc/src/rpc/wallet.rs
+++ b/monero-rpc/src/rpc/wallet.rs
@@ -268,14 +268,8 @@ impl Client {
         address: &str,
         spend_key: &str,
         view_key: &str,
-        restore_height: Option<u32>,
+        restore_height: u32,
     ) -> Result<GenerateFromKeys> {
-        let restore_height = if let Some(restore_height) = restore_height {
-            restore_height
-        } else {
-            0
-        };
-
         let params = GenerateFromKeysParams {
             restore_height,
             filename: view_key.into(),
--- a/swap/src/monero.rs
+++ b/swap/src/monero.rs
@@ -8,6 +8,8 @@ use crate::bitcoin;
 use ::bitcoin::hashes::core::fmt::Formatter;
 use anyhow::Result;
 use async_trait::async_trait;
+use monero::Address;
+use monero_rpc::wallet::{BlockHeight, Refreshed};
 use rand::{CryptoRng, RngCore};
 use rust_decimal::{
     prelude::{FromPrimitive, ToPrimitive},
@@ -226,6 +228,21 @@ pub trait OpenWallet {
 #[async_trait]
 pub trait CreateWallet {
     async fn create_wallet(&self, file_name: &str) -> Result<()>;
+}
+
+#[async_trait]
+pub trait WalletBlockHeight {
+    async fn block_height(&self) -> Result<BlockHeight>;
+}
+
+#[async_trait]
+pub trait GetAddress {
+    async fn get_main_address(&self) -> Result<Address>;
+}
+
+#[async_trait]
+pub trait Refresh {
+    async fn refresh(&self) -> Result<Refreshed>;
 }
 
 #[derive(thiserror::Error, Debug, Clone, PartialEq)]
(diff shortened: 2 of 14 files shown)


monero-rpc/src/rpc/wallet.rs
```diff
@@ -268,14 +268,8 @@ impl Client {
         address: &str,
         spend_key: &str,
         view_key: &str,
-        restore_height: Option<u32>,
+        restore_height: u32,
     ) -> Result<GenerateFromKeys> {
-        let restore_height = if let Some(restore_height) = restore_height {
```
thomaseizinger, line 273: This could have been an `unwrap_or_else` from the beginning.

swap/src/monero.rs
```diff
@@ -228,6 +240,21 @@ pub trait CreateWallet {
     async fn create_wallet(&self, file_name: &str) -> Result<()>;
 }
 
+#[async_trait]
+pub trait WalletBlockHeight {
```
thomaseizinger, line 244: Are we ever using these traits meaningfully to create new abstractions? Differently asked: Is there a trait-bound somewhere that uses this trait?

If not, then I would propose to simplify this and just make these `pub` functions on the wallet type itself.

That doesn't break encapsulation aslong as all fields are still private, which they should be anyway, regardless of the traits.

swap/src/monero/wallet.rs
```diff
@@ -1,38 +1,45 @@
...
     time::Duration,
 };
+use tokio::sync::Mutex;
 use tracing::info;
 use url::Url;
 
 #[derive(Debug)]
 pub struct Wallet {
-    pub inner: wallet::Client,
+    pub inner: Mutex<wallet::Client>,
```
thomaseizinger, line 26: These should all be private.

thomaseizinger (approved):
Thanks @da-kami !

Clean commits, very easy to follow! Also, good balance between refactoring and delivering the actual feature :)

None of my comments are blocking, feel free to act in a follow-up or just leave them as well :D

swap/src/monero/wallet.rs
da-kami: I totally agree and actually had made them private, but then I would have to add a separate constructor for the tests. Maybe more generally asked: How shall we deal with the wallet in the tests? - I find it weird that we have to "make things public" for the test...

swap/src/monero/wallet.rs
```diff
@@ -129,15 +172,17 @@ impl WatchForTransfer for Wallet {
         }
 
         let address = Address::standard(self.network, public_spend_key, public_view_key.into());
-        let wallet = self.inner.clone();
```
da-kami, line 132: @thomaseizinger note that I changed the way the mutex is locked here (within each retry rather than locking the wallet beforehand) because I think it will behave differently and the wallet should not be locked for the complete run of the function. However, I am not totally sure the behavior really changed. Would be nice if we can think about this together tomorrow.

swap/src/monero.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/225/commits/0945cee459aeae17f3e7234ec3aa49cb26293c5d

swap/src/monero/wallet.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/225/commits/578d23d7fc9d2f703e96384a2b3e911aaec176a6

PR #230, opened 2021-02-25
Use testnet demo defaults if config file or cli args are not provided

### LINKED ISSUE #229: Replace setup wizard with default values

rishflab:
Closes #229 

Also connects to the ASB running by default if peer id and addr are not supplied in the cli args 

`swap_cli -- buy-xmr --send-btc 0.0006` will trigger a swap

### DIFF
--- a/swap/src/cli/command.rs
+++ b/swap/src/cli/command.rs
@@ -4,6 +4,9 @@ use libp2p::{core::Multiaddr, PeerId};
 use std::path::PathBuf;
 use uuid::Uuid;
 
+const DEFAULT_ALICE_MULTIADDR: &str = "/dns4/xmr-btc-asb.coblox.tech/tcp/9876";
+const DEFAULT_ALICE_PEER_ID: &str = "12D3KooWCdMKjesXMJz1SiZ7HgotrxuqhQJbP5sgBm2BwP1cqThi";
+
 #[derive(structopt::StructOpt, Debug)]
 pub struct Arguments {
     #[structopt(
@@ -21,10 +24,13 @@ pub struct Arguments {
 #[structopt(name = "xmr_btc-swap", about = "XMR BTC atomic swap")]
 pub enum Command {
     BuyXmr {
-        #[structopt(long = "connect-peer-id")]
+        #[structopt(long = "connect-peer-id", default_value = DEFAULT_ALICE_PEER_ID)]
         alice_peer_id: PeerId,
 
-        #[structopt(long = "connect-addr")]
+        #[structopt(
+            long = "connect-addr",
+            default_value = DEFAULT_ALICE_MULTIADDR
+        )]
         alice_addr: Multiaddr,
 
         #[structopt(long = "send-btc", help = "Bitcoin amount as floating point nr without denomination (e.g. 1.25)", parse(try_from_str = parse_btc))]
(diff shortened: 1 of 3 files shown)


swap/src/cli/config.rs
```diff
@@ -33,6 +31,27 @@ impl Config {
...
+                    .expect("default electrum http str is a valid url"),
+                electrum_rpc_url: DEFAULT_ELECTRUM_RPC_URL
+                    .parse()
+                    .expect("default electrum rpc str is a valid url"),
+            },
+            monero: Monero {
+                wallet_rpc_url: DEFAULT_MONERO_WALLET_RPC_TESTNET_URL
+                    .parse()
+                    .expect("default monero wallet rpc testnet str is a valid url"),
+            },
```
thomaseizinger, line 52: Need to rebase, this is already gone on master :)

swap/src/cli/command.rs
```diff
@@ -22,10 +22,10 @@ pub struct Arguments {
 pub enum Command {
     BuyXmr {
         #[structopt(long = "connect-peer-id")]
-        alice_peer_id: PeerId,
+        alice_peer_id: Option<PeerId>,
 
         #[structopt(long = "connect-addr")]
-        alice_addr: Multiaddr,
+        alice_addr: Option<Multiaddr>,
```
thomaseizinger, line 28: Structopt supports default values, that might make the whole thing more concise because you don't need to do `map_or` everywhere. https://docs.rs/structopt/0.3.21/structopt/index.html#how-to-derivestructopt

thomaseizinger (approved):
Nice, thank you!

swap/src/cli/command.rs
```diff
@@ -90,3 +111,11 @@ fn parse_btc(str: &str) -> Result<bitcoin::Amount> {
     let amount = bitcoin::Amount::from_str_in(str, ::bitcoin::Denomination::Bitcoin)?;
     Ok(amount)
 }
+
+fn parse_multiaddr(str: &str) -> Multiaddr {
```
thomaseizinger, line 115: Are you certain these are necessary? I think structopt uses `FromStr` by default so you should be able to get away with not passing a parsing function at all.

Haven't tried it though :)

swap/src/cli/command.rs
```diff
@@ -90,3 +111,11 @@ fn parse_btc(str: &str) -> Result<bitcoin::Amount> {
...
+    str.parse().expect("default alice multiaddr is valid")
```
da-kami, line 116: I think the `expect` message `"default alice multiaddr is valid"` is incorrect. `parse_multiaddr` is not only used in the default case but also when an address is provided by the user. This should be a better error message!

`structopt` is able to handle `anyhow::Result` - have a look at how we parse the Bitcoin amount - I think this should be similar and the actual error should be yielded if the parsing fails.

swap/src/cli/command.rs
```diff
@@ -90,3 +111,11 @@ fn parse_btc(str: &str) -> Result<bitcoin::Amount> {
...
+}
+
+fn parse_peer_id(str: &str) -> PeerId {
+    str.parse().expect("default alice peer id is valid")
```
da-kami, line 120: Same as comment for the other prase

swap/src/cli/command.rs
da-kami: Event better that what I suggested below :D

rishflab: Yep you are right

swap/src/cli/command.rs
rishflab: fixed by using default parse from_str

thomaseizinger (approved):
Awesome! Just needs a rebase to fixup the commits and then let's get this in!

da-kami (approved):

PR #231, opened 2021-02-25
Error only on close message when fetching the rate

da-kami:
Ping/Pong messages disturb the rate requests quite frequently resulting in failed swap setup because there is no rate available.

As a result messages Ping, Pong and Binary are now ignored and not reported as error.

### DIFF
--- a/swap/src/asb/kraken.rs
+++ b/swap/src/asb/kraken.rs
@@ -7,7 +7,8 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::convert::TryFrom;
 use tokio::sync::watch;
-use tokio_tungstenite::tungstenite::Message;
+use tokio_tungstenite::tungstenite::{protocol::CloseFrame, Message};
+use tracing::{error, trace};
 use watch::Receiver;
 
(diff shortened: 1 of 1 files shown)


swap/src/asb/kraken.rs
```diff
@@ -77,8 +77,11 @@ impl RateService {
             while let Some(msg) = rate_stream.next().await {
                 let msg = match msg {
                     Ok(Message::Text(msg)) => msg,
+                    Ok(Message::Close(..)) => {
+                        let _ = rate_update.send(Err(Error::CloseMessage));
```
thomaseizinger, line 81: I think we should log an error here and also investigate the `CloseFrame` if it has useful information.

swap/src/asb/kraken.rs
```diff
@@ -77,8 +77,11 @@ impl RateService {
...
+                        continue;
+                    }
                     Ok(_) => {
```
thomaseizinger, line 84: A trace log on `Ping` and `Pong` might be useful.

thomaseizinger (approved):

rishflab (approved):

thomaseizinger (approved):
bors r+

PR #232, opened 2021-02-25
ASB only sends quote response if sufficient XMR balance

da-kami:


da-kami:
Monero harness test failed (https://github.com/comit-network/xmr-btc-swap/pull/232/checks?check_run_id=1975587854) - unfortunately activating debug did not add much value - still unclear what is the problem.

swap/src/monero.rs
```diff
@@ -206,12 +206,18 @@ pub trait WatchForTransfer {
...
 #[derive(Debug, Clone, Copy, thiserror::Error)]
-#[error("transaction does not pay enough: expected {expected:?}, got {actual:?}")]
+#[error("transaction does not pay enough: expected {expected}, got {actual}")]
 pub struct InsufficientFunds {
     pub expected: Amount,
     pub actual: Amount,
 }
 
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+#[error("THe balance is too low, current balance: {balance}")]
```
rishflab, line 216:
```suggestion
#[error("The balance is too low, current balance: {balance}")]
```

da-kami:
> Monero harness test failed (https://github.com/comit-network/xmr-btc-swap/pull/232/checks?check_run_id=1975587854) - unfortunately activating debug did not add much value - still unclear what is the problem.

https://github.com/comit-network/xmr-btc-swap/pull/235

rishflab (commented):
What is bob's behaviour in the case Alice has insufficient xmr? It looks like we will timeout while waiting for that message without any indication of what went wrong? Can we send this error/rejection message to bob and display it on his side for UX

thomaseizinger:
> What is bob's behaviour in the case Alice has insufficient xmr? It looks like we will timeout while waiting for that message without any indication of what went wrong? Can we send this error/rejection message to bob and display it on his side for UX

Yeah we should at least properly react to not getting a response. Not sure if we directly want to display the error.

I would start with a "sorry, failed to get a quote, please retry later" message.

da-kami:
> > What is bob's behaviour in the case Alice has insufficient xmr? It looks like we will timeout while waiting for that message without any indication of what went wrong? Can we send this error/rejection message to bob and display it on his side for UX
> 
> Yeah we should at least properly react to not getting a response. Not sure if we directly want to display the error.
> 
> I would start with a "sorry, failed to get a quote, please retry later" message.

I agree that we can add sending an Error later. We could close the connection from Alice's side upon error though - that would allow Bob to fail faster.

thomaseizinger (approved):
I think we can merge this without touching Bob within this PR. Bob doesn't do any error handling currently despite it being possible that things fail. As such, the situation gets strictly better with this PR by having Alice protected from draining.

da-kami:
> > > What is bob's behaviour in the case Alice has insufficient xmr? It looks like we will timeout while waiting for that message without any indication of what went wrong? Can we send this error/rejection message to bob and display it on his side for UX
> > 
> > 
> > Yeah we should at least properly react to not getting a response. Not sure if we directly want to display the error.
> > I would start with a "sorry, failed to get a quote, please retry later" message.
> 
> I agree that we can add sending an Error later. We could close the connection from Alice's side upon error though - that would allow Bob to fail faster.

The current behavior already defaults to closing the connection on Alice's side because Alice drops the channel upon error which results in a closed connection on Bob's side. For error handling on Bob's side please refer to https://github.com/comit-network/xmr-btc-swap/pull/238

PR #233, opened 2021-02-25
ASB max sell amount

da-kami:


rishflab:
Shouldn't max sell amount NOT have a default?  This is something that should be explicitly specified or the program should abort.

rishflab (changes requested):
Shouldn't max sell amount NOT have a default? This is something that should be explicitly specified or the program should abort.

da-kami:
> Shouldn't max sell amount NOT have a default? This is something that should be explicitly specified or the program should abort.

The default was added because at the moment the asb is just a demo application and should always have a limit. It just makes it easier to run the asb without having to alter the setup on the raspberry - and the default does not do any harm.

da-kami:
@thomaseizinger what's your opinion on this default?

thomaseizinger (approved):
I am fine with a default value as it makes it easier to run for us.

PR #235, opened 2021-02-25
monero-rpc debug log for moner-harness tests 

da-kami:
This allows us to see the response from the monerod client on debug.
When the monero-harness was split up into harness and rpc this was overlooked.
We need the debug logs to investigate the monero harness CI fail bug.

### DIFF
--- a/monero-harness/tests/testutils/mod.rs
+++ b/monero-harness/tests/testutils/mod.rs
@@ -16,12 +16,13 @@ pub fn init_tracing() -> DefaultGuard {
     let global_filter = tracing::Level::WARN;
     let test_filter = tracing::Level::DEBUG;
     let monero_harness_filter = tracing::Level::DEBUG;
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
I guess you could add a `refresh_until` function to the wallet but for that to be reusable, you might also need to customize the refresh internal, number of tries before bail, etc.

At that stage, it doesn't provide much value and every user might as well write their own loop.

Keep it isolated to the tests if it increases our CI resilience.

da-kami:
Unfortunately what I though is the problem is actually not the problem - the test is still failing, will investigate again.

da-kami:
The problem is actually that the monerod client returns an error when generating blocks. We overlooked updating the init tracing for the monero harness test when introducing the `monero-rpc` - I added that filter to the logs now. I hope we run into the failure again to finally see the bug.

rishflab (approved):

PR #236, opened 2021-02-26
Some wallet cleanup + watch for deposit

thomaseizinger:


rishflab (approved):

swap/src/bitcoin/wallet.rs
```diff
@@ -111,19 +111,18 @@ impl Wallet {
         Ok(())
     }
 
-    pub async fn build_tx_lock_psbt(
+    pub async fn send_to_address(
```
da-kami, line 115: 🔧 I find this rename confusing, because `send_to_address` for me implies publishing the transaction to the network. However, this function just constructs a PSBT and has nothing to do with publication as far as I can see. How about `build_psbt`?

swap/src/bin/swap_cli.rs
```diff
@@ -107,22 +107,30 @@ async fn main() -> Result<()> {
             )
             .await?;
 
-            let swap_id = Uuid::new_v4();
+            // TODO: Also wait for more funds if balance < dust
+            if bitcoin_wallet.balance().await? == Amount::ZERO {
+                debug!(
+                    "Waiting for BTC at address {}",
```
da-kami, line 113: 💡 Eventually we could actually query the ASB for minimum and maximum amounts - in the end the CLI will only be able to swap if the ASB allows it. We configure min/max in XMR at the moment, but we could change that in the ASB. Just an idea, but I think that would add to user experience in the future, because the user would know the boundaries beforehand.

da-kami (approved):

swap/src/bitcoin/wallet.rs
```diff
@@ -38,10 +38,9 @@ enum Error {
 }
 
 pub struct Wallet {
-    pub inner: Arc<Mutex<bdk::Wallet<ElectrumBlockchain, bdk::sled::Tree>>>,
-    pub network: bitcoin::Network,
```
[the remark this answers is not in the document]
thomaseizinger: Yes, encapsulation is quite important! Thanks for giving it some thought!

swap/src/protocol/alice/state.rs
```diff
@@ -53,7 +53,7 @@ pub enum AliceState {
     },
     BtcPunishable {
         monero_wallet_restore_blockheight: BlockHeight,
-        tx_refund: TxRefund,
+        tx_refund: Box<TxRefund>,
```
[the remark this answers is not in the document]
thomaseizinger: Because clippy warned about differences in enum size!

swap/src/bitcoin/wallet.rs
thomaseizinger: I see, yes the name could use improvement to say it only builds a transaction and doesn't send it. Will give it some more thought!

PR #238, opened 2021-02-26
Bob error handling

da-kami:


swap/src/bin/swap_cli.rs
```diff
@@ -118,8 +118,16 @@ async fn main() -> Result<()> {
             );
             let (swap, event_loop) = bob_factory.with_init_params(send_bitcoin).build().await?;
 
-            tokio::spawn(async move { event_loop.run().await });
-            bob::run(swap).await?;
+            let handle = tokio::spawn(async move { event_loop.run().await });
+            let swap = bob::run(swap);
+            tokio::select! {
+                event_loop_result = handle => {
+                    event_loop_result??;
```
thomaseizinger, line 125: I wonder if Rust will ever add an exclamation mark operator: `??!!?!`

swap/src/protocol/bob/event_loop.rs
```diff
@@ -193,7 +193,7 @@ impl EventLoop {
                         }
                         OutEvent::ResponseSent => {}
                         OutEvent::CommunicationError(err) => {
-                            error!("Communication error: {:#}", err)
+                            anyhow!("Communication error: {:#}", err);
```
thomaseizinger, line 196: This doesn't actually terminate the loop, you are just constructing an error. You need to either explicitly say `return Err(anyhow!(""))` or use `bail!` which implies the `return`.

thomaseizinger (changes requested):

swap/src/protocol/bob/event_loop.rs
da-kami: Fixed - changed to `bail!`! For some reason I thought that `anyhow` is equivalent to `bail`, but as you pointed out it does not return. How would it ever return whit a questionmark in the end. What a noob mistake 🤦‍♂️

swap/src/bin/swap_cli.rs
da-kami: Eineinself!

thomaseizinger (approved):

PR #248, opened 2021-03-01
Wait for wallet to catch up instead of block generation

da-kami:
The monero harness wallet always starts a miner that mines new blocks every second.
This can conflict with additionally triggering block generation  and cause this error:

```
monero_rpc::rpc::monerod: generate blocks response: {
  "error": {
    "code": -7,
    "message": "Block not accepted"
  },
  "id": "1",
  "jsonrpc": "2.0"
}
```

See error in CI run here: https://github.com/comit-network/xmr-btc-swap/runs/2001131193

Since the miner is generating blocks anyway we can wait for the wallet to catch up.
Refresh is done upon querying the balance, thus the refresh calls were removed.

---

Note: we could also opt for starting the miner optionally. This would make this test more efficient, but is a more intrusive change, thus I opted for this fix for now.

### DIFF
--- a/monero-harness/tests/wallet.rs
+++ b/monero-harness/tests/wallet.rs
@@ -1,6 +1,7 @@
 use crate::testutils::init_tracing;
-use monero_harness::Monero;
+use monero_harness::{Monero, MoneroWalletRpc};
 use spectral::prelude::*;
+use std::{thread::sleep, time::Duration};
 use testcontainers::clients::Cli;
 
 mod testutils;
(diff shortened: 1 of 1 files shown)


monero-harness/tests/wallet.rs
```diff
@@ -66,3 +57,16 @@ async fn fund_transfer_and_check_tx_key() {
...
+async fn wait_for_wallet_to_catch_up(wallet: &MoneroWalletRpc, expected_balance: u64) {
+    let max_retry = 15;
+    let mut retry = 0;
+    loop {
+        retry += 1;
+        let balance = wallet.balance().await.unwrap();
+        if balance == expected_balance || max_retry == retry {
+            break;
+        }
+        sleep(Duration::from_secs(1));
```
thomaseizinger, line 70: This should be a `tokio::time::sleep`.

thomaseizinger (changes requested):
Thank you! Just one comment :)

rishflab (approved):

thomaseizinger (approved):

PR #259, opened 2021-03-02
Upgrade bitcoin wallet to use BIP84 derivation scheme

### LINKED ISSUE #258: Don't reuse addresses
Currently, we Bitcoin wallet is defined using a single private key.
This harms privacy due to address reuse. We should define the wallet using an extended private key that allows derivation.

rishflab:
Closes #258

### DIFF
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -7,9 +7,10 @@ use anyhow::{anyhow, bail, Context, Result};
 use backoff::{backoff::Constant as ConstantBackoff, future::retry};
 use bdk::{
     blockchain::{noop_progress, Blockchain, ElectrumBlockchain},
+    descriptor::Segwitv0,
     electrum_client::{self, Client, ElectrumApi},
-    miniscript::bitcoin::PrivateKey,
-    FeeRate,
+    keys::DerivableKey,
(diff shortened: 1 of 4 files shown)


swap/src/bitcoin/wallet.rs
```diff
@@ -56,7 +57,7 @@ impl Wallet {
         let db = bdk::sled::open(wallet_dir)?.open_tree(SLED_TREE_NAME)?;
 
         let bdk_wallet = bdk::Wallet::new(
-            bdk::template::P2WPKH(private_key),
+            bdk::template::BIP84(key, KeychainKind::External),
             None,
```
thomaseizinger, line 61: Hmm, I guess we don't need a change descriptor because we always swap everything... Thoughts?

thomaseizinger (approved):

swap/src/bitcoin/wallet.rs
rishflab: Yeh that was what I thought. In the case we somehow had change I assumed it would default to the standard descriptor to derive a change address, but I could not confirm this. I will explicitly specify a change address so we can be sure of the behaviour in case we end up having change somehow.

thomaseizinger (approved):

PR #261, opened 2021-03-02
Sweep xmr funds from generated temp wallet

### LINKED ISSUE #252: Transfer XMR to address after swap
* Remove buy-xmr command
* Move all the options of buy-xmr to the global scope
* Add obligatory parameter --xmr-address (try to use required if command is none of structopt to avoid having to specify it for other commands)

Note (Daniel): Once we do this we will have to make sure to wait the full 10 blocks on stagenet to be able to spend from the wallet created with generate_from_keys. At the moment this is set to 5 blocks which does not make much sense - it should either be 10 or 0 - if we spend it should be 10 if don't spend it can be 0, then the user has to make sure to wait the required 10 blocks to be able to spend.

da-kami:
Fixes #252

Please review by commit :)
Did a few cleanups before actually doing the feature.

Please note the comment that influenced this solution: https://github.com/comit-network/xmr-btc-swap/issues/252#issuecomment-789387074

### DIFF
--- a/monero-rpc/src/rpc/wallet.rs
+++ b/monero-rpc/src/rpc/wallet.rs
@@ -147,7 +147,7 @@ impl Client {
         Ok(())
     }
 
-    /// Opens a wallet using `filename`.
+    /// Close the currently opened wallet, after trying to save it.
     pub async fn close_wallet(&self) -> Result<()> {
         let request = Request::new("close_wallet", "");
 
--- a/monero-rpc/src/rpc/wallet.rs
+++ b/monero-rpc/src/rpc/wallet.rs
@@ -147,6 +147,28 @@ impl Client {
         Ok(())
     }
 
+    /// Opens a wallet using `filename`.
+    pub async fn close_wallet(&self) -> Result<()> {
+        let request = Request::new("close_wallet", "");
+
+        let response = self
+            .inner
+            .post(self.url.clone())
+            .json(&request)
+            .send()
+            .await?
+            .text()
+            .await?;
+
+        debug!("close wallet RPC response: {}", response);
+
+        if response.contains("error") {
+            bail!("Failed to close wallet")
+        }
+
+        Ok(())
+    }
+
     /// Creates a wallet using `filename`.
     pub async fn create_wallet(&self, filename: &str) -> Result<()> {
         let params = CreateWalletParams {
@@ -313,6 +335,28 @@ impl Client {
         let r: Response<Refreshed> = serde_json::from_str(&response)?;
         Ok(r.result)
     }
+
+    /// Transfers the complete balance of the account to `address`.
+    pub async fn sweep_all(&self, address: &str) -> Result<SweepAll> {
+        let params = SweepAllParams {
+            address: address.into(),
+        };
+        let request = Request::new("sweep_all", params);
+
+        let response = self
+            .inner
+            .post(self.url.clone())
+            .json(&request)
+            .send()
+            .await?
+            .text()
+            .await?;
+
+        debug!("sweep_all RPC response: {}", response);
+
+        let r: Response<SweepAll> = serde_json::from_str(&response)?;
+        Ok(r.result)
+    }
 }
 
 #[derive(Serialize, Debug, Clone)]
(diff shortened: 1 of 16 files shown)


swap/src/monero.rs
```diff
@@ -240,7 +240,7 @@ pub trait CreateWalletForOutputThenLoadDefaultWallet {
 
 #[async_trait]
 pub trait OpenWallet {
```
thomaseizinger, line 241: For Bitcoin, I got rid of most of these traits. Should we do the same thing for Monero?

swap/src/bin/asb.rs
```diff
@@ -167,14 +167,11 @@ async fn init_wallets(
     // Setup the Monero wallet
     let open_wallet_response = monero_wallet.open().await;
     if open_wallet_response.is_err() {
-        monero_wallet
-            .create_wallet(DEFAULT_WALLET_NAME)
-            .await
-            .context(format!(
-                "Unable to create Monero wallet.\
+        monero_wallet.create().await.context(format!(
```
thomaseizinger: Should we just combine this into `open_or_create`?

swap/src/monero.rs
```diff
@@ -229,8 +229,8 @@ pub trait CreateWalletForOutput {
 }
 
 #[async_trait]
-pub trait CreateWalletForOutputThenLoadDefaultWallet {
-    async fn create_and_load_wallet_for_output_then_load_default_wallet(
+pub trait CreateWalletForOutputThenReloadWallet {
```
thomaseizinger, line 231: I would just call this `create_from` and don't mention this reloading behaviour in the name. We simply guarantee to the caller that the current wallet remains loaded when you create a new one.

monero-rpc/src/rpc/wallet.rs
```diff
@@ -313,6 +335,28 @@ impl Client {
...
+            .post(self.url.clone())
+            .json(&request)
+            .send()
+            .await?
+            .text()
+            .await?;
+
+        debug!("sweep_all RPC response: {}", response);
+
+        let r: Response<SweepAll> = serde_json::from_str(&response)?;
```
thomaseizinger, line 357: I usually try to avoid type hints and use turbofish syntax instead because it is more refactoring friendly, i.e.:

```rust
serde_json::from_str::<Response<SweepAll>(&response)
```

This way, your are co-locating the type with the expression that requires it. This makes it easier to take this expression and compose it with other ones without having to touch the type hint on the variable.

For example, it can easily transform this code:

```rust
let r = serde_json::from_str::<Response<SweepAll>(&response);

Ok(r.result)
```

into this:


```rust
let result = serde_json::from_str::<Response<SweepAll>(&response).result;

Ok(result)
```

and the type inference of the rust compiler will do the rest.

Specifying type hints gets in the way of this, which is why I only place them as a last resort.

swap/src/monero/wallet.rs
```diff
@@ -61,6 +61,15 @@ impl Wallet {
         self.inner.lock().await.refresh().await
     }
 
+    pub async fn sweep_all(&self, address: Address) -> Result<Vec<TxHash>> {
+        self.inner
+            .lock()
+            .await
+            .sweep_all(address.to_string().as_str())
+            .await
+            .map(|sweep_all| sweep_all.tx_hash_list.into_iter().map(TxHash).collect())
```
thomaseizinger, line 70: Just `?` this instead of mapping on `Result`.

thomaseizinger:
```rust
let sweep_all = self.inner
            .lock()
            .await
            .sweep_all(address.to_string().as_str())
            .await?;

sweep_all.tx_hash_list.into_iter().map(TxHash).collect()
```

thomaseizinger (approved):

swap/src/monero.rs
da-kami: 👍 Will do as follow up

monero-rpc/src/rpc/wallet.rs
da-kami: 👍 I will add it to the follow-up cleaning that removes the traits (because it should be changed throughout the file)

swap/src/monero.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/266

monero-rpc/src/rpc/wallet.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/266

thomaseizinger (approved):

rishflab (approved):
https://github.com/comit-network/xmr-btc-swap/pull/261/commits/adc5c7c9817267b4fef946943bb9489b7de0a586 This commit could have a better name/description

Issue #255, opened 2021-03-02
Tell the user about the swap limits


thomaseizinger:
The ASB currently limits the amount we can swap to a specific XMR value.

We should tell the user a max amount of BTC that we can swap while we are waiting for their funds.

We should also display the rate to them and potentially a timeout for how long this rate is valid. To do that, we need more information from the ASB about the rate _before_ the swap even starts.

Suggestion:

1. Add a protocol to the ASB that allows to query the latest price + a limit (+ optionally a timeout).
2. When waiting for funds from the user on the CLI, display this information (+ optionally abort if we hit the timeout).
3. Swap everything the user transfers or the limit if they transfer more.

da-kami:
In my understanding we touch two different topics in this issue:

1. The maximum amount allowed by the ASB 
2. How long an offer is valid

I don't think we should invest time into 2. because it is a rabbit hole. 

Concerning 1. I think it would be easier to change the ASB's maximum amount acceptable to BTC. By doing so we take the problem of "what is allowed" away from the user, but then have the problem that the ASB has to decide if a trade of x BTC can be fulfilled in XMR.
Since we don't have an elaborate strategy for making sure we can fulfill trades (i.e. bookkeeping like in the old Nectar maker) I would not invest too much time in making this work perfectly. 

Thus, my suggestion:

* Change the ASB max amount acceptable to BTC; when user triggers swap:
    * check if BTC amount is below max (if not fail)
    * check if quote can be fulfilled by balance (if not fail)
* Hard-code the max amount in the CLI (in a second iteration we can have an endpoint on the ASB)

I think that is the faster way forward for now.

da-kami:
In my understanding we touch three different topics in this issue:

1. The maximum tradable amount and currency allowed by the ASB 
2. The possibility of getting a price (or quote) without triggering a swap
3. The time until an ASB offer is "valid"

I don't think we should invest time into 3. because it is a rabbit hole. 

I see the biggest problem at our execution_setup, that is tightly coupled to quote request/response.
This was done, because the ASB is dictating the price. 
 
I am not sure we need an entirely new protocol to achieve 1. and 2. - but we will have to separate the swap_execution trigger from the quote request/response.

I think the CLI should be able to do a quote request and just get a response without any execution setup.
Additionally the CLI should be able to trigger the execution setup, which **implies** the quote at the time of triggering the execution setup (i.e. the quote might be different). 

[...]

PR #264, opened 2021-03-03
Untangle Bob's `EventLoop` from the `Builder`

thomaseizinger:
This is work towards #255.

### DIFF
--- a/swap/src/network.rs
+++ b/swap/src/network.rs
@@ -2,10 +2,8 @@ pub mod peer_tracker;
 pub mod request_response;
 pub mod transport;
 
-use crate::seed::SEED_LENGTH;
-use bitcoin::hashes::{sha256, Hash, HashEngine};
 use futures::prelude::*;
-use libp2p::{core::Executor, identity::ed25519};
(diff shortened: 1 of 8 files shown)


swap/src/network.rs
```diff
@@ -19,35 +17,3 @@ impl Executor for TokioExecutor {
...
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub struct Seed([u8; SEED_LENGTH]);
-
-impl Seed {
-    /// prefix "NETWORK" to the provided seed and apply sha256
-    pub fn new(seed: crate::seed::Seed) -> Self {
-        let mut engine = sha256::HashEngine::default();
-
-        engine.input(&seed.bytes());
-        engine.input(b"NETWORK");
```
da-kami, line 32: 🔧 If we remove this, the default `PeerId` we set in the `swap_cli` will change.
I think we should do the change backwards compatible by adding the `NETWORK_` prefix when we derive the peer-id.

da-kami (approved):

rishflab (approved):

PR #267, opened 2021-03-03
Use gtar to create release archive on Mac

### LINKED ISSUE #251: MacOS release binary doesn't work
Trying to execute the binary results in an exec error.

rishflab:
Follow up commit uses 7zip to create a zip for windows and appends the binary with a .exe

Here are the releases produced by this PR on my fork:
https://github.com/rishflab/xmr-btc-swap/releases/tag/0.34.0
I have tested the above releases to work up to asking for a btc deposit on :
- [x] macOS
- [x] windows
- [ ] ubuntu

Fixes #251.


da-kami (approved):

.github/actions/create-release-archive/action.yml
```diff
@@ -35,4 +35,4 @@ runs:
 
     - name: Make archive
       shell: bash
-      run: tar -C ./target/${{ inputs.target }}/release --create --file=${{ steps.create-archive-name.outputs.archive }} ${{ inputs.binary }}
\ No newline at end of file
+      run: gtar -C ./target/${{ inputs.target }}/release --create --file=${{ steps.create-archive-name.outputs.archive }} ${{ inputs.binary }}
```
thomaseizinger, line 38: I assume you tested that this actually works on all platforms? i.e. `windows-latest` actually ships with a `gtar` binary?

thomaseizinger (approved):

thomaseizinger:
Instead of disabling windows again, can we just make the binary configurable for now and use tar on linux + windows and gtar on macos?

thomaseizinger:
Later, we can worry about making or using an action that does the right thing for each operating system with less configuration.

rishflab:
> Later, we can worry about making or using an action that does the right thing for each operating system with less configuration.

I got them all working now. We need gtar for Mac and tar for ubuntu at least so afaik we will have to create at least one seperate action.

.github/actions/create-release-archive/action.yml
rishflab: it doesn't : (. Had to create a new action

.github/actions/create-release-archive/macos/action.yml
```diff
@@ -0,0 +1,38 @@
+name: Create release archive
```
da-kami, line 1: Why are we duplicating the complete file for the different systems, can we not use an `if` on the operating system just for the command?

.github/actions/create-release-archive/windows/action.yml
```diff
@@ -0,0 +1,40 @@
+name: Create release archive
+description: Creates a tar archive for a release binary
```
da-kami, line 2: This one does create a `zip` archive ;)

da-kami (commented):
Why all that duplication? As far as I can see the only difference is the archive command - what am I missing?

thomaseizinger:
> Why all that duplication? As far as I can see the only difference is the archive command - what am I missing?

It is a bit tricky. I am also not a fan of the duplication but iterating on GitHub workflows is painful so if we have something here that works, I am keen to get it in just so we resolve the blocker of the MacOS binary not working. We always improve things on top of that.

We might be able to solve this elegantly using [Job outputs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs).

.github/actions/create-release-archive/windows/action.yml
thomaseizinger: Correct, that should be changed to avoid future confusion.

rishflab:
> Why all that duplication? As far as I can see the only difference is the archive command - what am I missing?

believe me I tried doing it without duplication. Eventually just settled on something that works.

.github/actions/create-release-archive/windows/action.yml
rishflab: fixed

.github/actions/create-release-archive/macos/action.yml
rishflab: I tried but couldn't get it to work. It might of been because I was using 7z incorrectly and the if statements were actually working fine. There is no meaningful error message and iterations take time. I think we should be able reduce duplication now that we have a working version.

thomaseizinger (approved):

thomaseizinger:
I am going to merge this. We can always iterate on top of it!

bors r+

da-kami (approved):
Let's merge it then. But let's please do the follow up to reduce duplication within this project. Next time someone has to touch these actions it might be pain with the duplication. I'd record a technical debt task to clean that up :)

PR #268, opened 2021-03-03
Introduce `quote` protocol and display it to the user before they fund

### LINKED ISSUE #255: Tell the user about the swap limits

thomaseizinger:
Fixes #255.

### DIFF
--- a/swap/src/bin/swap_cli.rs
+++ b/swap/src/bin/swap_cli.rs
@@ -16,6 +16,7 @@ use anyhow::{bail, Context, Result};
 use prettytable::{row, Table};
 use reqwest::Url;
 use std::cmp::min;
+use std::future::Future;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
@@ -25,6 +26,7 @@ use swap::cli::command::{Arguments, Command};
 use swap::cli::config::{read_config, Config};
 use swap::database::Database;
 use swap::execution_params::GetExecutionParams;
+use swap::network::quote::BidQuote;
 use swap::protocol::bob;
 use swap::protocol::bob::cancel::CancelError;
 use swap::protocol::bob::{Builder, EventLoop};
@@ -124,39 +126,22 @@ async fn main() -> Result<()> {
             )?;
             let handle = tokio::spawn(event_loop.run());
 
-            let bid_quote = event_loop_handle
-                .request_quote()
-                .await
-                .context("failed to request quote")?;
+            let send_bitcoin = determine_btc_to_swap(
+                event_loop_handle.request_quote(),
+                bitcoin_wallet.balance(),
+                bitcoin_wallet.new_address(),
+                async {
+                    while bitcoin_wallet.balance().await? == Amount::ZERO {
+                        bitcoin_wallet.sync_wallet().await?;
 
-            info!("Received quote: 1 XMR ~ {}", bid_quote.price);
+                        tokio::time::sleep(Duration::from_secs(1)).await;
+                    }
 
-            // TODO: Also wait for more funds if balance < dust
-            if bitcoin_wallet.balance().await? == Amount::ZERO {
-                info!(
-                    "Please deposit the BTC you want to swap to {} (max {})",
-                    bitcoin_wallet.new_address().await?,
-                    bid_quote.max_quantity
-                );
-
-                while bitcoin_wallet.balance().await? == Amount::ZERO {
-                    bitcoin_wallet.sync_wallet().await?;
-
-                    tokio::time::sleep(Duration::from_secs(1)).await;
-                }
-
-                debug!("Received {}", bitcoin_wallet.balance().await?);
-            } else {
-                info!(
-                    "Still got {} left in wallet, swapping ...",
-                    bitcoin_wallet.balance().await?
-                );
-            }
-
-            let max_giveable = bitcoin_wallet.max_giveable(TxLock::script_size()).await?;
-            let max_accepted = bid_quote.max_quantity;
-
-            let send_bitcoin = min(max_giveable, max_accepted);
+                    bitcoin_wallet.balance().await
+                },
+                bitcoin_wallet.max_giveable(TxLock::script_size()),
+            )
+            .await?;
 
             let swap = Builder::new(
                 db,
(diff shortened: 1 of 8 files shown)


swap/src/bin/swap_cli.rs
```diff
@@ -112,14 +112,29 @@ async fn main() -> Result<()> {
...
+
+            debug!("Requesting quote");
+
+            let bid_quote = event_loop_handle
+                .request_quote()
+                .await
+                .context("failed to request quote")?;
 
-            let swap_id = Uuid::new_v4();
+            info!("Received quote: 1 XMR = {}", bid_quote.price);
```
da-kami, line 130:
```suggestion
            info!("The current price is 1 XMR = {} (the price during the swap may differ)", bid_quote.price);
```

This message is correct, but we should probably deliver that information differently, because the user has to understand that this quote is not equivalent to the amount that will be exchanged during the swap. It is a quote at the given time, where the swap will use the price at the point where the swap is started.

swap/src/bin/swap_cli.rs
```diff
@@ -112,14 +112,29 @@ async fn main() -> Result<()> {
...
 
             // TODO: Also wait for more funds if balance < dust
             if bitcoin_wallet.balance().await? == Amount::ZERO {
                 info!(
-                    "Please deposit BTC to {}",
-                    bitcoin_wallet.new_address().await?
+                    "Please deposit BTC to {} (max {})",
```
da-kami, line 135:
```suggestion
                    "Please deposit BTC (max {}) to be swapped to XMR to the swap-cli at address {}",
```

I think we need a bit more context here, so the user understands what will happen. It might also help to state that this BTC address is the swap-tool address. (Of course it is implied, but might not be bad to state it.)
Note that this would depend on the name of the application then - not super happy about that, but I think it helps the understanding.

swap/src/bin/swap_cli.rs
```diff
@@ -136,19 +151,14 @@ async fn main() -> Result<()> {
...
+            let max_accepted = bid_quote.max_quantity;
 
-            let (event_loop, event_loop_handle) = EventLoop::new(
-                &seed.derive_libp2p_identity(),
-                alice_peer_id,
-                alice_addr,
-                bitcoin_wallet.clone(),
-            )?;
-            let handle = tokio::spawn(event_loop.run());
+            let send_bitcoin = min(max_giveable, max_accepted);
```
da-kami, line 157: We default to the `max_accepted` in case the user transfers more - I like that, because this solves the issue with some faucets not allowing to enter an amount!
I assume the behavior when starting it after a swap would be that it immediately triggers another swap with the remaining balance?.

da-kami (approved):

swap/src/bin/swap_cli.rs
thomaseizinger: Yes, that would be the behaviour.

swap/src/bin/swap_cli.rs
thomaseizinger: I don't think we need to explicitly state that, where this address comes from is an implementation detail IMO. People can look into the code and check whether it is actually an atomic swap :)

swap/src/bin/swap_cli.rs
thomaseizinger: Yep, will update the message!

thomaseizinger:
@da-kami what do you think of 8085fe8?

da-kami:
> @da-kami what do you think of [8085fe8](https://github.com/comit-network/xmr-btc-swap/commit/8085fe8b5de408ad80ad47436df5ab60b1d4f34f)?

Fine with me - let's wait for user feedback and adapt if necessary :)

One more comment on the logs (I am currently running it):

```
2021-03-04 11:57:30   INFO Received quote: 1 XMR = 0.00434100 BTC
2021-03-04 11:57:30   INFO Still got 0.01519511 BTC left in wallet, swapping ...
2021-03-04 11:57:30   INFO Spot price for 0.00500000 BTC is 1.151808339092 XMR
```

I think the message `Still got 0.01519511 BTC left in wallet, swapping ...` should be adapted in case we have more than the max allowed - this would read better:

```
2021-03-04 11:57:30   INFO Received quote: 1 XMR = 0.00434100 BTC
2021-03-04 11:57:30   INFO Still got 0.01519511 BTC left in wallet, swapping maximum of 0.005 ...
2021-03-04 11:57:30   INFO Spot price for 0.00500000 BTC is 1.151808339092 XMR
```

(Upon restart the user does not see the maximum if there is still funds left, so the user might be confused about the amount being swapped.)

thomaseizinger:
@da-kami Checkout [`ea9745e` (#268)](https://github.com/comit-network/xmr-btc-swap/pull/268/commits/ea9745e42c45bf414cf341332fda890c0fe6bd25). I've added a couple of tests that also allow us to quickly check what the output is like without executing the full program every time :)

swap/src/bin/swap_cli.rs
```diff
@@ -320,3 +305,133 @@ async fn init_monero_wallet(
...
+    } else {
+        info!("Found {} in wallet", initial_balance);
+    }
+
+    let max_giveable = max_giveable.await?;
+    let max_accepted = bid_quote.max_quantity;
+
+    if max_giveable > max_accepted {
+        info!(
+            "Max giveable amount {} exceeds max accepted amount {}!",
```
da-kami, line 344: I think it should still say that it will default to the max accepted for the swap!

da-kami (approved):

swap/src/bin/swap_cli.rs
thomaseizinger: Well, the very next log message will be the spot price one where the user will see how much is being swapped. I also don't want to overload them with information ...

da-kami: Fair enough - let's just wait for feedback and then change things :)

PR #269, opened 2021-03-03
Monero address validity check

da-kami:


swap/src/bin/swap_cli.rs
```diff
@@ -107,6 +107,14 @@ async fn main() -> Result<()> {
             alice_peer_id,
             alice_addr,
         } => {
+            if receive_monero_address.network != monero_network {
+                bail!(
+                    "Given monero address is on network {:?}, expected address on network {:?}",
```
thomaseizinger, line 112: Can we use Display instead?

swap/src/cli/command.rs
```diff
@@ -72,6 +73,12 @@ pub enum Command {
     },
 }
 
+fn parse_monero_address(s: &str) -> Result<monero::Address, anyhow::Error> {
```
thomaseizinger, line 76: Could use anyhow::Result import.

swap/src/cli/command.rs
```diff
@@ -72,6 +73,12 @@ pub enum Command {
...
+    monero::Address::from_str(s)
+        .map_err(|err| anyhow!(err))
+        .context("Unable to parse the given monero address, make sure it is a valid address")
```
thomaseizinger, line 78: Using `with_context` you could say something like:

```
failed to parse "foobar" as a monero address
```

to give the user a clearer error message

thomaseizinger (approved):

swap/src/bin/swap_cli.rs
da-kami: The `monero::Network` does not implement `Display` - but it is an `enum`, so it looks like this:

```
Error: Given monero address is on network Mainnet, expected address on network Stagenet
```


I did not see the necessity to add a custom `Display` impl, given that it loops good anyway. 
Should we upstream a display impl/derive ...?

thomaseizinger: If that is what the `Debug` print looks like then it is fine :)

swap/src/cli/command.rs
```diff
@@ -72,6 +73,13 @@ pub enum Command {
     },
 }
 
+fn parse_monero_address(s: &str) -> Result<monero::Address> {
+    monero::Address::from_str(s).context(format!(
```
thomaseizinger, line 77: You will want to use `with_context` as soon as you use `format!` otherwise you are incurring an unnecessary allocation in case the parsing succeeds and the error is not actually needed.

thomaseizinger (approved):

swap/src/bin/swap_cli.rs
da-kami: > If that is what the `Debug` print looks like then it is fine :)

It is - I copied it after running it with a mainnet address :)

PR #273, opened 2021-03-04
Make buy-xmr the default behaviour

### LINKED ISSUE #234: Make `buy-xmr` the default behaviour when starting the CLI
We should probably also rename the tool to buy-xmr

rishflab:
Closes #234


swap/src/cli/command.rs
```diff
@@ -50,25 +148,53 @@ pub enum Command {
...
+
+        #[structopt(
+            long = "config",
+            help = "Provide a custom path to the configuration file. The configuration file must be a toml file.",
+            parse(from_os_str)
+        )]
+        config: Option<PathBuf>,
+
+        #[structopt(long, help = "Activate debug logging.")]
+        debug: bool,
```
thomaseizinger, line 163: Seems a bit excessive to duplicate these everywhere. Can we use [`flatten`](https://docs.rs/structopt/0.3.21/structopt/index.html#flattening) to reduce this duplication?

thomaseizinger (commented):
Interesting! Thanks for looking into this!

Had an idea that could reduce the duplication a bit. Let me know what you think.

swap/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 description = "XMR/BTC trustless atomic swaps."
 
 [[bin]]
-name = "swap_cli"
+name = "buy_xmr"
```
thomaseizinger, line 9: I think this section is actually redundant? We are not specifying the asb here and things still work so I think you can just delete that.

swap/src/cli/command.rs
da-kami: Agree with @thomaseizinger  and can confirm it is possible, we have done this for the configuration and other parameters before :)

rishflab: done

swap/Cargo.toml
rishflab: yeh it is. I have also forgotten to change all the CI mentions of swap_cli

thomaseizinger (dismissed):
Nice! Good that we got some tests in :)

I think you will like this @da-kami !

da-kami (dismissed):

da-kami:
Sorry to be the party pooper for putting my veto in for this change. I think the change from `swap buy-xmr --receive-address` to `buy_xmr --receive-address` does not justify the changes that make the command structure more complicated. 
Additionally, I think we don't gain much by renaming the `swap` binary and I think the binary should actually be called `swap` and not `buy_xmr` because eventually we want to add selling. Yes, at the moment it is not there yet, but I still think the binary should not be named `buy_xmr`. I think this requires further discussion :)

thomaseizinger:
@rishflab Thanks for putting this effort in! I actually agree with @da-kami in that `swap` works better as the binary name and makes it easier to extend once we add selling.

That being said, I think we made some great progress in this PR in regards to testing. Would you mind porting that to current master?

thomaseizinger:
Let's close this for now. The rename to `swap` already happened and with the outlook of eventually introducing `sell-xmr`, we should not be changing the binary name.

@rishflab It would still be good to get tests in for the commandline parsing if you get around to it.

PR #276, opened 2021-03-04
Use Rust cache action instead of rolling it ourselves

thomaseizinger:


.github/workflows/release-cli.yml
```diff
@@ -27,6 +27,8 @@ jobs:
           ref: ${{ github.event.release.target_commitish }}
           token: ${{ secrets.BOTTY_GITHUB_TOKEN }}
 
+      - uses: Swatinem/rust-cache@v1.2.0
```
thomaseizinger, line 30: Esp. using the cache action here should significantly speed up the release job and we should have our release binaries a lot earlier :)

PR #304, opened 2021-03-11
Try to renew the connection upon transient error

### LINKED ISSUE #303: Kraken websocket stream closes connection unexpectedly
This causes the rate-module to report a permanent error, which results in quote failing quote requests from the CLI.

Error log from the ASB:

```
Mar 10 14:57:52 raspberrypi cargo[5639]: Mar 10 14:57:52.136 DEBUG swap{id=8108a02b-377e-48f6-a4d6-90880887e94c}: swap::bitcoin::wallet: confirmations: BlockHeight(1) txid=eca50dfa7e53f1d65c759136ea67c0895a05afe9778a023cc344ffe705a7e5f4
Mar 10 14:57:52 raspberrypi cargo[5639]: Mar 10 14:57:52.138  INFO swap{id=8108a02b-377e-48f6-a4d6-90880887e94c}: swap::protocol::alice::swap: Current state: btc is redeemed
Mar 10 15:33:39 raspberrypi cargo[5639]: Mar 10 15:33:39.326  WARN swap::kraken: Rate updates incurred an unrecoverable error: Failed to read message from websocket stream: IO error: Connection reset by peer (os error 104)
Mar 10 23:45:56 raspberrypi cargo[5639]: Mar 10 23:45:56.680 DEBUG swap::protocol::alice::event_loop: Connection Established with 12D3KooWRvzPVmsH6Y7hLaMFjQU39LZqtWAhd8zNQDsz3TqNfev9
Mar 10 23:45:56 raspberrypi cargo[5639]: Mar 10 23:45:56.685  WARN swap::protocol::alice::event_loop: failed to make quote: Failed to get latest rate: Permanently failed to retrieve rate from Kraken peer=12D3KooWRvzPVmsH6Y7hLaMFjQU39LZqtWAhd8zNQDsz3TqNfev9
Mar 10 23:45:56 raspberrypi cargo[5639]: Mar 10 23:45:56.686 ERROR swap::protocol::alice::event_loop: Communication error: quote protocol with peer 12D3KooWRvzPVmsH6Y7hLaMFjQU39LZqtWAhd8zNQDsz3TqNfev9 failed due to ResponseOmission
```

The Kraken documentation states, that unexpected closing of connections can happen due to the same IP address being used over longer period of time which might cause Cloudflare to close the connection: 

https://support.kraken.com/hc/en-us/articles/360044504011-WebSocket-API-unexpected-disconnections-from-market-data-feeds

da-kami:
@thomaseizinger this is an attempt to fix #303 

I don't think this solution would work yet but I am not exactly sure how to rearrange the code to make it work.
I thought I put a Draft PR up :)

Currently `clippy` complains about:

```
warning: value assigned to `stream` is never read
  --> swap/src/kraken.rs:46:41
   |
46 | ...                   stream = connection::new().await?;
   |                       ^^^^^^
   |
   = note: `#[warn(unused_assignments)]` on by default
   = help: maybe it is overwritten before being read?
```

### DIFF
--- a/swap/src/kraken.rs
+++ b/swap/src/kraken.rs
@@ -22,17 +22,44 @@ pub fn connect() -> Result<RateUpdateStream> {
                 async move {
                     let mut stream = connection::new().await?;
 
-                    while let Some(update) = stream.try_next().await.map_err(to_backoff)? {
-                        let send_result = rate_update.send(Ok(update));
+                    loop {
+                        let update = stream.try_next().await;
 
-                        if send_result.is_err() {
-                            return Err(backoff::Error::Permanent(anyhow!(
-                                "receiver disconnected"
-                            )));
+                        match update {
+                            Ok(rate) => {
+                                if let Some(rate) = rate {
+                                    let send_result = rate_update.send(Ok(rate));
+
+                                    if send_result.is_err() {
+                                        return Err(backoff::Error::Permanent(anyhow!(
+                                        "receiver disconnected"
+                                    )));
+                                    }
+                                }
+                            }
+                            Err(e) => {
+                                match e {
+                                    // Connection closures and websocket errors will be retried
+                                    connection::Error::ConnectionClosed => {
+                                        // Try to renew the connection in case of websocket failure
+                                        // This failure can be caused by Cloudflare closing he connection for specific IPs
+                                        stream = connection::new().await?;
+                                        return Err(backoff::Error::Transient(anyhow::Error::from(e)));
+                                    },
+                                    connection::Error::WebSocket(_) => {
+                                        // Try to renew the connection in case of websocket failure
+                                        // This failure can be caused by Cloudflare closing he connection for specific IPs
+                                        stream = connection::new().await?;
+                                        return Err(backoff::Error::Transient(anyhow::Error::from(e)));
+                                    },
+
+                                    // Failures while parsing a message are permanent because they most likely present a
+                                    // programmer error
+                                    connection::Error::Parse(_) => { return Err(backoff::Error::Permanent(anyhow::Error::from(e))) },
+                                }
+                            }
                         }
                     }
-
-                    Err(backoff::Error::Transient(anyhow!("stream ended")))
                 }
             },
             |error, next: Duration| {
@@ -83,22 +110,6 @@ pub enum Error {
 }
 
 type RateUpdate = Result<Rate, Error>;
-
-/// Maps a [`connection::Error`] to a backoff error, effectively defining our
-/// retry strategy.
-fn to_backoff(e: connection::Error) -> backoff::Error<anyhow::Error> {
-    use backoff::Error::*;
-
-    match e {
-        // Connection closures and websocket errors will be retried
-        connection::Error::ConnectionClosed => Transient(anyhow::Error::from(e)),
-        connection::Error::WebSocket(_) => Transient(anyhow::Error::from(e)),
-
-        // Failures while parsing a message are permanent because they most likely present a
-        // programmer error
-        connection::Error::Parse(_) => Permanent(anyhow::Error::from(e)),
-    }
-}
 
 /// Kraken websocket connection module.
 ///


swap/src/kraken.rs
```diff
@@ -22,17 +22,44 @@ pub fn connect() -> Result<RateUpdateStream> {
...
+                                        "receiver disconnected"
+                                    )));
+                                    }
+                                }
+                            }
+                            Err(e) => {
+                                match e {
+                                    // Connection closures and websocket errors will be retried
+                                    connection::Error::ConnectionClosed => {
+                                        // Try to renew the connection in case of websocket failure
```
da-kami, line 44:
```suggestion
                                        // Try to renew the connection in case of connection failure
```

thomaseizinger:
> I don't think this solution would work yet but I am not exactly sure how to rearrange the code to make it work.
> I thought I put a Draft PR up :)
> 
> Currently `clippy` complains about:

Well clippy is right. You are over-writing the local `stream` variable and then returning from the closure before ever using `stream` again. That doesn't do anything.

Generally, I don't see a functional change in this PR. Unless I am mistaken, you just replaced the `map_err` with a manual match statement but the code path is still the same I think.

I am wondering why the current code doesn't work because `connection::Error::Websocket` errors should always be mapped to `Transient`. Is there maybe a codepath in `backoff` where it completely gives up even with `Transient` errors?

I am also wondering why we never see the `info` log about retrying.

thomaseizinger:
> I am wondering why the current code doesn't work because `connection::Error::Websocket` errors should always be mapped to `Transient`. Is there maybe a codepath in `backoff` where it completely gives up even with `Transient` errors?

Reading `backoff`'s source code, I think I found the problem and posted up a fix here: https://github.com/comit-network/xmr-btc-swap/pull/305

da-kami:
> > I don't think this solution would work yet but I am not exactly sure how to rearrange the code to make it work.
> > I thought I put a Draft PR up :)
> > Currently `clippy` complains about:
> 
> Well clippy is right. You are over-writing the local `stream` variable and then returning from the closure before ever using `stream` again. That doesn't do anything.
> 
> Generally, I don't see a functional change in this PR. Unless I am mistaken, you just replaced the `map_err` with a manual match statement but the code path is still the same I think.
> 
> I am wondering why the current code doesn't work because `connection::Error::Websocket` errors should always be mapped to `Transient`. Is there maybe a codepath in `backoff` where it completely gives up even with `Transient` errors?
> 
> I am also wondering why we never see the `info` log about retrying.

😬 Yeah, my solution does not really do much.
Still wondering if the current code makes sense without trying to re-connect. But maybe I just lack the understanding of websocket connections there.

We currently map `Connection` and `WebSocket` errors to be `Transient`:

https://github.com/comit-network/xmr-btc-swap/blob/9ad2160c69fd170bed1eae21b1b93881c4af6272/swap/src/kraken.rs#L87-L101

But we never re-establish a connection upon those transient error, but use the stream as it was before:

https://github.com/comit-network/xmr-btc-swap/blob/9ad2160c69fd170bed1eae21b1b93881c4af6272/swap/src/kraken.rs#L22-L36

Wouldn't it make sense to re-connect on the stream if we get a `Connection` or `WebWocket` error or is that just not how websocket connections work?

thomaseizinger:
> But we never re-establish a connection upon those transient error, but use the stream as it was before:

That is not true.

All of the above code is within the `retry` closure. The very first thing that happens is establishing a connection. If we fail, we `?` which bails out of the closure with an `Err` that is either `backoff::Transient` or `backoff::Permanent` and defines whether `retry` will call the closure again.

da-kami:
> > But we never re-establish a connection upon those transient error, but use the stream as it was before:
> 
> That is not true.
> 
> All of the above code is within the `retry` closure. The very first thing that happens is establishing a connection. If we fail, we `?` which bails out of the closure with an `Err` that is either `backoff::Transient` or `backoff::Permanent` and defines whether `retry` will call the closure again.

Yeah, you're right. The `backoff` block actually wraps the complete `async` block... Forget what I said 😅

Issue #202, opened 2021-02-16
Polling for changes to balance / transactions does not scale


da-kami:
## Problem

At the moment, when watching for transactions, we bombard the Bitcoin/Monero setup by checking every second. This is currently hardcoded.

## Goal

A scalable solution for waiting until a certain event (transaction included, balance changed, etc)

thomaseizinger:
Replaced by https://github.com/comit-network/xmr-btc-swap/issues/295.

PR #306, opened 2021-03-11
Fix logging and retrying of Monero transaction watching

thomaseizinger:
Hopefully, this should also reduce the load because I am not asking the node every second.

Related: https://github.com/comit-network/xmr-btc-swap/issues/202


da-kami (approved):
Thank you! I love the tests :)
I like the `while` loop representing the main condition, that really adds to understandability.

I'm fine with not using `backoff` for the RPC call here as described in comment:

```
// treating every error as transient and retrying
 // is obviously wrong but the jsonrpc client is
// too primitive to differentiate between all the
// cases
```

Note: I added a patch on top of your work, that does not pass the full execution parameters to the monero wallet. I think that the wallet should only know what it cares about. Feel free to squash / drop it :)

thomaseizinger:
It seems you fixed the clippy problems in the follow-up commit. I'll quickly fix that up!

da-kami:
> It seems you fixed the clippy problems in the follow-up commit. I'll quickly fix that up!

Naa, I amended that first :)

PR #307, opened 2021-03-11
Reduce load on electrum

rishflab:
.

PR #310, opened 2021-03-12
Filter script history blockheight to ensure latest tx status info

rishflab:
Not 100% if this will work but should get rid of a potential warning/bug and ensure we get the latest tx status info


thomaseizinger:
I don't think we need this, there should never be more than a single TX within the history entries.

rishflab:
Included in #307

PR #311, opened 2021-03-12
Add asb to release including ARM build

da-kami:
https://github.com/da-kami/xmr-btc-swap/releases/tag/0.3-test

The asb ARM build does neither run tests on CI not smoke test for the binary upon release because the there is no standard runner for ARM. One can configure a self-hosted one but I think that might be an overkill: https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners

I did download the produced ARM tar, unpack and test on the raspi, it works :)

Concerning openssl depdency see:

https://github.com/da-kami/xmr-btc-swap/runs/2093523365?check_suite_focus=true
https://www.reddit.com/r/rust/comments/axaq9b/opensslsys_error_when_crosscompiling_for/


rust-toolchain
```diff
@@ -1,3 +1,4 @@
 [toolchain]
 channel = "nightly-2021-01-31"
 components = ["rustfmt", "clippy"]
+targets = [ "x86_64-unknown-linux-gnu", "armv7-unknown-linux-gnueabihf", "x86_64-apple-darwin", "x86_64-pc-windows-msvc"]
```
thomaseizinger, line 4: Do we really need all targets here? That is gonna install all of them on all machines no?

.github/workflows/release-cli.yml
```diff
@@ -29,11 +48,19 @@ jobs:
...
+          sudo apt-get install gcc-arm-linux-gnueabihf
+
+      - name: Build ${{ matrix.target }} ${{ matrix.bin }} release binary
+        env:
+          CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER: arm-linux-gnueabihf-gcc
+        run: cargo build --target=${{ matrix.target }} --release --package swap --bin ${{ matrix.bin }}
 
       - name: Smoke test the binary
-        run: target/${{ matrix.target }}/release/swap --help
+        run: target/${{ matrix.target }}/release/${{ matrix.bin }} --help
```
thomaseizinger, line 63: I thought you didn't want to smoke test the arm one?

rust-toolchain
da-kami: I remember we discussed this in the context of the error "target not installed" on CI and you mentioned that I could solve that by defining the targets in the toolchain file. Is there a better way?

thomaseizinger: I think you only need to add the non-obvious target `armv7-unknown-linux-gnueabihf`. All others are installed by default if the current OS requires it. In other words, on Linux, `x86_64-unknown-linux-gnu` will be installed anyway because that is the current host OS and is what the binary will be built against.

However, if you add all of them here, rustup is going to install the apple and windows targets on a Linux machine as well even though we are never going to cross-compile to them and it is not even that easy I believe.

In summary, I think the only the thing you need to add here is the `armv7-unknown-linux-gnueabihf` target and you should be good to go to cross-compile to it from the `ubuntu-latest` runner in CI.

.github/workflows/release-cli.yml
thomaseizinger: Need an `if` that skips this step on the `armv7-unknown-linux-gnueabihf` target then.

rust-toolchain
da-kami: I see, thanks for the clarification!

.github/workflows/release-cli.yml
da-kami: I wonder how that slipped out... it only works with the if, thanks for spotting that one, should be fixed now!

.github/workflows/ci.yml
```diff
@@ -77,7 +85,7 @@ jobs:
       RUST_TEST_TASKS: 2
     strategy:
       matrix:
-        target: [ x86_64-unknown-linux-gnu, x86_64-apple-darwin ]
+        target: [ x86_64-unknown-linux-gnu, x86_64-apple-darwin, armv7-unknown-linux-gnueabihf ]
```
thomaseizinger, line 88: Is there any benefit of running the tests on `armv7`?

swap/Cargo.toml
```diff
@@ -31,6 +31,7 @@ libp2p-async-await = { git = "https://github.com/comit-network/rust-libp2p-async
 miniscript = { version = "5", features = ["serde"] }
 monero = { version = "0.10", features = ["serde_support"] }
 monero-rpc = { path = "../monero-rpc" }
+openssl = { version = "0.10", features = ["vendored"] }
```
thomaseizinger, line 34: Explicitly adding a transitive dependency might cause trouble down the line when we want to update them because we need to match the versions.

Which dependency brings in openssl? You can use `cargo tree` to figure that out.
Can we either toggle vendoring through a feature flag on that dependency or switch to a rust-based TLS implementation to avoid this?

.github/workflows/ci.yml
```diff
@@ -56,7 +56,15 @@ jobs:
...
 
+      - name: Install compiler for armhf arch
+        if: matrix.target == 'armv7-unknown-linux-gnueabihf'
+        run: |
+          sudo apt-get update
+          sudo apt-get install gcc-arm-linux-gnueabihf
+
       - name: Build binary
+        env:
+          CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER: arm-linux-gnueabihf-gcc
```
thomaseizinger, line 67: We could replace this by adding a `.cargo/config.toml` file to the repository and specifying the linker there: https://doc.rust-lang.org/cargo/reference/config.html#targettriplelinker

I think that is preferable because we don't specify it in the GitHub-specific workflow then which allows us to cross-compile easier on local machines.

thomaseizinger (changes requested):
Sorry for the two pass review but I think there are a number of things we should still tackle:

1. Use a cargo config file instead of the env variable
2. Don't run the tests on armv7 (one linux target is enough I'd say)
3. Don't depend on `openssl` directly if possible

Additionally, there is one more thing we need to keep in mind!
The code for downloading monero-wallet-rpc is only going to work on x64 architecture. We should add a better `cfg` guard there to make sure we don't compile `swap` on platforms that it is not gonna work on. This will cause additional changes to the CI script (because we need to skip building swap on `armv7`) so I am fine with delaying it but we should definitely open an issue about it!

thomaseizinger:
Also, the build jobs no longer seem to run for some reason?

rishflab (approved):

da-kami:
Thanks for the input @thomaseizinger - should be all fixed up now, please have another look :)

thomaseizinger (approved):
Nice work, thanks for incorporating the changes!

thomaseizinger:
Nice, should work now I believe!

PR #312, opened 2021-03-14
UI

da-kami:
Did a weekend project just to see if a UI can be achieved with druid. I though I share this so you can have a look. I don't intend to merge this any time soon.

Even though druid it still quite immature a simple UI is possible and does not take too much effort. 

This is a proof of concept far from being complete - the most immediate issues are:

* There is no handling of the BTC funds yet (no wallet features, no waiting for funds) - at the moment the UI expects that there are `0.001` available, otherwise just errors.
* I was able to run a swap, but eventually the program crashed with:

```
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
```

I feel, with showcasing the UI we can see that for driving the swap the generator approach would be more suitable than the recursive function. The generator would give us far more control over driving the swap in fine granular steps.

I did not invest any time into look and feel - so far it is extremely rudimentary just to proof the point that it can be done...

image

### DIFF
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -31,6 +31,7 @@ libp2p-async-await = { git = "https://github.com/comit-network/rust-libp2p-async
 miniscript = { version = "5", features = ["serde"] }
 monero = { version = "0.10", features = ["serde_support"] }
 monero-rpc = { path = "../monero-rpc" }
+openssl = { version = "0.10", features = ["vendored"] }
 pem = "0.8"
 prettytable-rs = "0.8"
 rand = "0.7"
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -24,6 +24,7 @@ conquer-once = "0.3"
 curve25519-dalek = "3"
 dialoguer = "0.7"
 directories-next = "2"
+druid = { version = "0.7", features = ["image", "svg"] }
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", features = ["libsecp_compat", "serde"] }
 futures = { version = "0.3", default-features = false }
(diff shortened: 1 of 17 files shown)


swap/src/protocol/bob/swap.rs
```diff
@@ -404,16 +464,24 @@ async fn run_until_internal(
...
-) -> Result<bob::state::State2> {
+) -> Result<monero::Amount> {
     let xmr = event_loop_handle.request_spot_price(btc).await?;
 
     tracing::info!("Spot price for {} is {}", btc, xmr);
 
+    Ok(xmr)
+}
+
+pub async fn execution_setup(
```
thomaseizinger, line 478: Happy to merge clean-up like this into the main branch if you split it out into a dedicated commit!

swap/Cargo.toml
```diff
@@ -24,6 +24,7 @@ conquer-once = "0.3"
 curve25519-dalek = "3"
 dialoguer = "0.7"
 directories-next = "2"
+druid = { version = "0.7", features = ["image", "svg"] }
```
thomaseizinger, line 27: I think this should be a separate crate or at least feature flagged!

da-kami:
I had a few more thoughts on this and had a look at flutter. I would not go forward with `druid` because it is just too immature and we just don't have time to contribute much. 
Once I find time I would invest into @thomaseizinger 's idea of emitting the tracing events as JSON so they can be processed - that way we can create a UI in clutter, that spins up the CLI and then processes the tracing logs. We should invest a bit more time into thinking about stability (what happens if we change/remove log output...) - would be great if we could have a chat about this @thomaseizinger :)

PR #314, opened 2021-03-15
Remove CLI config file in favour of parameters

### LINKED ISSUE #282: Consider removing the CLI config file in favor of command-line arguments

da-kami:
Fixes #282 

The CLI has sensible default values for all parameters,
thus a config file is not really an advantage but just
keeps getting in our way, so re remove it.

Trait impls on `Data` needed for structopt, see https://docs.rs/structopt/0.3.21/structopt/#default-values


swap/src/cli/command.rs
```diff
@@ -111,6 +124,64 @@ pub struct MoneroParams {
...
+
+    #[structopt(long = "electrum-rpc",
+    help = "Provide the Bitcoin Electrum RPC URL",
+    default_value = DEFAULT_ELECTRUM_RPC_URL
+    )]
+    pub electrum_rpc_url: Url,
+}
+
+#[derive(structopt::StructOpt, Debug)]
+pub struct DataParams {
```
thomaseizinger, line 143: You are only using `DataParams` once, can we not just inline that and get rid of the `FromStr` impls etc?

thomaseizinger (approved):

swap/src/cli/command.rs
rishflab: yep agreed

rishflab (approved):

swap/src/cli/command.rs
da-kami: As discussed: removed `DataParams` in favor of just using `Data` - but the trait impls have to stay for structopt https://docs.rs/structopt/0.3.21/structopt/#default-values - changing the parameter to option instead is not good, because then the help does not properly show the default parameter...

da-kami:
bors retry

PR #318, opened 2021-03-16
Alice spawns swaps outside the event loop

da-kami:
Instead of spawning the swap inside the event loop we send the swap back
to the caller to be spawned. This means we no longer need the remote handle
that was only used in the tests.
This now properly logs the swap results in production.
It also gives us more control over Alice's swap in the tests.

### DIFF
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -27,7 +27,7 @@ use swap::database::Database;
 use swap::execution_params::{ExecutionParams, GetExecutionParams};
 use swap::fs::default_config_path;
 use swap::monero::Amount;
-use swap::protocol::alice::EventLoop;
+use swap::protocol::alice::{run, EventLoop};
 use swap::seed::Seed;
 use swap::trace::init_tracing;
 use swap::{bitcoin, execution_params, kraken, monero};
@@ -95,7 +95,7 @@ async fn main() -> Result<()> {
 
             let kraken_rate_updates = kraken::connect()?;
 
-            let (event_loop, _) = EventLoop::new(
+            let (event_loop, mut swap_receiver) = EventLoop::new(
                 config.network.listen,
                 seed,
                 execution_params,
@@ -106,6 +106,22 @@ async fn main() -> Result<()> {
                 max_buy,
             )
             .unwrap();
+
+            tokio::spawn(async move {
+                while let Some(swap) = swap_receiver.recv().await {
+                    tokio::spawn(async move {
+                        let swap_id = swap.swap_id;
+                        match run(swap).await {
+                            Ok(state) => {
+                                tracing::debug!(%swap_id, "Swap finished with state {}", state)
+                            }
+                            Err(e) => {
+                                tracing::error!(%swap_id, "Swap failed with {:#}", e)
+                            }
+                        }
+                    });
+                }
+            });
 
             info!("Our peer id is {}", event_loop.peer_id());
 
--- a/swap/tests/bob_refunds_using_cancel_and_refund_command.rs
+++ b/swap/tests/bob_refunds_using_cancel_and_refund_command.rs
@@ -1,17 +1,20 @@
 pub mod testutils;
 
-use swap::protocol::bob;
 use swap::protocol::bob::BobState;
+use swap::protocol::{alice, bob};
 use testutils::bob_run_until::is_btc_locked;
 use testutils::FastCancelConfig;
 
 #[tokio::test]
 async fn given_bob_manually_refunds_after_btc_locked_bob_refunds() {
     testutils::setup_test(FastCancelConfig, |mut ctx| async move {
-        let (bob_swap, bob_join_handle) = ctx.new_swap_as_bob().await;
+        let (bob_swap, bob_join_handle) = ctx.bob_swap().await;
+        let bob_swap = tokio::spawn(bob::run_until(bob_swap, is_btc_locked));
 
-        let bob_state = bob::run_until(bob_swap, is_btc_locked).await.unwrap();
+        let alice_swap = ctx.alice_next_swap().await;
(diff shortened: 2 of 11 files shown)


swap/src/bin/asb.rs
```diff
@@ -107,6 +107,22 @@ async fn main() -> Result<()> {
...
+            tokio::spawn(async move {
+                while let Some(swap) = swap_receiver.recv().await {
+                    tokio::spawn(async move {
+                        let swap_id = swap.swap_id;
+                        match run(swap).await {
+                            Ok(state) => {
+                                tracing::debug!(%swap_id, "Swap finished with state {}", state)
+                            }
+                            Err(e) => {
+                                tracing::error!(%swap_id, "{:#}", e)
```
thomaseizinger, line 119:
```suggestion
                                tracing::error!(%swap_id, "Swap failed with {:#}", e)
```

swap/tests/bob_refunds_using_cancel_and_refund_command.rs
```diff
@@ -1,17 +1,20 @@
...
 #[tokio::test]
 async fn given_bob_manually_refunds_after_btc_locked_bob_refunds() {
     testutils::setup_test(FastCancelConfig, |mut ctx| async move {
-        let (bob_swap, bob_join_handle) = ctx.new_swap_as_bob().await;
+        let (bob_swap, bob_join_handle) = ctx.bob_swap().await;
+        let bob_swap = tokio::spawn(bob::run_until(bob_swap, is_btc_locked));
 
-        let bob_state = bob::run_until(bob_swap, is_btc_locked).await.unwrap();
+        let alice_swap = ctx.alice_swap_triggered_by_bob().await;
+        let _alice_swap = tokio::spawn(alice::run(alice_swap));
```
thomaseizinger, line 15: If you don't care about the handle, I would just not assign it a variable. It will not emit a warning :)


```suggestion
        tokio::spawn(alice::run(alice_swap));
```

swap/tests/happy_path_restart_bob_before_comm.rs
```diff
@@ -1,16 +1,20 @@
...
     testutils::setup_test(SlowCancelConfig, |mut ctx| async move {
-        let (bob_swap, bob_join_handle) = ctx.new_swap_as_bob().await;
+        let (bob_swap, bob_join_handle) = ctx.bob_swap().await;
+        let bob_swap = tokio::spawn(bob::run_until(bob_swap, is_xmr_locked));
 
-        let bob_state = bob::run_until(bob_swap, is_xmr_locked).await.unwrap();
+        let alice_swap = ctx.alice_swap_triggered_by_bob().await;
+        let alice_swap = tokio::spawn(alice::run(alice_swap));
+
+        let bob_state = bob_swap.await.unwrap().unwrap();
```
thomaseizinger, line 17: Might be worth considering to allow this function to return errors, then you could make this less noisy by just doing `??`.

swap/tests/testutils/mod.rs
```diff
@@ -107,7 +106,12 @@ pub struct TestContext {
 }
 
 impl TestContext {
-    pub async fn new_swap_as_bob(&mut self) -> (bob::Swap, BobEventLoopJoinHandle) {
+    pub async fn alice_swap_triggered_by_bob(&mut self) -> alice::Swap {
+        // return the swap that is triggered by bob
```
thomaseizinger, line 110: This comment does not add anything the function name doesn't already say :)

swap/tests/testutils/mod.rs
```diff
@@ -107,7 +106,12 @@ pub struct TestContext {
 }
 
 impl TestContext {
-    pub async fn new_swap_as_bob(&mut self) -> (bob::Swap, BobEventLoopJoinHandle) {
+    pub async fn alice_swap_triggered_by_bob(&mut self) -> alice::Swap {
```
thomaseizinger, line 109: Maybe `next_swap` would be a better name together with a doc comment (`///`) that explains that in a test context, the next swap will be the one that Bob triggered because there will always be just one swap.

thomaseizinger (approved):
Nice! Good work and thanks for adding the punish test back! :)

swap/src/bin/asb.rs
da-kami: Note: The `with_context` inside the run function caused problems with the lifetime of the database, that's why I remove it. Better to be added here anyway.

rishflab (approved):

swap/tests/bob_refunds_using_cancel_and_refund_command.rs
da-kami: Yeah, in this case `_` is OK - I often like to to use `_purpose` to express what we are "ignoring" - find it more readable

swap/tests/testutils/mod.rs
da-kami: nice spot, I actually had renamed the function and then forgot to remove the comment. good all rule: don't add a comment before you have not checked if you can fix it by making the code around better...

da-kami:
Also properly added `punish` test to CI and bors in the cleanup ;)

swap/src/bin/asb.rs
```diff
@@ -107,6 +107,22 @@ async fn main() -> Result<()> {
             )
             .unwrap();
 
+            tokio::spawn(async move {
+                while let Some(swap) = swap_receiver.recv().await {
+                    tokio::spawn(async move {
+                        let swap_id = swap.swap_id;
+                        match run(swap).await {
+                            Ok(state) => {
+                                tracing::debug!(%swap_id, "Swap finished with state {}", state)
```
thomaseizinger, line 116: This could be info because it is actually quite the important state change.

thomaseizinger (approved):

PR #321, opened 2021-03-17
Properly handle concurrent messages to and from peers

thomaseizinger:
Previously, we were forwarding incoming messages from peers to all
swaps that were currently running. That is obviously wrong. The new
design scopes an `EventLoopHandle` to a specific PeerId to avoid
this problem.


da-kami (approved):

rishflab (approved):

thomaseizinger:
bors retry

PR #338, opened 2021-03-18
Timeout improvements

rishflab:
If TxLock does not confirm in a reasonable amount of time, Alice should
give up on the swap rather than waiting forever. Watching for TxLock in
the mempool is no longer required and causes unnecessary complexity.
What if Alice does not see the transaction in mempool but it is already
confirmed? She will abort the swap for no reason.

### DIFF
--- a/swap/src/protocol/alice/swap.rs
+++ b/swap/src/protocol/alice/swap.rs
@@ -69,13 +69,25 @@ async fn next_state(
     Ok(match state {
         AliceState::Started { state3 } => {
             let tx_lock_status = bitcoin_wallet.subscribe_to(state3.tx_lock.clone()).await;
-            timeout(env_config.bob_time_to_act, tx_lock_status.wait_until_seen())
-                .await
-                .context("Failed to find lock Bitcoin tx")??;
-
-            tx_lock_status.wait_until_final().await?;
-
-            AliceState::BtcLocked { state3 }
+            match timeout(
+                env_config.bitcoin_lock_confirmed_timeout,
+                tx_lock_status.wait_until_final(),
+            )
+            .await
+            {
+                Err(_) => {
+                    tracing::info!(
+                        "TxLock lock did not get {} confirmations in {} minutes",
+                        env_config.bitcoin_finality_confirmations,
+                        env_config.bitcoin_lock_confirmed_timeout.as_secs_f64() / 60.0
+                    );
+                    AliceState::SafelyAborted
+                }
+                Ok(res) => {
+                    res?;
+                    AliceState::BtcLocked { state3 }
+                }
+            }
         }
         AliceState::BtcLocked { state3 } => {
             // Record the current monero wallet block height so we don't have to scan from
--- a/swap/src/env.rs
+++ b/swap/src/env.rs
@@ -5,7 +5,7 @@ use time::NumericalStdDurationShort;
 
 #[derive(Debug, Copy, Clone)]
 pub struct Config {
-    pub bob_time_to_act: Duration,
+    pub bitcoin_lock_confirmed_timeout: Duration,
     pub bitcoin_finality_confirmations: u32,
     pub bitcoin_avg_block_time: Duration,
     pub bitcoin_cancel_timelock: CancelTimelock,
@@ -42,7 +42,7 @@ pub struct Regtest;
 impl GetConfig for Mainnet {
     fn get_config() -> Config {
         Config {
-            bob_time_to_act: 10.minutes(),
+            bitcoin_lock_confirmed_timeout: 24.hours(),
             bitcoin_finality_confirmations: 3,
             bitcoin_avg_block_time: 10.minutes(),
             bitcoin_cancel_timelock: CancelTimelock::new(72),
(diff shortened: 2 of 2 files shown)


swap/src/protocol/alice/swap.rs
```diff
@@ -85,11 +85,14 @@ async fn run_until_internal(
...
+                timeout(
+                    env_config.bitcoin_lock_confirmed_timeout(),
+                    bitcoin_wallet.watch_until_status(&state3.tx_lock, |status| {
                         status.is_confirmed_with(env_config.bitcoin_finality_confirmations)
-                    })
-                    .await?;
+                    }),
+                )
+                .await
+                .context("Lock Bitcoin TX failed to confirm in time")??;
```
thomaseizinger, line 95: Isn't that exactly what happens a couple of lines above?

We first check that we see the TX in mempool within the given timeout and then we wait for confirmations.

thomaseizinger (changes requested):
One question, otherwise LGTM.

swap/src/protocol/alice/swap.rs
rishflab: Do you mean the .context message? I just pushed an amended commit to improve it

thomaseizinger: No, I mean the whole `timeout` call. We literally have the same timeout a couple of lines above where we wait for the transaction to be seen.

I also think that this is actually wrong. It is fine to race a timeout against _seeing_ the transaction but waiting for finality can take quite a bit longer than the now configured 10 minutes. What if we configure finality for more than one block, which is what would happen on mainnet?

rishflab: Ah crap I was supposed to use this function
```
pub fn bitcoin_lock_confirmed_timeout(&self) -> Duration {
         self.bitcoin_avg_block_time * self.bitcoin_finality_confirmations * 4
}
```

da-kami (dismissed):
Two things: 

1. I am not sure the second timeout is relevant. 

Once we see the transaction in the mempool there is no reason to assume that it will not be included. Granted, it can happen if there is a re-org I suppose - we can add the timeout for such kind of rare scenarios. But 4 times the blocktime seems to be arbitrary, we might want to go a bit higher to allow "checker" transactions? 

2. Please also fix the state transition as part of this:

We should not just time out and fail, but the swap should properly transition to `SafelyAborted`.

rishflab:
> Once we see the transaction in the mempool there is no reason to assume that it will not be included. Granted, it can happen if there is a re-org I suppose - we can add the timeout for such kind of rare scenarios. But 4 times the blocktime seems to be arbitrary, we might want to go a bit higher to allow "checker" transactions?

Its 4 * block time * desired_finality_confirmations.

swap/src/protocol/alice/swap.rs
thomaseizinger: Ah I didn't see that these are actually two different timeouts. My bad.

I am not fully convinced to be honest, do we really need to timeouts here or should we just wait for the transaction to be confirmed? I guess it might fall out of the mempool or something but that is pretty unlikely.

I'd rather keep it simple for now to be honest.

rishflab: If we don't timeout that means we have a "state" that we can get stuck in. Seems like a bad design to me. Maybe a better approach would be to timeout if the transaction falls out of the mempool: goes to unseen after being seen in the mempool?

swap/src/protocol/alice/swap.rs
```diff
@@ -78,20 +78,31 @@ async fn run_until_internal(
         match state {
             AliceState::Started { state3 } => {
                 timeout(
```
da-kami, line 80: I think this timeout should also lead to `SavelyAborted` - or should the swap remain in `Started` forever in the database?

rishflab: yep you are right

swap/src/protocol/alice/swap.rs
```diff
@@ -75,13 +75,25 @@ async fn run_until_internal(
...
-                .await
-                .context("Failed to find lock Bitcoin tx")??;
-
-            tx_lock_status.wait_until_final().await?;
-
-            AliceState::BtcLocked { state3 }
+            match timeout(
+                env_config.bitcoin_lock_confirmed_timeout,
+                tx_lock_status.wait_until_final(),
+            )
```
thomaseizinger, line 81: Should we just make this timeout part of `wait_until_final`? We could introduce a dedicated error type that represents the timeout so the user can still decide on what to do in case we don't want this for every `wait_until_final`.

rishflab: Going to hold off for now because there is no type safety preventing  `.wait_until_final()` from being called on other transactions. Won't really make sense for any other transaction to timeout with `bitcoin_lock_confirmed timeout`.

da-kami (approved):

thomaseizinger (changes requested):
I don't think the first two patches are meaningful given the last one. Can we drop / squash those?

rishflab:
> I don't think the first two patches are meaningful given the last one. Can we drop / squash those?

will do

thomaseizinger:
@rishflab @da-kami do you think this is changelog worthy?

rishflab:
> @rishflab @da-kami do you think this is changelog worthy?

eh don't think users would be interested in this implementation detail

PR #339, opened 2021-03-18
Bump dependency versions

thomaseizinger:
Otherwise it will take a long time for dependabot to update all of
these.


thomaseizinger:
It looks like one of the dependency upgrades causes problems with the arm build :(
Need to find out which one.

thomaseizinger:
> It looks like one of the dependency upgrades causes problems with the arm build :(
> Need to find out which one.

This should fix it: https://github.com/rust-lang/libc/pull/2115

Bumped libc to 0.2.90 which is meant to include this fix.

thomaseizinger:
@da-kami @rishflab This is ready for review!

PR #340, opened 2021-03-18
Remove connection handling from swap execution and cleanup network protocol implementations

thomaseizinger:
TODO:

- [x] Re-establish connection on Bob's side if it is dropped.
- [x] Make sure Alice buffers messages if there isn't currently a connection
- [x] Do a testnet swap to make sure it works


da-kami (commented):
❓ As far as I understand, with this change Alice never actually sets up a connection but always waits for Bob to do so.
So, Bob has to ensure that he has a connection to Alice for receiving the transfer proof. How can Bob ensure that? Whenever the connection is dropped he reconnects? - I could not find any code doing so but you mentioned that Bob should do it automatically?

Part of the last commit messages states:

```
In case we ever loose the connection to Alice, we simply stop the
event loop to abort the swap execution.
```

Our previous thinking was, that when the connection is lost we wait for refunding. I find it a bit hard to put error handling on the user. Previously we ran into this scenario quite often, but that was probably linked to another network setup problem. 
Is Bob reconnecting somehow ensured in the current solution?

thomaseizinger:
> As far as I understand, with this change Alice never actually sets up a connection but always waits for Bob to do so.

This was always the case, because Bob is likely always behind NAT so any dialling attempt would fail.

> So, Bob has to ensure that he has a connection to Alice for receiving the transfer proof. How can Bob ensure that? Whenever the connection is dropped he reconnects? - I could not find any code doing so but you mentioned that Bob should do it automatically?

My last commit was on Thursday, I did not yet actually change the code to do that. The current code fails if we loose the connection. That means as long as Bob is running, we should have a connection.

> Our previous thinking was, that when the connection is lost we wait for refunding.

I don't see how that is currently implementation. All of the calls to `dial()` where question-marked so the swap function would always fail.

> Is Bob reconnecting somehow ensured in the current solution?

No but I am planning to add that!

swap/src/protocol/bob/event_loop.rs
```diff
@@ -175,7 +174,7 @@ impl EventLoop {
         Ok((event_loop, handle))
     }
 
-    pub async fn run(mut self) -> Result<Infallible> {
+    pub async fn run(mut self) {
```
da-kami, line 177: 👍

swap/src/protocol/alice/event_loop.rs
```diff
@@ -172,6 +177,13 @@ where
                         }
                         SwarmEvent::ConnectionEstablished { peer_id: peer, endpoint, .. } => {
                             tracing::debug!(%peer, address = %endpoint.get_remote_address(), "New connection established");
+
+                            if let Some(transfer_proof) = self.buffered_transfer_proofs.remove(&peer) {
+                                tracing::debug!(%peer, "Found buffered transfer proof for peer");
+
+                                let result = self.swarm.send_transfer_proof(peer, transfer_proof);
+                                assert!(result.is_ok(), "must be able to send transfer proof after connection was established");
```
da-kami, line 185: ❓ What is the reason that you use an `assert` here instead of using `expect` directly on the send call ?

da-kami (approved):
Nice work!

swap/src/protocol/alice/event_loop.rs
thomaseizinger: No reason actually, just didn't think of `expect` :)

swap/src/protocol/bob/event_loop.rs
```diff
@@ -228,22 +208,22 @@ impl EventLoop {
...
-                            debug!("Dialing alice at {}", peer_id);
-                            libp2p::Swarm::dial(&mut self.swarm, &peer_id).context("Failed to dial alice")?;
+                        SwarmEvent::ConnectionEstablished { peer_id, endpoint, .. } if peer_id == self.alice_peer_id => {
+                            tracing::debug!("Connected to Alice at {}", endpoint.get_remote_address());
                         }
+                        SwarmEvent::Dialing(peer_id) if peer_id == self.alice_peer_id => {
+                            tracing::debug!("Dialling Alice at {}", peer_id);
+                        }
+                        SwarmEvent::ConnectionClosed { peer_id, endpoint, num_established, .. } if peer_id == self.alice_peer_id && num_established == 0 => {
+                            bail!("Lost connection to Alice at {}", endpoint.get_remote_address())
```
rishflab: "In case we ever loose the connection to Alice, we simply stop the
event loop to abort the swap execution." Will the stop the event loop and abort swap execution? Won't this prevent bob from recovering if he cannot connect to Alice.

thomaseizinger:
@da-kami @rishflab I re-did the commit history of this one to be consistent. Functionally, nothing has changed but the changes are now grouped together better.

swap/tests/happy_path_restart_bob.rs
```diff
@@ -32,3 +32,31 @@ async fn given_bob_restarts_after_xmr_is_locked_resume_swap() {
     })
     .await;
 }
+
+#[tokio::test]
+async fn given_bob_restarts_before_xmr_is_locked_resume_swap() {
```
da-kami, line 37: In the past the test execution could not handle tests in the same file. Would be great if that works now.

bors.toml
```diff
@@ -7,7 +7,8 @@ status = [
     "test (x86_64-unknown-linux-gnu, ubuntu-latest)",
     "test (x86_64-apple-darwin, macos-latest)",
     "docker_tests (happy_path)",
-    "docker_tests (happy_path_restart_bob_before_comm)",
+    "docker_tests (given_bob_restarts_after_xmr_is_locked_resume_swap)",
+    "docker_tests (given_bob_restarts_before_xmr_is_locked_resume_swap)",
```
da-kami, line 11: So far we used file names to identify the running test task - do test names actually work?

da-kami: Additionally: I think you forgot to add these tests to CI - or was that on purpose only changed in bors.toml?

swap/src/protocol/alice/event_loop.rs
```diff
@@ -172,7 +185,12 @@ where
                 next_transfer_proof = self.send_transfer_proof.next() => {
                     match next_transfer_proof {
                         Some(Ok((peer, transfer_proof))) => {
-                            self.swarm.send_transfer_proof(peer, transfer_proof);
+                            let result = self.swarm.send_transfer_proof(peer, transfer_proof);
+
+                            if let Err(transfer_proof) = result {
+                                tracing::debug!(%peer, "No active connection to peer, buffering transfer proof");
```
da-kami, line 191: 🙃 I'd say this is a `warn` level given that Bob would always reconnect this should not happen.

da-kami (approved):
Concept approved, added comments concerning the test setup

rishflab (approved):

swap/tests/happy_path_restart_bob.rs
thomaseizinger: It doesn't :(

bors.toml
thomaseizinger: Yep, I misunderstood how this was all set up :D

PR #319, opened 2021-03-16
Alice sweeps refunded funds into default wallet

da-kami:
Alice's refund scenario starts with generating the temporary wallet
from keys to claim the XMR which results in Alice' unloading the wallet.
Alice then loads her original wallet to be able to handle more swaps.
Since Alice is in the role of the long running daemon handling concurrent
swaps, the operation to close, claim and re-open her default wallet must
be atomic.
This PR adds an additional step, that sweeps all the refunded XMR back into
the default wallet. In order to ensure that this is possible, Alice has to
ensure that the locked XMR got enough confirmations.
These changes allow us to assert Alice's balance after refunding.

PR #346, opened 2021-03-18
Make lock_xmr "idempotent"

rishflab:
Alice checks if xmr has already been locked (or in the process of being
locked) before locking xmr. This change means we no longer have to keep
track of whether we have locked xmr or not. We check the status of lock
xmr by looking for confirmed outgoing, pending or mempool transfers that
 match the shared public spend key and the amount.

### DIFF
--- a/swap/src/protocol/alice/steps.rs
+++ b/swap/src/protocol/alice/steps.rs
@@ -7,6 +7,7 @@ use crate::protocol::alice::TransferProof;
 use crate::{bitcoin, monero};
 use anyhow::{bail, Context, Result};
 use futures::pin_mut;
+use std::time::Duration;
 
 pub async fn lock_xmr(
     state3: alice::State3,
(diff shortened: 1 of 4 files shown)


swap/src/protocol/alice/steps.rs
```diff
@@ -18,20 +19,25 @@ pub async fn lock_xmr(
...
-    //  Waiting for XMR confirmations should not be done in here, but in a separate
-    //  state! We have to record that Alice has already sent the transaction.
-    //  Otherwise Alice might publish the lock tx twice!
-
-    event_loop_handle
-        .send_transfer_proof(TransferProof {
-            tx_lock_proof: transfer_proof,
-        })
-        .await?;
+    if !monero_wallet
```
da-kami, line 22: Once you rebase on this PR on the changes made in https://github.com/comit-network/xmr-btc-swap/pull/319 (that are about to merge) you will have to act on the TODO and most likely have to add another state to await the 10 block confirmations. 

I think that `transfer` and sending the `transfer_proof` should stay tied together, so we might want to remove the waiting for `1` confirmation before sending the proof. I think if we retry sending he transaction upon entering we will also have to send the transfer proof immediately after.

rishflab:
Closing because cannot get transfer proof without broadcasting lock xmr until we upgrade our json rpc client

PR #351, opened 2021-03-19
Show the actual BTC amount and fee to be swapped

da-kami:
We got user feedback, that it is confusing that the amount "found" in the wallet does not match the amount actually being swapped, thus with this PR we explicitly display the amount swapped and fees.


da-kami:
Hm... recently we keep running into problems with the Monero container not starting up correctly in CI:
https://github.com/comit-network/xmr-btc-swap/pull/351/checks?check_run_id=2146222978

Any ideas what could cause this @rishflab @thomaseizinger ?

thomaseizinger (approved):

PR #361, opened 2021-03-23
Introduce a more flexible transaction subscription system 

thomaseizinger:
TODO:

- [x] Make sure we unsubscribe once all receivers are gone. How do we handle repeated subscriptions?

Will squash the last 4 or 5 commits once approved

### DIFF
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -11,14 +11,13 @@ use bdk::keys::DerivableKey;
 use bdk::{FeeRate, KeychainKind};
 use bitcoin::Script;
 use reqwest::Url;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::convert::TryFrom;
 use std::fmt;
-use std::future::Future;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use tokio::sync::Mutex;
+use tokio::sync::{watch, Mutex};
 
 const SLED_TREE_NAME: &str = "default_tree";
 
@@ -162,14 +161,13 @@ impl Wallet {
         &self,
         transaction: Transaction,
         kind: &str,
-    ) -> Result<(Txid, impl Future<Output = Result<()>> + '_)> {
+    ) -> Result<(Txid, Subscription)> {
         let txid = transaction.txid();
 
         // to watch for confirmations, watching a single output is enough
-        let watcher = self.wait_for_transaction_finality(
-            (txid, transaction.output[0].script_pubkey.clone()),
-            kind.to_owned(),
-        );
+        let subscription = self
+            .subscribe_to((txid, transaction.output[0].script_pubkey.clone()))
+            .await;
 
         self.wallet
             .lock()
@@ -181,7 +179,7 @@ impl Wallet {
 
         tracing::info!(%txid, "Published Bitcoin {} transaction", kind);
 
-        Ok((txid, watcher))
+        Ok((txid, subscription))
     }
 
     pub async fn sign_and_finalize(&self, psbt: PartiallySignedTransaction) -> Result<Transaction> {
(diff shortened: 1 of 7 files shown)


rishflab (approved):

swap/src/bitcoin/wallet.rs
```diff
@@ -215,70 +211,110 @@ impl Wallet {
...
+                };
 
-            if status_fn(new_status) {
-                break;
+                if Some(new_status) != last_status {
+                    tracing::debug!(%txid, "Transaction is {}", new_status);
+                }
+                last_status = Some(new_status);
+
+                let all_receivers_gone = sender.send(new_status).is_err();
```
da-kami, line 238: ❓ Previously this loop was ended upon a certain status being reached. Now we exit by returning if there are no subscribers (i.e. all receivers were dropped) anymore. Is it ensured that the receiver is always dropped?
If the receiver is not dropped, then we would keep sending updates even if the target state was already reched. 
I feel breaking the loop once a reached target was reached was more logical.

thomaseizinger: > Is it ensured that the receiver is always dropped?

At the latest, the receivers will be dropped once we exit from the swap function (because the subscription is not returned). Generally though, as long as the receiver is alive, some code is holding on to the subscription which means some code part is apparently still interested in status changes of this transaction. That in turn means we should keep sending updates.

> I feel breaking the loop once a reached target was reached was more logical - I hope I did not misunderstand the concept.

The idea of this refactoring is/was that instead of calling this function from several places, we only call it once and then use the subscription to check how far we have progressed. That design implies that this loop can no longer know what the target state is.

Instead of the diff, maybe a look at the final result can clear things up. With all the deleted code in the wallet, it is a bit hard to spot how things work now.

---

Your suggestion made me realize that we should "unsubscribe" once the receivers are dropped, otherwise we keep polling updates about a script forever.

swap/src/bitcoin/wallet.rs
```diff
@@ -211,7 +213,24 @@ impl Wallet {
...
+            .get(&WatchableTransaction {
+                txid: tx.id(),
+                script: tx.script(),
+            })
+        {
+            return Subscription {
+                receiver: sub.clone(),
+                finality_confirmations: self.finality_confirmations,
+                txid: tx.id(),
+            };
```
thomaseizinger, line 231: I think you could also just `clone` the full `Subscription` struct instead of storing just the `Receiver`.

swap/src/bitcoin/wallet.rs
```diff
@@ -339,12 +358,19 @@ impl Watchable for (Txid, Script) {
...
+    script: Script,
+}
+
 struct Client {
     electrum: bdk::electrum_client::Client,
     latest_block: BlockHeight,
     last_ping: Instant,
     interval: Duration,
     script_history: BTreeMap<Script, Vec<GetHistoryRes>>,
+    subscriptions: HashMap<WatchableTransaction, watch::Receiver<ScriptStatus>>,
```
thomaseizinger, line 373: Where are you inserting into this `HashMap`?

swap/src/bitcoin/wallet.rs
```diff
@@ -214,26 +215,21 @@ impl Wallet {
...
+
+        let sub = Subscription {
+            receiver,
+            finality_confirmations: self.finality_confirmations,
+            txid,
+            script: tx.script(),
         };
 
-        let (sender, receiver) = watch::channel(ScriptStatus::Unseen);
+        if let Some(inner) = self.client.lock().await.subscriptions.get(&sub) {
```
thomaseizinger, line 228: It is kind of weird to construct `Subscription` just for the lookup and then potentially throw it away if you can find an exact copy in the HashSet. Can I suggest to use as `HashMap` instead where the key is either a distinct type or maybe just a tuple of TxId and Script? That will also get rid of the custom implementation of `PartialEq` and `Hash` which are not ideal given that this type is publicly exposed from this module.

Also, because we are locking here twice (once here and once on line 261), this is actually subject to a race condition.

We should do the locking only once to make sure this can't happen. You might also want to look into the `Entry` API of `HashMap`. That one allows you to ergonomically handle a "return or insert if not present" case.

swap/src/bitcoin/wallet.rs
```diff
@@ -215,70 +213,117 @@ impl Wallet {
...
-        T: Watchable,
-    {
+    pub async fn subscribe_to(&self, tx: impl Watchable + Send + 'static) -> Subscription {
         let txid = tx.id();
+        self.client
+            .lock()
+            .await
+            .subscriptions
+            .entry(txid)
+            .or_insert({
```
thomaseizinger, line 223: This MUST be lazy, otherwise it will evaluate every time this function is called and not just when the entry is absent!

The function to use is [`or_insert_with`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with)!

It might also be worth moving this into a new function so that it is a bit clearer. Eventually, I would also like to de-couple this more from the `tokio` runtime and instead of `spawn`ing the future, just return it to the caller and they have to drive it forward (which likely means spawning).

thomaseizinger (commented):
Can't request changes but this is quite important :)

swap/src/bitcoin/wallet.rs
rishflab: fixed the lazy eval issue. 

I am not entirely sure how returning the future to caller will work so I am leaving it for now. Will it be wrapped in an option to prevent spawning multiple watchers?

thomaseizinger (commented):
LGTM!

Can't approve because it is my own PR :)

swap/src/bitcoin/wallet.rs
thomaseizinger: > I am not entirely sure how returning the future to caller will work so I am leaving it for now. Will it be wrapped in an option to prevent spawning multiple watchers?

Multiple ways to do it I guess. Could also return a dummy future that instantly finishes :)

PR #369, opened 2021-03-24
Allow ASB users to use custom Monero wallet by specyfying wallet name and password

### LINKED ISSUE #352: Give asb operators the ability to reuse an existing Monero wallet
Currently if you run asb, it creates the XMR wallet for you.

Operators should have the ability to specify a wallet that already exists.

da-kami:
Fixes #352 

Upon initial setup one can specify a wallet name and password to be used.
This allows the user to re-use an existing wallet.

If the given wallet is not found in the Monero RPC wallet folder it will be created with the given password.

### DIFF
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -2,7 +2,7 @@ use crate::fs::{default_data_dir, ensure_directory_exists};
 use anyhow::{Context, Result};
 use config::ConfigError;
 use dialoguer::theme::ColorfulTheme;
-use dialoguer::Input;
+use dialoguer::{Input, Password};
 use libp2p::core::Multiaddr;
 use serde::{Deserialize, Serialize};
 use std::ffi::OsStr;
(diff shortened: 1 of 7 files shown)


da-kami:
@thomaseizinger one note on "Is the ASB setup wizzrd useful?" - I was playing with it for testing and I find it very nice. I think it is a good way to learn what is needed to run the ASB and where the config is stored and what it contains :)

swap/src/asb/config.rs
```diff
@@ -139,6 +150,8 @@ pub fn query_user_for_initial_testnet_config() -> Result<Config> {
         bitcoin: Bitcoin { electrum_rpc_url },
         monero: Monero {
             wallet_rpc_url: monero_wallet_rpc_url,
+            wallet_name: monero_wallet_name,
+            wallet_password: monero_wallet_password,
```
thomaseizinger, line 154: This stores the password in plaintext on the disk ...

Can we change this to an env variable? E.g. if the env variable is set, we use it as the password, otherwise we default to empty.

da-kami: Yes, it does. We also store the seed there which is basically the password for the Bitcoin wallet.
How is this a worse security assumption?

da-kami:
Unfortunately not that straight forward without adding security assumptions, see https://github.com/comit-network/xmr-btc-swap/issues/352#issuecomment-805481584

PR #371, opened 2021-03-24
Configure static linking for linux-gnu and windows

### LINKED ISSUE #347: Linux Binary releases are dynamically linked
Not every Linux has libgcc.

```
/opt/xmr-btc-swap # ldd
musl libc (x86_64)
Version 1.2.2
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
/opt/xmr-btc-swap # ldd asb
        /lib64/ld-linux-x86-64.so.2 (0x7f3533d42000)
Error loading shared library libgcc_s.so.1: No such file or directory (needed by asb)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f3533d42000)
        libm.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f3533d42000)
        libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x7f3533d42000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f3533d42000)
Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by asb)
Error relocating asb: _Unwind_Resume: symbol not found
Error relocating asb: __register_atfork: symbol not found
Error relocating asb: _Unwind_SetIP: symbol not found
Error relocating asb: __res_init: symbol not found
Error relocating asb: _Unwind_GetRegionStart: symbol not found
Error relocating asb: _Unwind_GetTextRelBase: symbol not found
Error relocating asb: _Unwind_FindEnclosingFunction: symbol not found
Error relocating asb: _Unwind_DeleteException: symbol not found
Error relocating asb: _Unwind_RaiseException: symbol not found
Error relocating asb: _Unwind_GetIP: symbol not found
Error relocating asb: _Unwind_GetIPInfo: symbol not found
Error relocating asb: _Unwind_GetDataRelBase: symbol not found
Error relocating asb: _Unwind_SetGR: symbol not found
Error relocating asb: _Unwind_GetCFA: symbol not found
Error relocating asb: _Unwind_Backtrace: symbol not found
Error relocating asb: _Unwind_GetLanguageSpecificData: symbol not found
```

Would the team be open to releasing statically linked binaries?

da-kami:
I think configuring this in the `.cargo/config.toml` is the way to go.

I read through this: 
https://doc.rust-lang.org/reference/linkage.html#static-and-dynamic-c-runtimes
https://stackoverflow.com/questions/31770604/how-to-generate-statically-linked-executables

Since it mentions that this is a common problems with Windows I configured static linking for both the `linux-gnu` and `windows` target. 

I did not see any reason to release a version that is dynamically linked (i.e. release both, dynamic and static linked), but if there is a good reason please let me know!

Fixes #347


da-kami:
Unfortunately this problem is not that easy to solve, because our code contains proc-macros and those are not supported when building statically as far as I understand from these threads:

https://stackoverflow.com/questions/57965805/alpine-dockerfile-cannot-produce-proc-macro-does-not-support-these-crate-typ
https://github.com/rust-lang/rust/issues/40174
https://github.com/rust-lang/cargo/issues/7563

Not sure if this can be solved - any ideas?

da-kami:
> Unfortunately this problem is not that easy to solve for Linux, because our code contains proc-macros and those are not supported when building statically as far as I understand from these threads:
> 
> https://stackoverflow.com/questions/57965805/alpine-dockerfile-cannot-produce-proc-macro-does-not-support-these-crate-typ
> [rust-lang/rust#40174](https://github.com/rust-lang/rust/issues/40174)
> [rust-lang/cargo#7563](https://github.com/rust-lang/cargo/issues/7563)
> 
> Not sure if this can be solved - any ideas?

If I understand it correct this issue should have been solved, see: https://github.com/rust-lang/rust/issues/45601
I don't fully understand why we run into this problem.

thomaseizinger:
I am closing this as a wontfix.

Linking the binaries statically seems non-trivial.

PR #376, opened 2021-03-25
ASB resumes unfinished swaps after startup

### LINKED ISSUE #374: Resume unfinished swaps after startup
Includes:

- Fetching all unfinished swaps from database
- Resume all unfinished swaps by starting them in their known state from the database

Strategy: 

Each state of the swap execution has to be able to decide how to move forward, given that the asb might have been offline prior to starting the swap execution at the given state.

da-kami:
Fixes #374 

- [x] Save Bob peer-id in database for Alice
- [x]  Alice: Wait for `10` Monero confirmations in `BtcRefunded` instead of `XmrLocked` (requires extending the RPC to distinguish locked / unlocked balance)
- [x] Save Alice peer-id in database for Bob ~~(+ multiaddress and remove params from resume)~~
- [ ] ~~Refactor Bob in test setup (handle event-loop in test setup similar to Alice)~~

I decided against refactoring Bob in the test setup, because eventually we might still want to add concurrent swap tests with multiple Bobs. The refactoring I had in mind would not allow such kind of tests. 
Generally, the current state of the changes already contains enough added value to open the PR :)

Follow ups out of scope

- [ ] Parametrize database with role (Alice / Bob) and remove all the (currently useless) mapping between DB and protocol types.
- [ ]  Alice: Wait for transfer proof ack before transitioning to new `XmrLocked`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The `resume` command of the `swap` CLI no longer require the `--seller-peer-id` parameter.
+  This information is now saved in the database.
+
 ### Added
 
 - A changelog file.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - A changelog file.
+- Automatic resume of unfinished swaps for the `asb` upon startup.
+  Unfinished swaps from earlier versions will be skipped.
 
 ### Fixed
 
--- a/swap/src/cli/command.rs
+++ b/swap/src/cli/command.rs
@@ -37,8 +37,15 @@ pub struct Arguments {
 pub enum Command {
     /// Start a XMR for BTC swap
     BuyXmr {
+        #[structopt(
+        long = "seller-peer-id",
+        default_value = DEFAULT_ALICE_PEER_ID,
+        help = "The peer id of a specific swap partner can be optionally provided"
+        )]
+        alice_peer_id: PeerId,
+
         #[structopt(flatten)]
-        connect_params: AliceConnectParams,
+        alice_multi_addr: AliceMultiaddress,
 
         #[structopt(long = "electrum-rpc",
         help = "Provide the Bitcoin Electrum RPC URL",
@@ -60,7 +67,7 @@ pub enum Command {
         swap_id: Uuid,
 
         #[structopt(flatten)]
-        connect_params: AliceConnectParams,
+        alice_multi_addr: AliceMultiaddress,
 
         #[structopt(long = "electrum-rpc",
         help = "Provide the Bitcoin Electrum RPC URL",
@@ -108,14 +115,7 @@ pub enum Command {
 }
 
 #[derive(structopt::StructOpt, Debug)]
-pub struct AliceConnectParams {
-    #[structopt(
-        long = "seller-peer-id",
-        default_value = DEFAULT_ALICE_PEER_ID,
-        help = "The peer id of a specific swap partner can be optionally provided"
-    )]
-    pub peer_id: PeerId,
-
+pub struct AliceMultiaddress {
     #[structopt(
         long = "seller-addr",
         default_value = DEFAULT_ALICE_MULTIADDR,
--- a/swap/src/database.rs
+++ b/swap/src/database.rs
@@ -2,6 +2,7 @@ pub use alice::Alice;
 pub use bob::Bob;
 
 use anyhow::{anyhow, bail, Context, Result};
+use itertools::Itertools;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::fmt::Display;
@@ -38,11 +39,26 @@ impl Display for Swap {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq)]
+#[error("Not in the role of Alice")]
+struct NotAlice;
+
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq)]
+#[error("Not in the role of Bob")]
+struct NotBob;
+
 impl Swap {
+    pub fn try_into_alice(self) -> Result<Alice> {
+        match self {
+            Swap::Alice(alice) => Ok(alice),
+            Swap::Bob(_) => bail!(NotAlice),
+        }
+    }
+
     pub fn try_into_bob(self) -> Result<Bob> {
         match self {
             Swap::Bob(bob) => Ok(bob),
-            Swap::Alice(_) => bail!("Swap instance is not Bob"),
+            Swap::Alice(_) => bail!(NotBob),
         }
     }
 }
@@ -90,22 +106,42 @@ impl Database {
         Ok(state)
     }
 
-    pub fn all(&self) -> Result<Vec<(Uuid, Swap)>> {
-        self.0
-            .iter()
-            .map(|item| match item {
-                Ok((key, value)) => {
-                    let swap_id = deserialize::<Uuid>(&key);
-                    let swap = deserialize::<Swap>(&value).context("Failed to deserialize swap");
+    pub fn all_alice(&self) -> Result<Vec<(Uuid, Alice)>> {
+        self.all_alice_iter().collect()
+    }
 
-                    match (swap_id, swap) {
-                        (Ok(swap_id), Ok(swap)) => Ok((swap_id, swap)),
-                        (Ok(_), Err(err)) => Err(err),
-                        _ => bail!("Failed to deserialize swap"),
-                    }
-                }
-                Err(err) => Err(err).context("Failed to retrieve swap from DB"),
-            })
+    fn all_alice_iter(&self) -> impl Iterator<Item = Result<(Uuid, Alice)>> {
+        self.all_swaps_iter().map(|item| {
+            let (swap_id, swap) = item?;
+            Ok((swap_id, swap.try_into_alice()?))
+        })
+    }
+
+    pub fn all_bob(&self) -> Result<Vec<(Uuid, Bob)>> {
+        self.all_bob_iter().collect()
+    }
+
+    fn all_bob_iter(&self) -> impl Iterator<Item = Result<(Uuid, Bob)>> {
+        self.all_swaps_iter().map(|item| {
+            let (swap_id, swap) = item?;
+            Ok((swap_id, swap.try_into_bob()?))
+        })
+    }
+
+    fn all_swaps_iter(&self) -> impl Iterator<Item = Result<(Uuid, Swap)>> {
+        self.0.iter().map(|item| {
+            let (key, value) = item.context("Failed to retrieve swap from DB")?;
+
+            let swap_id = deserialize::<Uuid>(&key)?;
+            let swap = deserialize::<Swap>(&value).context("Failed to deserialize swap")?;
+
+            Ok((swap_id, swap))
+        })
+    }
+
+    pub fn unfinished_alice(&self) -> Result<Vec<(Uuid, Alice)>> {
+        self.all_alice_iter()
+            .filter_ok(|(_swap_id, alice)| !matches!(alice, Alice::Done(_)))
             .collect()
     }
 }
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -2,14 +2,19 @@
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
 ### Added
 
 - A changelog file.
+- The `asb` resumes unfinished swaps upon startup.
+  This is a breaking changed because we now save the peer-id of the counterparty in the database.
+  It is recommended to reset the database before running the upgraded version of the `asb` because any unfinished swaps will be tried for resume but will fail without the counterparty peer-id being present.
```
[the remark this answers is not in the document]
thomaseizinger: Also, can we please squash this commit into the one with the actual feature?

thomaseizinger (changes requested):
Nice effort @da-kami and thank you for structuring the commits nicely. Was fairly easy to review :)

Couple of comments!

swap/src/protocol/alice/event_loop.rs
```diff
@@ -264,8 +291,19 @@ where
...
+        // swaps save peer id so we can resume
+        match self.db.insert_peer_id(swap_id, bob_peer_id).await {
+            Ok(_) => {
+                if let Err(error) = self.swap_sender.send(swap).await {
+                    tracing::warn!(%swap_id, "Swap cannot be spawned: {}", error);
+                }
+            }
+            Err(error) => {
+                tracing::warn!(%swap_id, "Unable to save peer-id, swap cannot be spawned: {}", error);
+            }
```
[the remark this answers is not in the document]
da-kami: Because then we would save it again when we resume. We cannot distinguish if we start or resume a swap on the outside. We could say that is OK (just overwrite the peer-id upon resuming) or create a separate channel for resuming. thoughts?

CHANGELOG.md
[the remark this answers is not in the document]
da-kami: > This seems to be an unnecessary burden for the user? I'd say let's just skip the swaps where we don't have a PeerId.

Can do - and print a warning in the logs. Otherwise the user does not know that "unsupported" swaps are being skipped.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `asb` resumes unfinished swaps upon startup.
   This is a breaking changed because we now save the peer-id of the counterparty in the database.
   It is recommended to reset the database before running the upgraded version of the `asb` because any unfinished swaps will be tried for resume but will fail without the counterparty peer-id being present.
+- The `swap` CLI saves the seller-peer-id into the database to be re-used for `resume`, so the `--seller-peer-id` parameter for resume was removed.
```
thomaseizinger, line 18: I know it is not always easy to do but I think it would be worth it to strive for a wording where these entries use the "heading" as their first word.

Like:

**Added** _a changelog file_.

Additionally, I think we should word it in a "user-first" mentality. What is important to the user is that the `resume` command **changed**, not that we added stuff to the database.

My suggestion would be to move this to the **Changed** section and say:

### Changed

- The `resume` command of the `swap` CLI to no longer require the `--seller-peer-id` parameter. This information is now saved in the database.

swap/src/cli/command.rs
```diff
@@ -108,14 +115,7 @@ pub enum Command {
...
 #[derive(structopt::StructOpt, Debug)]
-pub struct AliceConnectParams {
-    #[structopt(
-        long = "seller-peer-id",
-        default_value = DEFAULT_ALICE_PEER_ID,
-        help = "The peer id of a specific swap partner can be optionally provided"
-    )]
-    pub peer_id: PeerId,
-
+pub struct AliceMultiAddress {
```
thomaseizinger, line 118: I think `Multiaddress` is one word.

swap/src/protocol/alice/event_loop.rs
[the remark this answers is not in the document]
thomaseizinger: Fair enough. I guess the 2nd element could also be an `Option<PeerId>` or you model the two different payloads as an `enum`: `NewSwap`, `ResumeSwap` where only `NewSwap` has a field for the `PeerId`.

swap/src/database.rs
```diff
@@ -38,11 +38,26 @@ impl Display for Swap {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq)]
+#[error("Not in the role of Alice")]
+struct NotAlice;
+
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq)]
+#[error("Not in the role of Bob")]
+struct NotBob;
```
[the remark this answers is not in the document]
da-kami: it's used to compare the exact error in tests :)

swap/src/database.rs
```diff
@@ -134,6 +172,12 @@ impl Database {
             Ok((swap_id, swap.try_into_bob()?))
         })
     }
+
+    pub fn unfinished_alice(&self) -> Result<Vec<(Uuid, Alice)>> {
+        self.all_iter_alice()
+            .filter_ok(|(_swap_id, alice)| !matches!(alice, Alice::Done(_)))
+            .collect()
+    }
```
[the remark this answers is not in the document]
da-kami: This adds complexity on the caller side, because we have an async call when sending the swap through the channel. Additionally I find it more readable in the current state. Won't optimize for now :)

swap/src/protocol/alice/event_loop.rs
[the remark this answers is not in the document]
da-kami: Unfortunately this has implications on what we can do in the tests. If we handle saving the peer-id in main we cannot really test this properly. Given the current test setup this is hard to change - we could extract resume into a separate component that can be instantiated in both main and test, but I would see this in a follow up.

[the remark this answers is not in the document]
da-kami: After some more thinking the proper solution would be to extract building the swap to be done on the outside.

Rough outline of this solution:

- Have a separate component to load unfinished swaps upon startup.
- Have a swap builder on the outside (main / test-setup) that constructs the swap upon events from the event loop (execution setup finished) or upon loading unfinished swaps component
    - Send the peer-id back through the channel in case we construct a new swap.
    - Load the peer-id from the database in case of resume
- Allow constructing a handle through the event loop (peer id known by caller)
- Extract spawning the swap into dedicated function (to be used for new/resume - would be great if we can use main and test).

I will see to give this a try as a follow-up. I'd say we get this merged first so we don't waste time.

swap/src/database.rs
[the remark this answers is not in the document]
thomaseizinger: Sure! I don't quite understand why it adds complexity, there is virtually no difference in iterating over a `Vec` vs an `Iterator` except that the latter evaluates lazily and if you already have an `Iterator` here, you save yourself an allocation by not calling `collect` on it.

swap/src/database.rs
```diff
@@ -62,7 +65,10 @@ impl Swap {
     }
 }
 
-pub struct Database(sled::Db);
+pub struct Database {
+    swap_state: sled::Tree,
```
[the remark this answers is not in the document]
da-kami: This is manly for separation of concerns. The peer-id is a network layer detail that the swap state should not be concerned about. When we load the swap-state we don't necessarily need the peer-id.
A second sled tree does not harm us and it brings us closer to the (eventual) relational database.

swap/src/database.rs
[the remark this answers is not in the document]
da-kami: When I gave it a try I used combinators on the outside to operate on the iterator. We have an async call to fetch the peer ID for each swap - which was then done within the combinator and that caused problems. Not sure why, maybe it was just me :)

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - A changelog file.
+- Automatic resume of unfinished swaps for the `asb` upon startup.
+  Unfinished swaps that were triggered prior to this change will be skipped by resume.
```
thomaseizinger, line 19:
```suggestion
  Unfinished swaps from earlier versions will be skipped.
```

swap/tests/happy_path_restart_alice_after_xmr_locked.rs
```diff
@@ -0,0 +1,40 @@
...
+
+        assert!(matches!(
+            alice_state,
+            AliceState::XmrLockTransactionSent { .. }
+        ));
+
+        ctx.restart_alice().await;
+        let alice_swap = ctx.alice_next_swap().await;
+        assert!(matches!(
+            alice_state,
```
thomaseizinger, line 27: I think you may have wanted to assert the state of the loaded swap here? Currently, this assertion does not add anything over the one on line 19.

swap/tests/harness/mod.rs
```diff
@@ -105,11 +122,32 @@ pub struct TestContext {
...
+        self.alice_handle = alice_handle;
+        self.alice_swap_handle = alice_swap_handle;
+    }
+
     pub async fn alice_next_swap(&mut self) -> alice::Swap {
-        self.alice_swap_handle.recv().await.unwrap()
+        match timeout(Duration::from_secs(10), self.alice_swap_handle.recv()).await {
+            Err(_) => {
+                panic!("No Alice swap within 10 seconds, aborting because this test is waiting for a swap forever...")
+            }
```
thomaseizinger, line 145: This could also be a `.expect`.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The `resume` command of the `swap` CLI to no longer require the `--seller-peer-id` parameter.
```
thomaseizinger, line 12:
```suggestion
- The `resume` command of the `swap` CLI no longer require the `--seller-peer-id` parameter.
```

swap/tests/harness/mod.rs
```diff
@@ -484,7 +484,7 @@ where
     let cli = Cli::default();
 
     let _guard = tracing_subscriber::fmt()
-        .with_env_filter("warn,swap=debug,monero_harness=debug,monero_rpc=debug,bitcoin_harness=info,testcontainers=info")
+        .with_env_filter("warn,swap=debug,monero_harness=debug,monero_rpc=info,bitcoin_harness=info,testcontainers=info")
```
thomaseizinger: I think this change should go into an earlier commit :)

swap/tests/happy_path_restart_alice_after_xmr_locked.rs
```diff
@@ -12,7 +12,7 @@ async fn given_alice_restarts_after_xmr_is_locked_resume_swap() {
         let bob_swap = tokio::spawn(bob::run(bob_swap));
 
         let alice_swap = ctx.alice_next_swap().await;
-        let alice_swap = tokio::spawn(alice::run_until(alice_swap, is_xmr_locked));
+        let alice_swap = tokio::spawn(alice::run_until(alice_swap, is_xmr_lock_transaction_sent));
```
thomaseizinger, line 15: Given that this commit changes an earlier test, we could just move this change into the commit where `AliceState::XmrLockTransactionSent` was introduced? Seems more correct.

swap/tests/alice_punishes_after_restart_punish_timelock_expired.rs
```diff
@@ -0,0 +1,62 @@
...
+use harness::bob_run_until::is_btc_locked;
+use harness::FastPunishConfig;
+use swap::protocol::alice::AliceState;
+use swap::protocol::bob::BobState;
+use swap::protocol::{alice, bob};
+
+/// Bob locks Btc and Alice locks Xmr. Bob does not act; he fails to send Alice
+/// the encsig and fail to refund or redeem. Alice punishes.
+#[tokio::test]
+async fn alice_punishes_after_restart_if_punish_timelock_expired() {
```
thomaseizinger, line 13: If all this commit is doing is adding a test (see comments on other changes in here), it can could also go into the previous one :)

swap/src/protocol/bob/cancel.rs
```diff
@@ -26,7 +26,15 @@ pub async fn cancel(
         BobState::XmrLocked(state4) => state4.cancel(),
         BobState::EncSigSent(state4) => state4.cancel(),
         BobState::CancelTimelockExpired(state6) => state6,
-        _ => bail!(
+
```
thomaseizinger, line 29: Is this newline on purpose? Looks odd.

thomaseizinger (approved):
Thank you for working in the comments!

Only a few nits left, feel free to address in a follow-up or not at all!

thomaseizinger (approved):

swap/tests/alice_refunds_after_restart_bob_refunded.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use harness::alice_run_until::is_xmr_lock_transaction_sent;
+use harness::FastCancelConfig;
+use swap::protocol::alice::AliceState;
+use swap::protocol::{alice, bob};
+
+/// Bob locks Btc and Alice locks Xmr. Alice does not act so Bob refunds.
+/// Eventually Alice comes back online and refunds as well.
+#[tokio::test]
+async fn alice_punishes_after_restart_if_bob_already_refunded() {
```
thomaseizinger, line 11: Test name is wrong, says `punish` :)

da-kami: wuups thanks! - classic copy-paster. I'll fix it :)

PR #379, opened 2021-03-25
Use language agnostic heuristic to check if monero_wallet_rpc is ready

### LINKED ISSUE #353: Bad line detection when swap start monero_wallet_rpc
The line detection to detect if monero_wallet_rpc has been started in source swap/src/monero/wallet_rpc.rs:150 can be translated in many different language. So the line detection can't work if the local settings on the computer is configured to be other than english like in french for example.

rishflab:
Our strategy of searching for a english string to determine if
monero_wallet_rpc is ready is not compatible with languages other than
english. Instead we assume the monero rpc is ready if it has stopped
writing to stdout. We make a json rpc request to confirm this. A better
solution would have been to configure the monero_wallet_rpc to always
output in english but there is not command line argument to configure
the language.

Closes #353.

NOTE: will squash after approved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,4 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - A changelog file.
 
+### Fixed
+
+- Make monero_wallet_rpc readiness check language agnostic. The readiness check was
+  failing on non-english language systems preventing users from starting the swap-cli
+  and asb.
+
 [Unreleased]: https://github.com/comit-network/xmr-btc-swap/compare/v0.3...HEAD
--- a/swap/src/monero/wallet_rpc.rs
+++ b/swap/src/monero/wallet_rpc.rs
@@ -2,11 +2,13 @@ use ::monero::Network;
 use anyhow::{Context, Result};
 use big_bytes::BigByte;
 use futures::{StreamExt, TryStreamExt};
+use monero_rpc::wallet::Client;
 use reqwest::header::CONTENT_LENGTH;
 use reqwest::Url;
 use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::time::Duration;
 use tokio::fs::{remove_file, OpenOptions};
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::process::{Child, Command};
@@ -146,11 +148,15 @@ impl WalletRpc {
 
         let mut reader = BufReader::new(stdout).lines();
 
-        while let Some(line) = reader.next_line().await? {
-            if line.contains("Starting wallet RPC server") {
-                break;
-            }
+        // If we do not hear from the monero_wallet_rpc process for 3 seconds we assume
+        // it is is ready
+        while let Ok(line) = tokio::time::timeout(Duration::from_secs(3), reader.next_line()).await
+        {
+            line?;
         }
+
+        // Send a json rpc request to make sure monero_wallet_rpc is ready
+        Client::localhost(port).get_version().await?;
 
         Ok(WalletRpcProcess {
             _child: child,
--- a/swap/src/monero/wallet_rpc.rs
+++ b/swap/src/monero/wallet_rpc.rs
@@ -8,7 +8,6 @@ use reqwest::Url;
 use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
-use std::time::Duration;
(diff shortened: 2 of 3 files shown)


swap/src/monero/wallet_rpc.rs
```diff
@@ -146,12 +148,18 @@ impl WalletRpc {
...
+        // it is is ready
+        while let Ok(res) = tokio::time::timeout(Duration::from_secs(3), reader.next_line()).await {
+            res?;
         }
 
+        // Send a json rpc request to make sure monero_wallet_rpc is ready
+        Client::localhost(port)
+            .get_version()
+            .await
+            .expect("monero wallet rpc is ready to accept json rpc requests");
```
thomaseizinger, line 161: Why are we panicking here instead of returning a Result?

swap/src/monero/wallet_rpc.rs
```diff
@@ -146,12 +148,18 @@ impl WalletRpc {
 
         let mut reader = BufReader::new(stdout).lines();
 
-        while let Some(line) = reader.next_line().await? {
-            if line.contains("Starting wallet RPC server") {
-                break;
-            }
+        // If we do not hear from the monero_wallet_rpc process for 3 seconds we assume
+        // it is is ready
+        while let Ok(res) = tokio::time::timeout(Duration::from_secs(3), reader.next_line()).await {
```
thomaseizinger, line 153: I know that technically, `res` is not the read line but the `Result` but can we still call it `line`? Makes it easier to understand I think.

thomaseizinger (approved):
Fine with me. Out of curiosity: Have you tried figuring out how `monero-wallet-rpc` determines what the current locale is? I think at least on Unix, this happens through the `LANG` env variable so we could just overwrite that when starting it.

thomaseizinger:
Please update the changelog! :)

swap/src/monero/wallet_rpc.rs
rishflab: fixed

rishflab:
> Fine with me. Out of curiosity: Have you tried figuring out how `monero-wallet-rpc` determines what the current locale is? I think at least on Unix, this happens through the `LANG` env variable so we could just overwrite that when starting it.

Just tried it and I can change the language by setting the `LANG` env variable. I will have to investigate further to make sure windows works the same way.

CHANGELOG.md
```diff
@@ -11,4 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - A changelog file.
 
+### Fixed
+
+- Make monero_wallet_rpc readiness check language agnostic. The readiness check was
+  failing on non-english language systems preventing users from starting the swap-cli
+  and asb.
```
thomaseizinger, line 18: I'd prefer if we stick to the one-sentence-per-line rule for text source files like this.
Git works on a line-by-line basis and in natural language, one usually changes full sentences.

thomaseizinger:
> > Fine with me. Out of curiosity: Have you tried figuring out how `monero-wallet-rpc` determines what the current locale is? I think at least on Unix, this happens through the `LANG` env variable so we could just overwrite that when starting it.
> 
> Just tried it and I can change the language by setting the `LANG` env variable. I am not sure how to test it on windows without merging it and making a release nor could I find anything in the source. Not sure how to proceed, maybe a compile time flag for windows or we can just release it and see if windows users are having the issue?

Yeah not sure if that would work on Windows. We could add an integration test for the whole `wallet_rpc` module maybe? We need to start running tests on Windows in CI then but that might be a good idea anyway.

Once that is in place, you could test it with a commit that sets the language to french for example and if the test then fails on Windows, then you know that it respects the `LANG` variable.

I just checked and `monero-wallet-rpc` starts up fine even if the provided `daemon-host` is garbage (i.e. `localhost` for example) so you can specify that in the tests without need a network connection.

Thinking about it, we would need a network connection anyway for the download so I am not sure if that is such a good idea in the end :sweat_smile:

CHANGELOG.md
rishflab: We will have to configure dprint to accept it then. Doesn't pass CI unless I `dprint fmt`

thomaseizinger: Oh really? :(

My beloved dprint is wrapping markdown files at a certain character range? Damn.

thomaseizinger: I guess we could just set the line length to 200 chars or something: https://dprint.dev/plugins/markdown/config/

thomaseizinger (approved):
I think we can merge this as is :)

PR #383, opened 2021-03-26
Improve resilience of balance assertions

thomaseizinger:
The final commit is the relevant patch!

It sits on top of several refactoring commits that happened while I was debugging why things didn't work as expected. Turned out to be reasonably useful so I just left them in :)


swap/tests/testutils/mod.rs
```diff
@@ -226,11 +226,6 @@ impl TestContext {
             self.bob_starting_balances.btc - self.btc_amount - lock_tx_bitcoin_fee
         );
 
-        // unload the generated wallet by opening the original wallet
-        self.bob_monero_wallet.re_open().await.unwrap();
```
da-kami, line 245: You cannot remove this! You need this on Bob's side unless you reload his default wallet.

Alice: We re-load the default wallet (because concurrent swaps...)
Bob: We **don't** reload the default wallet - we just sweep the wallet that was `generated_from_keys` - in the tests we send the money to the original wallet (in reality it would go to any user wallet specified with address upon startup).

da-kami (changes requested):
You cannot remove the `re-load` of Bob's wallet in the tests, see my comment.

da-kami (approved):
Thanks for all the cleanup on top of the feature :)

I totally agree with the change, but would like to know for which specific scenario did we run into the problem of having to await the balance because the "block was not mined yet"? 
I am surprised that we did not run into this problem earlier and would like to understand what specifically triggered it now.

thomaseizinger:
> I totally agree with the change, but would like to know for which specific scenario did we run into the problem of having to await the balance because the "block was not mined yet"?

[This](https://github.com/comit-network/xmr-btc-swap/runs/2197842702?check_suite_focus=true) particular test run triggered looking into this and you suggested that it might have to do with the transaction not yet having a confirmation :)

thomaseizinger:
>  I am surprised that we did not run into this problem earlier and would like to understand what specifically triggered it now.

IIRC, you mentioned that it might have been the sweeping feature that was introduced recently: https://github.com/comit-network/xmr-btc-swap/pull/319

da-kami:
> > I totally agree with the change, but would like to know for which specific scenario did we run into the problem of having to await the balance because the "block was not mined yet"?
> 
> [This](https://github.com/comit-network/xmr-btc-swap/runs/2197842702?check_suite_focus=true) particular test run triggered looking into this and you suggested that it might have to do with the transaction not yet having a confirmation :)

Ah yes, the scenario where Alice asserts correctly that `XMR` is refunded, but the XMR balance did not change correctly. 
If this is actually a problem, I am surprised we did not run into this problem with Bob's redeem before. For Bob's redeem we:

1. use `generate_from_keys` to create and load the wallet to claim the XMR
2. refreshed the newly created wallet, 
3. sweeped all funds to the default wallet, 
4. reload the default wallet, 
5. refresh the default wallet
6. assert redeem balance

I would have assumed that the same should work for Alice's refund. Maybe it was a coincident that it never happened for Bob and it was brittle all along.
The "improved assertion" is definitely more resilient. Still, but strange that this was needed now 😅

PR #387, opened 2021-03-30
Improve the resilience of the network layer

thomaseizinger:
We improve the resilience in two ways:

1. Use a timeout on Bob's side for the execution-setup.
2. Use the `bmrng` library to model the communication between Alice and Bob.

See commit messages for details.

### DIFF
--- a/swap/src/protocol/alice/event_loop.rs
+++ b/swap/src/protocol/alice/event_loop.rs
@@ -10,14 +10,25 @@ use anyhow::{bail, Context, Result};
 use futures::future;
 use futures::future::{BoxFuture, FutureExt};
 use futures::stream::{FuturesUnordered, StreamExt};
+use libp2p::request_response::{RequestId, ResponseChannel};
 use libp2p::swarm::SwarmEvent;
 use libp2p::{PeerId, Swarm};
 use rand::rngs::OsRng;
 use std::collections::HashMap;
 use std::convert::Infallible;
 use std::sync::Arc;
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::mpsc;
 use uuid::Uuid;
+
+/// A future that resolves to a tuple of `PeerId`, `transfer_proof::Request` and
+/// `Responder`.
+///
+/// When this future resolves, the `transfer_proof::Request` shall be sent to
+/// the peer identified by the `PeerId`. Once the request has been acknowledged
+/// by the peer, i.e. a `()` response has been received, the `Responder` shall
+/// be used to let the original sender know about the successful transfer.
+type OutgoingTransferProof =
+    BoxFuture<'static, Result<(PeerId, transfer_proof::Request, bmrng::Responder<()>)>>;
 
 #[allow(missing_debug_implementations)]
 pub struct EventLoop<RS> {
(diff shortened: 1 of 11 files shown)


rishflab (approved):

swap/src/protocol/bob/event_loop.rs
```diff
@@ -88,6 +88,7 @@ impl EventLoop {
         let _ = Swarm::dial(&mut self.swarm, &self.alice_peer_id);
 
         loop {
+            // Note: We are making very elaborate use of `select!` macro's feature here. Make sure to read the documentation thoroughly: https://docs.rs/tokio/1.4.0/tokio/macro.select.html
```
da-kami, line 91:
```suggestion
            // Note: We are making very elaborate use of `select!` macro's precondition feature here. Make sure to read the documentation thoroughly: https://docs.rs/tokio/1.4.0/tokio/macro.select.html
```

swap/src/protocol/bob/event_loop.rs
```diff
@@ -164,22 +165,26 @@ impl EventLoop {
                         _ => {}
                     }
                 },
-                Some((request, responder)) = self.spot_price_requests.next().fuse() => {
+
+                // Handle to-be-sent requests for all our network protocols.
+                // Use `self.is_connected_to_alice` as a guard to "buffer" requests until we are connected.
+                Some((request, responder)) = self.spot_price_requests.next().fuse(), if self.is_connected_to_alice() => {
```
da-kami, line 171: ❓ Does the spot price request have a timeout? Otherwise the program might hang upon startup if Alice is not available.

swap/src/protocol/bob/event_loop.rs
```diff
@@ -164,22 +165,26 @@ impl EventLoop {
...
                     let id = self.swarm.spot_price.send_request(&self.alice_peer_id, request);
                     self.inflight_spot_price_requests.insert(id, responder);
                 },
-                Some(((), responder)) = self.quote_requests.next().fuse() => {
+                Some(((), responder)) = self.quote_requests.next().fuse(), if self.is_connected_to_alice() => {
```
da-kami, line 175: ❓ To re-validate my understanding - please correct me if I'm wrong: The spot-price protocol is independent of the swap and just requests the current BTC price for one XMR. The quote requests returns the amount of XMR for a given amount of BTC and these amounts are used in the execution setup.

da-kami (approved):
A few questions - some are not directly related to the PR (sorry) but made sense in the context so I asked them as well :)

swap/src/protocol/bob/event_loop.rs
thomaseizinger: Actually, I meant to say _features_ here. Will fix!

swap/src/protocol/bob/event_loop.rs
thomaseizinger: Yes, all request-response protocols have the default timeout of 20 seconds.

swap/src/protocol/bob/event_loop.rs
thomaseizinger: Not quite:

Quote is independent from swap. You can make as many quote requests as you want.

Spot-price must happen as a pre-condition to the execution-setup and the amount of the spot price protocol will be used for the execution. Spot-price should really be merged into execution-setup ...

swap/src/protocol/bob/event_loop.rs
```diff
@@ -32,38 +51,34 @@ impl EventLoop {
...
         let handle = EventLoopHandle {
-            start_execution_setup: start_execution_setup.sender,
-            done_execution_setup: done_execution_setup.receiver,
-            recv_transfer_proof: recv_transfer_proof.receiver,
-            send_encrypted_signature: send_encrypted_signature.sender,
-            request_spot_price: request_spot_price.sender,
-            recv_spot_price: recv_spot_price.receiver,
-            request_quote: request_quote.sender,
-            recv_quote: recv_quote.receiver,
+            execution_setup: execution_setup.0,
```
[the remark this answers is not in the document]
thomaseizinger: I don't know, I found the amount and length of these variables really annoying and hard to understand :)

The type system makes sure you are not mixing up sender and receiver and making a channel always returns a tuple so I am not sure how much value additional variables add.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -160,13 +163,8 @@ where
...
 
-                            match self.swarm.send_quote(channel, quote) {
-                                Ok(_) => {},
-                                Err(e) => {
-                                    // if we can't respond, the peer probably just disconnected so it is not a huge deal, only log this on debug
-                                    debug!(%peer, "failed to respond with quote: {:#}", e);
-                                    continue;
-                                }
+                            if self.swarm.quote.send_response(channel, quote).is_err() {
+                                debug!(%peer, "failed to respond with quote");
```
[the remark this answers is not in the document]
thomaseizinger: Yes, it is the last `match` of this scope so jumping to the top of the loop is implicit :)

swap/src/protocol/bob/event_loop.rs
thomaseizinger: Might do in a follow-up, don't want to wait for too many builds again :)

PR #391, opened 2021-03-31
asb Bitcoin withdraw and balance commands

### LINKED ISSUE #368: Bitcoin withdraw and balance command
Currently there is no easy way to know how much Bitcoin the ASB holds, and there is no way to withdraw the Bitcoin. Eventually someone running the ASB will want to withdraw the Bitcoin. It is not trivial to access the ASB's Bitcoin wallet without the code of the ASB, thus we should add functionality for balance and withdraw.

Proposal:

`asb withdraw --amount BTC_AMOUNT --address BITCOIN_ADDRESS`

`asb balance` (print both monero and bitcoin balance for convenience)

Make sure to NOT access the ASB database when implementing these commands, so we can run it in parallel to the ASB `start` command (which locks the database).

da-kami:
Fixes #368

Note: Balance prints both balances - which assumes that the Monero wallet RPC is running. I think that is fine for now.

### DIFF
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -13,11 +13,8 @@
 #![allow(non_snake_case)]
 
 use anyhow::{Context, Result};
-use bdk::descriptor::Segwitv0;
-use bdk::keys::DerivableKey;
 use libp2p::Swarm;
 use prettytable::{row, Table};
-use std::path::Path;
 use std::sync::Arc;
 use structopt::StructOpt;
 use swap::asb::command::{Arguments, Command};
@@ -71,8 +68,6 @@ async fn main() -> Result<()> {
     let db = Database::open(config.data.dir.join(db_path).as_path())
         .context("Could not open database")?;
 
-    let wallet_data_dir = config.data.dir.join("wallet");
-
     match opt.cmd {
         Command::Start { max_buy } => {
             let seed = Seed::from_file_or_generate(&config.data.dir)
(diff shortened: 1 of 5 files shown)


da-kami:
Known limitation upon withdraw:

If there is a UTXO that is below dust we run into an error with te BDK wallet drain:

```
Mar 31 17:44:54.572  INFO Initialized tracing with level: debug
Mar 31 17:44:54.574  INFO Using config file at default path: /Users/dakami/CoBloX/TEMP/swap-data-dir/alice/config.toml
Mar 31 17:44:54.575  INFO Database and Seed will be stored in directory: /Users/dakami/CoBloX/TEMP/swap-data-dir/alice
Mar 31 17:44:54.575 DEBUG Opening database at /Users/dakami/CoBloX/TEMP/swap-data-dir/alice/database
Mar 31 17:44:54.762 DEBUG Reading in seed from /Users/dakami/CoBloX/TEMP/swap-data-dir/alice/seed.pem
thread 'main' panicked at 'attempt to add with overflow', /Users/dakami/.cargo/registry/src/github.com-1ecc6299db9ec823/bdk-0.5.0/src/wallet/coin_selection.rs:316:31
```

swap/src/bin/asb.rs
```diff
@@ -78,6 +78,23 @@ async fn main() -> Result<()> {
             let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
+            let bitcoin_balance = bitcoin_wallet.balance().await?;
+            info!(
+                "Connection to Bitcoin wallet succeeded, balance: {}",
```
thomaseizinger, line 83: I don't think this "Connection to Bitcoin wallet succeeded" part is very useful. Can we remove it? That would also nicely align the message with the Monero balance (where we indeed do connect to an external wallet, lol)

thomaseizinger (approved):

da-kami:
PR on BDK already merged so I updated the dependency against the BDK repo instead of the fork :)

PR #394, opened 2021-04-01
Add a configurable spread to the ASB

### LINKED ISSUE #381: Configurable spread
Currently we just go by the market rate of Kraken.

A configurable spreak in percent on top of the market price is needed.

thomaseizinger:
Fixes #381.


swap/src/protocol/alice/event_loop.rs
```diff
@@ -307,11 +307,14 @@ impl LatestRate for FixedRate {
     }
 }
 
-impl LatestRate for kraken::RateUpdateStream {
+impl LatestRate for kraken::PriceUpdates {
```
da-kami: ❓If I understand it correctly you renamed everything from Kraken to `price`, but once you apply the spread it is a `rate`. Additionally, a `SpreadRate` was introduced. Is there any reasoning for the distinction between price, rate and spread-rate?

swap/src/protocol/alice/event_loop.rs
```diff
@@ -307,12 +308,27 @@ impl LatestRate for FixedRate {
     }
 }
 
-impl LatestRate for kraken::PriceUpdates {
+#[derive(Debug)]
+pub struct SpreadRate {
```
da-kami, line 312: 🔧 I think `LatestRate` was a better name, because it expresses better what this is. This combines the latest price update with the `Rate` - the fact that the rate applies a spread is an implementation detail.

da-kami (approved):

rishflab (approved):

swap/src/protocol/alice/event_loop.rs
thomaseizinger: Yes.

So the idea was:

1. Kraken gives us price updates.
2. A `Rate` is what we use to calculate our quotes. With the addition of the spread, the rate is a product of a (market) price plus the spread.
3. The `SpreadRate` is a sister-type of the `FixedRate`. Both have an implementation of `LatestRate` that allows to compute whatever the current latest rate is.

swap/src/protocol/alice/event_loop.rs
thomaseizinger: `LatestRate` still exists, it is the trait name.

> This combines the latest price update with the Rate - the fact that the rate applies a spread is an implementation detail.

I'll try and think of a better name.

da-kami (approved):
Nice! I find it more understandable like this, thanks for the changes.
I am not sure if the rename from `RateUpdate` to `PriceUpdate` adds that much value, but I don't have strong feelings about that.

thomaseizinger:
> I am not sure if the rename from `RateUpdate` to `PriceUpdate` adds that much value, but I don't have strong feelings about that.

Well, if I no longer construct `Rate`s inside the kraken module then it is not a `RateUpdate` :)

da-kami:
> > I am not sure if the rename from `RateUpdate` to `PriceUpdate` adds that much value, but I don't have strong feelings about that.
> 
> Well, if I no longer construct `Rate`s inside the kraken module then it is not a `RateUpdate` :)

Fail enough :)

PR #401, opened 2021-04-06
Remodel `EventLoop` as a future instead of as a struct

thomaseizinger:
TODO: Commits need re-ordering / squashing before this can go in. Too many things changed with the latest rebase. Converted to draft for now.

After some discussions in the libp2p repo [here](https://github.com/libp2p/rust-libp2p/discussions/2024), I thought this design might actually be easier to understand because it involves less state in a way by moving things to local variables.

The naming is also slightly more accurate IMO because the variables `event_loop` now actually contains a `Future` that contains a loop! Before it was a struct that you had to call `run` on.

### DIFF
--- a/swap/src/protocol/alice/event_loop.rs
+++ b/swap/src/protocol/alice/event_loop.rs
@@ -4,13 +4,12 @@ use crate::env::Config;
 use crate::monero::BalanceTooLow;
 use crate::network::quote::BidQuote;
 use crate::network::{spot_price, transfer_proof};
-use crate::protocol::alice::{AliceState, Behaviour, OutEvent, State0, State3, Swap};
+use crate::protocol::alice::{AliceState, Behaviour, OutEvent, State0, Swap};
 use crate::{bitcoin, kraken, monero};
 use anyhow::{bail, Context, Result};
 use futures::future;
 use futures::future::{BoxFuture, FutureExt};
 use futures::stream::{FuturesUnordered, StreamExt};
-use libp2p::request_response::{RequestId, ResponseChannel};
 use libp2p::swarm::SwarmEvent;
 use libp2p::{PeerId, Swarm};
 use rand::rngs::OsRng;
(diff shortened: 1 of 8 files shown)


rishflab (approved):

swap/src/protocol/alice/event_loop.rs
```diff
@@ -105,34 +106,10 @@ pub async fn new<LR>(
...
-                        }
-
-                        let state0 = match State0::new(btc, xmr, env_config, bitcoin_wallet.as_ref(), &mut OsRng).await {
-                            Ok(state) => state,
-                            Err(e) => {
-                                tracing::warn!(%peer, "Failed to make State0 for execution setup: {:#}", e);
-                                continue;
-                            }
+                    SwarmEvent::Behaviour(OutEvent::SpotPriceRequested { request: spot_price::Request { btc }, channel, peer }) => {
+                        if let Err(e) = handle_spot_price_request(&mut latest_rate, &mut swarm, monero_wallet.as_ref(), bitcoin_wallet.as_ref(), env_config, channel, peer, btc, max_buy).await {
```
da-kami, line 110: Looks like `fmt` did not pick up the formatting in the `select!` - might be worth manually adding a breakline.

da-kami (approved):

thomaseizinger:
@da-kami @rishflab force pushed to resolve conflicts but also added 0c47cb1 on top of it because I wasn't quite happy with the loss of documentation due to the removal of fields. It is quite similar to what it was before :sweat_smile: 

I think it is still better though :)

swap/src/protocol/alice/event_loop.rs
thomaseizinger: Yeah, anything in the macro is never formatted. because it is opaque code to the formatter.

da-kami (approved):

thomaseizinger:
Going to close this for now due to being stale and maybe revisit another time.

PR #405, opened 2021-04-08
Concurrent swaps with same peer

### LINKED ISSUE #367: Allow multiple concurrent swaps with same peer on protocol level
Follow up to: https://github.com/comit-network/xmr-btc-swap/issues/334
The dis-allowance is to be removed once the protocol was extended. 

Change the network layer protocol to include the swap ID:

We have to agree on a swap ID between the parties as well. At the moment both parties generate their own swap ID, which makes it impossible for one party to identify the swap of another party by ID. 
One way to solve this, is for Alice to send a swap ID to be used to Bob during the setup phase. Alice can then identify incoming swaps as "known" - requests with unknown swap ID can be dropped.

da-kami:
Fixes #367 

- [x] Concurrent swaps with same peer

Not sure how much more time I should invest into this. We could just merge the current state and then do improvements on top...?

Improvements:

- [x] Think `// TODO: Remove unnecessary swap-id check` through and remove it
- [x] Add concurrent swap test, multiple swaps with same Bob
- [ ] Save swap messages without matching swap in execution in the database
- [ ] Assert the balances in the new concurrent swap tests
- [ ] ~~Add concurrent swap test, multiple swaps with different Bobs~~
- [ ] ~~Send swap-id in separate message, not on top of `Message0`~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Allow multiple concurrent swaps with the same peer on the ASB.
+  This is a breaking change because the swap ID is now agreed upon between CLI and ASB during swap setup.
+  Resuming swaps started prior to this change can result in unexpected behaviour.
+
 ## [0.4.0] - 2021-04-06
 
 ### Changed
--- a/monero-harness/src/lib.rs
+++ b/monero-harness/src/lib.rs
@@ -98,7 +98,7 @@ impl<'c> Monero {
         Ok(wallet)
     }
 
-    pub async fn init(&self, wallet_amount: Vec<(&str, u64)>) -> Result<()> {
+    pub async fn init_miner(&self) -> Result<()> {
         let miner_wallet = self.wallet("miner")?;
         let miner_address = miner_wallet.address().await?.address;
 
@@ -108,16 +108,33 @@ impl<'c> Monero {
         tracing::info!("Generated {:?} blocks", res.blocks.len());
         miner_wallet.refresh().await?;
 
-        for (wallet, amount) in wallet_amount.iter() {
-            if *amount > 0 {
-                let wallet = self.wallet(wallet)?;
-                let address = wallet.address().await?.address;
-                miner_wallet.transfer(&address, *amount).await?;
+        Ok(())
+    }
+
+    pub async fn init_wallet(&self, name: &str, amount_in_outputs: Vec<u64>) -> Result<()> {
+        let miner_wallet = self.wallet("miner")?;
+        let miner_address = miner_wallet.address().await?.address;
+        let monerod = &self.monerod;
+
+        let wallet = self.wallet(name)?;
+        let address = wallet.address().await?.address;
+
+        for amount in amount_in_outputs {
+            if amount > 0 {
+                miner_wallet.transfer(&address, amount).await?;
                 tracing::info!("Funded {} wallet with {}", wallet.name, amount);
                 monerod.client().generate_blocks(10, &miner_address).await?;
                 wallet.refresh().await?;
             }
         }
+
+        Ok(())
+    }
+
+    pub async fn start_miner(&self) -> Result<()> {
+        let miner_wallet = self.wallet("miner")?;
+        let miner_address = miner_wallet.address().await?.address;
+        let monerod = &self.monerod;
 
         monerod.start_miner(&miner_address).await?;
 
@@ -127,6 +144,13 @@ impl<'c> Monero {
             .wait_for_wallet_height(block_height)
             .await
             .unwrap();
+
+        Ok(())
+    }
+
+    pub async fn init_and_start_miner(&self) -> Result<()> {
+        self.init_miner().await?;
+        self.start_miner().await?;
 
         Ok(())
     }
--- a/monero-harness/src/lib.rs
+++ b/monero-harness/src/lib.rs
@@ -68,13 +68,15 @@ impl<'c> Monero {
 
         let miner = "miner";
         tracing::info!("Starting miner wallet: {}", miner);
-        let (miner_wallet, miner_container) = MoneroWalletRpc::new(cli, &miner, &monerod).await?;
+        let (miner_wallet, miner_container) =
+            MoneroWalletRpc::new(cli, &miner, &monerod, prefix.clone()).await?;
 
         wallets.push(miner_wallet);
         containers.push(miner_container);
         for wallet in additional_wallets.iter() {
             tracing::info!("Starting wallet: {}", wallet);
-            let (wallet, container) = MoneroWalletRpc::new(cli, &wallet, &monerod).await?;
+            let (wallet, container) =
+                MoneroWalletRpc::new(cli, &wallet, &monerod, prefix.clone()).await?;
             wallets.push(wallet);
             containers.push(container);
         }
@@ -209,6 +211,7 @@ impl<'c> MoneroWalletRpc {
         cli: &'c Cli,
         name: &str,
         monerod: &Monerod,
+        prefix: String,
     ) -> Result<(Self, Container<'c, Cli, image::Monero>)> {
         let wallet_rpc_port: u16 =
             port_check::free_local_port().ok_or_else(|| anyhow!("Could not retrieve free port"))?;
@@ -218,7 +221,8 @@ impl<'c> MoneroWalletRpc {
 
         let network = monerod.network.clone();
         let run_args = RunArgs::default()
-            .with_name(name)
+            // prefix the container name so we can run multiple tests
+            .with_name(format!("{}{}", prefix, name))
             .with_network(network.clone())
             .with_mapped_port(Port {
                 local: wallet_rpc_port,
--- a/monero-harness/tests/wallet.rs
+++ b/monero-harness/tests/wallet.rs
@@ -22,11 +22,10 @@ async fn fund_transfer_and_check_tx_key() {
     let alice_wallet = monero.wallet("alice").unwrap();
     let bob_wallet = monero.wallet("bob").unwrap();
 
-    // fund alice
-    monero
-        .init(vec![("alice", fund_alice), ("bob", fund_bob)])
-        .await
-        .unwrap();
+    monero.init_miner().await.unwrap();
+    monero.init_wallet("alice", vec![fund_alice]).await.unwrap();
+    monero.init_wallet("bob", vec![fund_bob]).await.unwrap();
+    monero.start_miner().await.unwrap();
 
     // check alice balance
     let got_alice_balance = alice_wallet.balance().await.unwrap();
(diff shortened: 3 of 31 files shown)


da-kami:
Note: I have not tested concurrent swaps with the same Bob yet :) Will see to deploy it tomorrow and trigger some with the same seed. 
I think we should invest the time to add concurrent swap e2e tests now, but so far I just worked on quickly getting a PoC together. Feedback please :)

swap/src/protocol/alice/event_loop.rs
```diff
@@ -447,7 +451,9 @@ pub struct EventLoopHandle {
 }
 
 impl EventLoopHandle {
-    pub async fn recv_encrypted_signature(&mut self) -> Result<bitcoin::EncryptedSignature> {
+    pub async fn recv_encrypted_signature(
+        &mut self,
+    ) -> Result<(bitcoin::EncryptedSignature, Uuid)> {
```
thomaseizinger, line 456: Why do we need to return the swap-id here? If a swap calls this `recv` function and a signature for the wrong swap shows up, it is already too late.

We should not even be passing on a signature for the wrong swap here.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -459,14 +465,21 @@ impl EventLoopHandle {
...
             .context("Failed to acknowledge receipt of encrypted signature")?;
 
-        Ok(request.tx_redeem_encsig)
+        Ok((request.tx_redeem_encsig, request.swap_id))
     }
 
-    pub async fn send_transfer_proof(&mut self, msg: monero::TransferProof) -> Result<()> {
+    pub async fn send_transfer_proof(
+        &mut self,
+        swap_id: Uuid,
```
thomaseizinger, line 473: Same goes here, I don't think the swap function (which is the user of this method) should have control over this swap-id. Instead, we should associate this at the level where we construct the channels such that whatever gets sent through this channel is automatically associated with the right swap-id on the receiver end.

swap/src/protocol/alice/swap.rs
```diff
@@ -160,13 +163,27 @@ async fn next_state(
...
+                        // TODO: Remove unnecessary swap-id check
+                        // This should never happen, because the encsig is sent through the
+                        // channel by swap_id, however, to be save this was included for now
+                        tracing::warn!("Unexpected encsig for swap {} was meant to be for swap {}. \
+                            This encsig will be ignored.", swap_id, enc_sig_swap_id);
+                        AliceState::XmrLockTransferProofSent {
+                            monero_wallet_restore_blockheight,
+                            transfer_proof,
+                            state3,
+                        }
```
thomaseizinger, line 186: This is exactly what we shouldn't do (as you say in the comment) and you should be able to avoid that if you associate it the messages with the swap-id when you construct the channels.

swap/src/protocol/bob/swap.rs
```diff
@@ -103,14 +107,25 @@ async fn next_state(
...
-
-                        tracing::info!(txid = %transfer_proof.tx_hash(), "Alice locked Monero");
-
-                        BobState::XmrLockProofReceived {
-                            state: state3,
-                            lock_transfer_proof: transfer_proof,
-                            monero_wallet_restore_blockheight
+                        let transfer_proof = transfer_proof?;
+
+                        if transfer_proof.swap_id == swap_id {
```
thomaseizinger, line 112: I think similar to Alice, we shouldn't even be passing on the message through the channel if it is not for the right swap.

swap/src/protocol/alice/event_loop.rs
da-kami: I was thinking about this one and actually changed it at some point to handle the error when receiving.
But the test behaviour did not change.

da-kami:
Note: Will squash those WIP commits later!

swap/src/protocol/bob/event_loop.rs
```diff
@@ -240,11 +254,15 @@ impl EventLoopHandle {
 
     pub async fn send_encrypted_signature(
         &mut self,
+        swap_id: Uuid,
```
thomaseizinger, line 257: This parameter should probably also go away like we did for the others.

da-kami:
Sorry, did not bother to split this in smaller commits any more, but squashed it together 😬

da-kami:
Asserting the balances can be done in a follow-up, I think the change adds enough value already to get merged.
Sending the swap-id in a separate message can also be done in follow-up, we can also decide to drop that and just keep it as part of the first message.

monero-harness/src/lib.rs
```diff
@@ -106,17 +106,34 @@ impl<'c> Monero {
...
+        let monerod = &self.monerod;
+
+        let wallet = self.wallet(name)?;
+        let address = wallet.address().await?.address;
+
+        for amount in amount_in_outputs {
+            if amount > 0 {
+                miner_wallet.transfer(&address, amount).await?;
                 tracing::info!("Funded {} wallet with {}", wallet.name, amount);
                 monerod.client().generate_blocks(10, &miner_address).await?;
```
da-kami, line 124: Note: I tried "optimizing" this by doing all the transfers and then generate 10 blocks (i.e. generate the blocks after the loop), but it had weird side effects on the complete test setup where Balances where suddenly not properly unlocked for assertion.
I am not completely sure why it would behave like that, but for no I would keep it like this.

da-kami: I had this optimization in a separate commit and dropped it :)

da-kami:
I vote for keeping the Monero RPC on debug level in the tests - https://github.com/comit-network/xmr-btc-swap/pull/405/commits/4d7dbe257028f672a61c212d0b09467ee95568e5
Whenever a test fails with a Monero RPC error (`... missing result ...`) I have to switch the debug flag on and run it again. The test output being cluttered is not a big issue.

da-kami:
@thomaseizinger should be all fixed now, please have another look :)
I packed all fixes on top of what was already there into distinct commits. I hope that helps. I did not take the time to split up the feature into multiple commits - I hope that's fine.

da-kami:
> Asserting the balances can be done in a follow-up, I think the change adds enough value already to get merged.
> Sending the swap-id in a separate message can also be done in follow-up, we can also decide to drop that and just keep it as part of the first message.

https://github.com/comit-network/xmr-btc-swap/issues/411

swap/tests/concurrent_bobs_after_xmr_lock_proof_sent.rs
```diff
@@ -0,0 +1,62 @@
...
+        let alice_swap_1 = tokio::spawn(alice::run(alice_swap_1));
+
+        let bob_state_1 = bob_swap_1.await??;
+        assert!(matches!(bob_state_1, BobState::XmrLocked { .. }));
+
+        // make sure Bob's swap one event loop is gone
+        bob_join_handle_1.abort();
+        // sync wallet to ensure that it can be re-used for the second swap, otherwise
+        // the lock tx will fail
+        bob_bitcoin_wallet.sync().await?;
```
thomaseizinger, line 27: Can we do this implicitly as part of `ctx.bob_swap()`?

That would save as a couple of lines in this test and make it easier to understand.

thomaseizinger (approved):
Looks good to me, couple of minor comments.

Can we have a changelog entry that explains what that we fixed a bug here and had to make a breaking change for that?

da-kami:
> Looks good to me, couple of minor comments.
> 
> Can we have a changelog entry that explains what that we fixed a bug here and had to make a breaking change for that?

Riiiight, we have a changelog 👍😅

swap/tests/harness/mod.rs
```diff
@@ -43,13 +43,52 @@ const BITCOIN_TEST_WALLET_NAME: &str = "testwallet";
...
+                    xmr_outputs.push(xmr);
+                    sum_xmr = sum_xmr + xmr;
+                }
+
+                Self {
+                    xmr: sum_xmr,
+                    xmr_outputs,
+                    btc,
+                }
+            }
```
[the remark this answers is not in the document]
da-kami: I found it more readable from the outside when the amount it zero - otherwise one would have to pass an empy vec there.

monero-harness/tests/wallet.rs
```diff
@@ -22,11 +22,10 @@ async fn fund_transfer_and_check_tx_key() {
...
 
-    // fund alice
-    monero
-        .init(vec![("alice", fund_alice), ("bob", fund_bob)])
-        .await
-        .unwrap();
+    monero.init_miner().await.unwrap();
+    monero.init_wallet("alice", vec![fund_alice]).await.unwrap();
+    monero.init_wallet("bob", vec![fund_bob]).await.unwrap();
+    monero.start_miner().await.unwrap();
```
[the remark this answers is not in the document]
da-kami: If we do that we might run into errors when generating blocks with the miner because we bulk-generate blocks when creating alice/bob wallet. It *might* work to just init and start the miner, but I would like to avoid errors that potentially cause side effects as well so I only start the miner *after* bulk generation of blocks i over.

swap/tests/concurrent_bobs_after_xmr_lock_proof_sent.rs
```diff
@@ -0,0 +1,62 @@
...
+        let (bob_swap_2, bob_join_handle_2) = ctx.bob_swap().await;
+        let bob_swap_2 = tokio::spawn(bob::run(bob_swap_2));
+
+        let alice_swap_2 = ctx.alice_next_swap().await;
+        let alice_swap_2 = tokio::spawn(alice::run(alice_swap_2));
+
+        // The second (inner) swap should ALWAYS finish successfully in this
+        // scenario
+
+        let bob_state_2 = bob_swap_2.await??;
```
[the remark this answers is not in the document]
da-kami: I think outer and inner is more confusing than it helps, I removed it and refer to the swaps as `1st` and `2nd`.

da-kami:
@thomaseizinger all changes done from my side - if you want to have another look, planning to `bors r+` this later tonight so it's in by tomorrow :)

swap/tests/harness/mod.rs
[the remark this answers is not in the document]
thomaseizinger: If you name the parameter `split_into_n_utxos` then it would always make sense to pass a value, the default would simply be 1.

Not too important though :)

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Allow multiple concurrent swaps with the same peer on the ASB.
```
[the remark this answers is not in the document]
da-kami: Feel free to change it :)

swap/tests/harness/mod.rs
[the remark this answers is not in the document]
da-kami: For zero it would be 0 and I find it more intuitive to have a `None` then. 
I don't have strong feelings about it though, feel free to change it if it bugs you :)

PR #406, opened 2021-04-09
WIP - concurrent swap tests

da-kami:
See comments in code for details on current open issues.

`concurrent_bobs_after_xmr_lock_proof_sent` behaves as expected - however, it was weird that I had to adapt the channel buffer for the transfer proofs. I don't think that should be needed if I understand the implementation correctly (see TODO comment in Alice's event loop `new_handle`).

`concurrent_bobs_before_xmr_lock_proof_sent` does NOT behave as expected. A "wrong" transfer proof is received by the inner (concurrent) swap (which is expected), but then we never receive the actual transfer proof for that swap.

### DIFF
--- a/swap/src/protocol/alice/event_loop.rs
+++ b/swap/src/protocol/alice/event_loop.rs
@@ -189,7 +189,7 @@ where
                             let _ = self.handle_execution_setup_done(bob_peer_id, swap_id, *state3).await;
                         }
                         SwarmEvent::Behaviour(OutEvent::TransferProofAcknowledged { peer, id }) => {
-                            tracing::trace!(%peer, "Bob acknowledged transfer proof");
+                            tracing::debug!(%peer, "Bob acknowledged transfer proof");
                             if let Some(responder) = self.inflight_transfer_proofs.remove(&id) {
                                 let _ = responder.respond(());
                             }
@@ -258,11 +258,12 @@ where
                                 continue;
                             }
 
+                            // TODO: This seems to internally dial the peer, Alice should not dial Bob - is this wanted behaviour?
                             let id = self.swarm.transfer_proof.send_request(&peer, transfer_proof);
                             self.inflight_transfer_proofs.insert(id, responder);
                         },
-                        Some(Err(_)) => {
-                            tracing::debug!("A swap stopped without sending a transfer proof");
+                        Some(Err(e)) => {
+                            tracing::debug!("A swap stopped without sending a transfer proof: {:#}", e);
                         }
                         None => {
                             unreachable!("stream of transfer proof receivers must never terminate")
@@ -340,7 +341,7 @@ where
             swap_id,
         };
 
-        // TODO: Consider adding separate components for start/rsume of swaps
+        // TODO: Consider adding separate components for start/resume of swaps
 
         // swaps save peer id so we can resume
         match self.db.insert_peer_id(swap_id, bob_peer_id).await {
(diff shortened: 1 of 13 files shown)


swap/src/protocol/alice/event_loop.rs
```diff
@@ -360,7 +361,14 @@ where
...
         // races these futures against a timelock
-        let (transfer_proof_sender, mut transfer_proof_receiver) = bmrng::channel(1);
+
+        // TODO: I had to increase the channel buffer to get
+        //  concurrent_bobs_after_xmr_lock_proof_sent test passing.
+        //  I don't really understand why. In my understanding the channel created here
+        //  is for one specific swap.  Send a transfer proof through this channel
+        //  should still only happen once, as this is a specific channel for a specific
+        //  swap. How is this triggered multiple times?
+        let (transfer_proof_sender, mut transfer_proof_receiver) = bmrng::channel(10);
```
thomaseizinger, line 371: It shouldn't be necessary, you are right. I wonder if this issue goes away if we implement it the way I suggest in https://github.com/comit-network/xmr-btc-swap/pull/405/.

swap/src/protocol/bob/swap.rs
```diff
@@ -120,6 +120,8 @@ async fn next_state(
                                 monero_wallet_restore_blockheight
                             }
                         } else {
+                            // TODO: More elaborate handling for "wrong" transfer proofs, we can store them in the db
```
thomaseizinger, line 123: Should happen in the eventloop, not inside the swap.

swap/tests/bob_refunds_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -10,6 +10,7 @@ use swap::protocol::{alice, bob};
 async fn given_bob_manually_cancels_when_timelock_not_expired_errors() {
     harness::setup_test(SlowCancelConfig, |mut ctx| async move {
         let (bob_swap, bob_join_handle) = ctx.bob_swap().await;
+        let bob_swap_id = bob_swap.swap_id;
```
thomaseizinger, line 13: Given that the swap id is the same for both, I think we don't need to qualify with a role here.

swap/tests/harness/mod.rs
```diff
@@ -58,12 +57,16 @@ struct BobParams {
...
 impl BobParams {
-    pub async fn builder(&self, event_loop_handle: bob::EventLoopHandle) -> Result<bob::Builder> {
+    pub async fn builder(
+        &self,
+        event_loop_handle: bob::EventLoopHandle,
+        swap_id: Uuid,
+    ) -> Result<bob::Builder> {
         let receive_address = self.monero_wallet.get_main_address();
 
         Ok(bob::Builder::new(
```
thomaseizinger, line 67: I think we should get rid of this builder on the next opportunity, I don't see any purpose in that abstraction anymore :)

swap/src/protocol/alice/event_loop.rs
```diff
@@ -258,11 +258,12 @@ where
                                 continue;
                             }
 
+                            // TODO: This seems to internally dial the peer, Alice should not dial Bob - is this wanted behaviour?
```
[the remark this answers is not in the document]
thomaseizinger: I guess there could be the possibility of us losing the connection after we checked it here and before the request can get sent. So instead of buffering it actively, we should maybe track that we tried to send it and if we fail to send it (we can listen for that event), simply try again once we are connected :)

PR #407, opened 2021-04-09
Make Bob re-dial Alice if he ever loses the connection

thomaseizinger:


swap/src/network/redial.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        // lost connection to the configured peer, trigger re-dialling with an
+        // exponential backoff
+        self.backoff.reset();
+        self.sleep = Some(Box::pin(tokio::time::sleep(self.backoff.initial_interval)));
+    }
+
+    fn inject_event(&mut self, _: PeerId, _: ConnectionId, _: Void) {}
+
+    fn poll(
```
da-kami, line 91: ❓ When / How is this behavior polled? 

Is this just polled with Bob's behavior that it is composed into?

da-kami (approved):
Nice solution creating a dedicated behavior for handling re-connecting!
Just one question concerning polling.

swap/src/network/redial.rs
thomaseizinger: > Is this just polled with Bob's behavior that it is composed into?

Yes.

It works the same way `Futures` and hence `async` functions are composed.
I can recommend looking to how the `derive(NetworkBehaviour)` works if this is unclear.

- Macro implementation: https://github.com/libp2p/rust-libp2p/tree/master/swarm-derive
- You can look at the generated code using https://github.com/dtolnay/cargo-expand

rishflab (approved):

PR #422, opened 2021-04-15
Websocket support

da-kami:
This allows us to request quotes via js-libp2p.

PoC used for testing: https://github.com/da-kami/js-libp2p-quote-poc

Thanks for all the input @thomaseizinger 😉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow multiple concurrent swaps with the same peer on the ASB.
   This is a breaking change because the swap ID is now agreed upon between CLI and ASB during swap setup.
   Resuming swaps started prior to this change can result in unexpected behaviour.
+- Quote protocol returns JSON encoded data instead of CBOR.
+  This is a breaking change in the protocol handling, old CLI versions will not be able to process quote requests of ASBs running this version.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   This is a breaking change because the swap ID is now agreed upon between CLI and ASB during swap setup.
   Resuming swaps started prior to this change can result in unexpected behaviour.
 
+### Added
+
+- Websocket support for the ASB.
+  The ASB is now capable to listen on both TCP and Websocket connections.
+  Default websocket listening port is 9940.
+
 ## [0.4.0] - 2021-04-06
 
 ### Changed
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -98,8 +98,11 @@ async fn main() -> Result<()> {
             let kraken_price_updates = kraken::connect()?;
 
             let mut swarm = swarm::alice(&seed)?;
-            Swarm::listen_on(&mut swarm, config.network.listen)
-                .context("Failed to listen network interface")?;
+
+            for listen in config.network.listen {
+                Swarm::listen_on(&mut swarm, listen.clone())
+                    .with_context(|| format!("Failed to listen on network interface {}", listen))?;
+            }
 
             let (event_loop, mut swap_receiver) = EventLoop::new(
                 swarm,
--- a/swap/src/network/transport.rs
+++ b/swap/src/network/transport.rs
@@ -6,11 +6,13 @@ use libp2p::core::{identity, Transport};
 use libp2p::dns::TokioDnsConfig;
 use libp2p::mplex::MplexConfig;
 use libp2p::noise::{self, NoiseConfig, X25519Spec};
+use libp2p::websocket::WsConfig;
 use libp2p::{yamux, PeerId};
 use std::time::Duration;
 
 /// Builds a libp2p transport with the following features:
 /// - TcpConnection
+/// - WebSocketConnection
 /// - DNS name resolution
 /// - authentication via noise
 /// - multiplexing via yamux or mplex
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   This is a breaking change because the swap ID is now agreed upon between CLI and ASB during swap setup.
   Resuming swaps started prior to this change can result in unexpected behaviour.
+- Quote protocol returns JSON encoded data instead of CBOR.
+  This is a breaking change in the protocol handling, old CLI versions will not be able to process quote requests of ASBs running this version.
+
+### Added
+
+- Websocket support for the ASB.
+  The ASB is not capable to listen on both TCP and Websocket connections.
+  Default websocket listening port is 9940, default TCP remains unchanged at 9939.
```
thomaseizinger, line 22: These two lines are sort of contradicting each other, don't you think? Just mentioning the new port is good enough. I think it is common understanding that you can only have one transport protocol per port.

swap/src/bin/asb.rs
```diff
@@ -98,8 +98,15 @@ async fn main() -> Result<()> {
...
-            Swarm::listen_on(&mut swarm, config.network.listen)
-                .context("Failed to listen network interface")?;
+
+            for listen in config.network.listen {
+                Swarm::listen_on(&mut swarm, listen.clone()).context(format!(
+                    "Failed to listen on network interface {}",
+                    listen.clone()
+                ))?;
+
+                tracing::info!("Listening for connections on {}", listen);
```
thomaseizinger, line 108: This is not quite true. If you want to emit such a log, then we should do it once we receive the `NewListenAddr` event in the eventloop which I think we already do :)

thomaseizinger (approved):
Having first and last commit squashed together would be nice!

Otherwise, LGTM!

rishflab (approved):

CHANGELOG.md
da-kami: That's because it should not be `not` but `now` 😅

swap/src/bin/asb.rs
da-kami: We don't emit any logs when we start listening, we only emit when there is a new connection. Happy to move it to `NewListenAddr` events.

da-kami: Well, that is interesting - we do have an info log on `NewListenAddr`, but I never see it in the logs.

da-kami: Thanks for the hint! I added this log before the rebase which brought in the listen logs on `NewListenAddr` - will remove them from main.

da-kami: Interesting side effect, the logs claim that we are listening on both ports for both protocols:

```
Apr 16 10:18:01.571  INFO Listening on /ip4/127.0.0.1/tcp/9940/ws
Apr 16 10:18:01.571  INFO Listening on /ip4/127.0.0.1/tcp/9939
Apr 16 10:18:01.571  INFO Listening on /ip4/192.168.1.2/tcp/9940/ws
Apr 16 10:18:01.571  INFO Listening on /ip4/192.168.1.2/tcp/9939
```

Is there a way to listen to the effective listening addresses, not just "new" ?

da-kami: Given that I had to use `or_transport` I don't think the application is actually listening for both protocols on both ports...?

swap/src/bin/asb.rs
```diff
@@ -98,8 +98,13 @@ async fn main() -> Result<()> {
...
 
             let mut swarm = swarm::alice(&seed)?;
-            Swarm::listen_on(&mut swarm, config.network.listen)
-                .context("Failed to listen network interface")?;
+
+            for listen in config.network.listen {
+                Swarm::listen_on(&mut swarm, listen.clone()).context(format!(
+                    "Failed to listen on network interface {}",
+                    listen.clone()
+                ))?;
```
thomaseizinger, line 106: Minor nit, but when using `format!`, you should always use `with_context` to avoid the allocation that is happening as part of `format!` in case there is no error.

Why?

Before you call a function, all its arguments need to be evaluated. Hence, with `.context` and `format!`, the `format!` needs to evaluated before `.context` can even be executed.
However, if there is no error (i.e. `listen_on` succeeds), then the argument passed to `.context` will never be needed because the internal control flow doesn't need this error message if the whole thing succeeds.

Using `with_context` avoids this because all that needs to be evaluated is the closure that is being passed to `with_context`. This closure acts as a layer of indirection that will only be called when needed (read: if there is an error) and hence also the allocation within `format!` only happens when really necessary.

Additionally, I am surprised you need to `.clone()` it again within the `format!` macro.

swap/src/network/transport.rs
```diff
@@ -21,9 +23,12 @@ pub fn build(id_keys: &identity::Keypair) -> Result<SwapTransport> {
...
 
     let tcp = TokioTcpConfig::new().nodelay(true);
-    let dns = TokioDnsConfig::system(tcp)?;
+    let dns = TokioDnsConfig::system(tcp.clone())?;
 
-    let transport = dns
+    let websocket = WsConfig::new(dns);
+
+    let transport = websocket
+        .or_transport(tcp)
```
thomaseizinger, line 31: Sorry, I didn't see this in my initial review but I think we might actually want to clone the `dns` transport instead of the `tcp` one. Otherwise we lose DNS support outside of the Websocket transport.

thomaseizinger: Sorry to only mention this now, didn't occur to me in my initial review!

thomaseizinger (approved):
Two minor things, otherwise good :)

PR #434, opened 2021-04-20
Introduce monero-wallet crate

thomaseizinger:
This PR:

1. ~Introduce a crate for the epee binary serialization as a serde format~: Released here: https://github.com/comit-network/monero-epee-bin-serde
2. Extends the MoneroRPC client with two binary calls
3. Introduces a `monero-wallet` crate that for now just provides functionality for choosing random key offsets. Together with the the ability to produce bulletproofs and ring signatures, this should be enough for signing Monero transactions locally.

(1) and (2) are a prerequisite for (3).

### DIFF
--- a/monero-rpc/src/monerod.rs
+++ b/monero-rpc/src/monerod.rs
@@ -1,5 +1,6 @@
 use anyhow::{Context, Result};
-use serde::Deserialize;
+use monero::{cryptonote::hash::Hash, util::ringct, PublicKey};
+use serde::{de::DeserializeOwned, Deserialize, Serialize, Serializer};
 
 #[jsonrpc_client::api(version = "2.0")]
 pub trait MonerodRpc {
@@ -7,6 +8,7 @@ pub trait MonerodRpc {
         -> GenerateBlocks;
     async fn get_block_header_by_height(&self, height: u32) -> BlockHeader;
     async fn get_block_count(&self) -> BlockCount;
+    async fn get_block(&self, height: u32) -> GetBlockResponse;
 }
 
 #[jsonrpc_client::implement(MonerodRpc)]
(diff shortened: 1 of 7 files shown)


swap/src/bin/swap.rs
```diff
@@ -103,17 +103,16 @@ async fn main() -> Result<()> {
...
+                bitcoin_wallet,
                 Arc::new(monero_wallet),
                 env_config,
-                event_loop_handle,
+                swap_id,
                 receive_monero_address,
-            )
-            .with_init_params(send_bitcoin)
-            .build()?;
+            );
```
thomaseizinger, line 115: TODO: Keep the order of the parameters the same to reduce noise.

monero-epee-bin-serde/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use crate::de::Deserializer;
+use crate::ser::Serializer;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use std::io::Read;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+const MAGIC: &[u8] = b"\x01\x11\x01\x01\x01\x01\x02\x01\x01";
```
rishflab, line 16: O.o What does this do? Maybe a comment would be good so you don't have to read `from_bytes`?

monero-rpc/src/monerod.rs
```diff
@@ -1,46 +1,92 @@
...
+use serde::{de::DeserializeOwned, Deserialize, Serialize, Serializer};
 
 #[jsonrpc_client::api(version = "2.0")]
 pub trait MonerodRpc {
     async fn generateblocks(&self, amount_of_blocks: u32, wallet_address: String)
         -> GenerateBlocks;
     async fn get_block_header_by_height(&self, height: u32) -> BlockHeader;
     async fn get_block_count(&self) -> BlockCount;
+    async fn get_block(&self, height: u32) -> GetBlockResponse;
+    async fn on_get_block_hash(&self, height: [u32; 1]) -> String;
```
rishflab, line 12: I cant see this being used anywhere, What does it do?

monero-rpc/src/monerod.rs
```diff
@@ -1,46 +1,92 @@
...
+use monero::{cryptonote::hash::Hash, util::ringct, PublicKey};
+use serde::{de::DeserializeOwned, Deserialize, Serialize, Serializer};
 
 #[jsonrpc_client::api(version = "2.0")]
 pub trait MonerodRpc {
     async fn generateblocks(&self, amount_of_blocks: u32, wallet_address: String)
         -> GenerateBlocks;
     async fn get_block_header_by_height(&self, height: u32) -> BlockHeader;
     async fn get_block_count(&self) -> BlockCount;
+    async fn get_block(&self, height: u32) -> GetBlockResponse;
```
rishflab, line 11: I cant see this being used anywhere?

monero-wallet/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+            .unwrap_or_else(|| block.blob.miner_tx.hash());
+
+        let indices = self.client.get_o_indexes(tx_hash).await?;
+
+        let last_index = indices
+            .o_indexes
+            .into_iter()
+            .max()
+            .context("Expected at least one output index")?;
+        let oldest_index = last_index - (last_index / 100) * 40; // oldest index must be within last 40% TODO: CONFIRM THIS
```
rishflab, line 39: Is this a protocol constraint?

monero-rpc/src/monerod.rs
thomaseizinger: You are right, turned out to not be needed.

monero-rpc/src/monerod.rs
thomaseizinger: It is used in `monero-wallet`.

monero-wallet/src/lib.rs
thomaseizinger: I am not fully sure yet. It is blindly copied based on the first caveat here: https://github.com/kayabaNerve/cryptonote-library#caveats

monero-rpc/src/monerod.rs
```diff
@@ -1,46 +1,91 @@
...
+    fn new(host: String, port: u16) -> Result<Self> {
         Ok(Self {
             inner: reqwest::ClientBuilder::new()
                 .connection_verbose(true)
                 .build()?,
-            base_url: format!("http://127.0.0.1:{}/json_rpc", port)
+            base_url: format!("http://{}:{}/json_rpc", host, port)
                 .parse()
                 .context("url is well formed")?,
+            get_o_indexes_bin_url: format!("http://{}:{}/get_o_indexes.bin", host, port)
```
luckysori, line 37: :open_mouth: So it's a different base URL for these particular calls.

luckysori (approved):
Amazing work.

monero-rpc/src/monerod.rs
thomaseizinger: Yes, they are in fact "normal" HTTP calls (except that they don't care about the method you pass in) with a binary payload.

PR #436, opened 2021-04-21
use '--seller-peer-id'

SamouraiDev:


da-kami (approved):
Those parameters just had too many renames 😅 - thanks for taking the time to correct it!

PR #438, opened 2021-04-22
Add support for swapping through Tor.

bonomat:
Resolves: #447 

This PR does a few things.
* It adds a TorTransport which either dials through Tor's socks5 proxy or via clearnet.
* It enables ASB to register hidden services for each network it is listening on. We assume that we only care about different ports and re-use the same onion-address for all of them. The ASB requires to have access to Tor's control port.
* It adds support to dial through a local Tor socks5 proxy. We assume that Tor is always available on localhost.  Swap cli only requires Tor to be running so that it can send messages via Tor's socks5 proxy.
* It adds a new e2e test which swaps through Tor. For this we assume that Tor is currently running on localhost. All other tests are running via clear net.

Note: It is expected that the new test will fail on CI because we do not have Tor running. That's why it wasn't added yet. I'll play around in my own fork to not use precious CI resources :) 
The rest is ready for review. 

// edit:

Swapping through Tor on one machine seems to be working: 

Asb: 
```
cargo run --bin asb -- --config "/Users/bonomat/Library/Application Support/xmr-btc-swap/asb/config.toml" start
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/asb --config '/Users/bonomat/Library/Application Support/xmr-btc-swap/asb/config.toml' start`
Apr 22 11:14:05.090  INFO Initialized tracing with level: debug
Apr 22 11:14:05.092  INFO Using config file at default path: /Users/bonomat/Library/Application Support/xmr-btc-swap/asb/config.toml
Apr 22 11:14:05.094  INFO Database and Seed will be stored in directory: /Users/bonomat/Library/Application Support/xmr-btc-swap/asb
Apr 22 11:14:05.094 DEBUG Opening database at /Users/bonomat/Library/Application Support/xmr-btc-swap/asb/database
Apr 22 11:14:05.226 DEBUG Reading in seed from /Users/bonomat/Library/Application Support/xmr-btc-swap/asb/seed.pem
Apr 22 11:14:06.903  INFO Tor found. Setting up hidden service. 
Apr 22 11:14:08.805  INFO /onion3/jugnrrdp7imu5tqphvlrctudguq3sapmjideu6d63i5y7qqtzzwvf4id:9939
Apr 22 11:14:08.805  INFO /onion3/jugnrrdp7imu5tqphvlrctudguq3sapmjideu6d63i5y7qqtzzwvf4id:9940
Apr 22 11:14:13.431 DEBUG Opened Monero wallet asb-wallet
Apr 22 11:14:13.433  INFO Bitcoin balance: 0.00000000 BTC
Apr 22 11:14:13.433  WARN The Monero balance is 0, make sure to deposit funds at: 5A3iBfDbhGfUUL5WKFcdY5JK7oLcMXYmD9VnUjxmBgFHXswtENMjFsHUDeeCWVvRYaNRCAJDRS7jY85iyNt7s3syVNJtwLd
Apr 22 11:14:13.436 DEBUG Trying to listen on: /ip4/0.0.0.0/tcp/9939
Apr 22 11:14:13.437 DEBUG Trying to listen on: /ip4/0.0.0.0/tcp/9940
Apr 22 11:14:13.437  INFO Our peer id is 12D3KooWNqWpoLUp6YSoUYTinwJFh92wzggeanLgSutmZbZYG9cN
Apr 22 11:14:13.438  INFO Listening on /ip4/127.0.0.1/tcp/9940/ws
Apr 22 11:14:13.438  INFO Listening on /ip4/127.0.0.1/tcp/9939
Apr 22 11:14:13.438  INFO Listening on /ip4/192.168.1.63/tcp/9940/ws
Apr 22 11:14:13.438  INFO Listening on /ip4/192.168.1.63/tcp/9939
Apr 22 11:14:14.275 DEBUG Connected to Kraken websocket API
Apr 22 11:14:14.462 DEBUG Subscribed to updates for ticker
Apr 22 11:23:11.833 DEBUG New connection established peer=12D3KooWHBMKm8Wnq4WD7ehVdSHaf4ccjdzR97UTFtMGT6LBmRev address=/ip4/127.0.0.1/tcp/63655
Apr 22 11:23:25.700  INFO swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: Current state: started
Apr 22 11:23:25.700  INFO swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: Waiting for 1 confirmation of Bitcoin transaction txid=537070d7a6f75e74caa65cebbbf0997b50914fd86b7a752f0d74eb0c5701291f
Apr 22 11:23:33.732 DEBUG Transaction is in mempool txid=537070d7a6f75e74caa65cebbbf0997b50914fd86b7a752f0d74eb0c5701291f
Apr 22 11:42:49.668 DEBUG Transaction is confirmed with 1 blocks txid=537070d7a6f75e74caa65cebbbf0997b50914fd86b7a752f0d74eb0c5701291f
Apr 22 11:42:49.671  INFO swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: Bitcoin tx has 1 out of 1 confirmation txid=537070d7a6f75e74caa65cebbbf0997b50914fd86b7a752f0d74eb0c5701291f
Apr 22 11:42:49.703  INFO swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: Current state: btc is locked
Apr 22 11:43:02.670 DEBUG swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: sent transfer of 0.695214974374 XMR to 05d77c136ad84eff1184a14971f8ec7f5884bb25b646a6c1eb95e0c26782f0a5 in c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
Apr 22 11:43:02.740  INFO swap{id=c4f51a28-b2fb-4885-9f7a-5d1852de5f0d}: Current state: xmr lock transaction sent

...
```

Swap:
```
 cargo run --bin swap buy-xmr --seller-peer-id 12D3KooWNqWpoLUp6YSoUYTinwJFh92wzggeanLgSutmZbZYG9cN --seller-addr /onion3/jugnrrdp7imu5tqphvlrctudguq3sapmjideu6d63i5y7qqtzzwvf4id:9939 --receive-address 5A3iBfDbhGfUUL5WKFcdY5JK7oLcMXYmD9VnUjxmBgFHXswtENMjFsHUDeeCWVvRYaNRCAJDRS7jY85iyNt7s3syVNJtwLd 
    Finished dev [unoptimized + debuginfo] target(s) in 0.37s
     Running `target/debug/swap buy-xmr --seller-peer-id 12D3KooWNqWpoLUp6YSoUYTinwJFh92wzggeanLgSutmZbZYG9cN --seller-addr '/onion3/jugnrrdp7imu5tqphvlrctudguq3sapmjideu6d63i5y7qqtzzwvf4id:9939' --receive-address 5A3iBfDbhGfUUL5WKFcdY5JK7oLcMXYmD9VnUjxmBgFHXswtENMjFsHUDeeCWVvRYaNRCAJDRS7jY85iyNt7s3syVNJtwLd`
 Connecting to Tor proxy ...
 Connection established
 Connected to Alice at /onion3/jugnrrdp7imu5tqphvlrctudguq3sapmjideu6d63i5y7qqtzzwvf4id:9939/p2p/12D3KooWNqWpoLUp6YSoUYTinwJFh92wzggeanLgSutmZbZYG9cN
 Received quote: 1 XMR ~ 0.00719202 BTC
 Found 0.01559763 BTC in wallet
 Swapping 0.00500000 BTC with 0.00000610 BTC fees
 Spot price for 0.00500000 BTC is 0.695214974374 XMR
 Published Bitcoin lock transaction txid=537070d7a6f75e74caa65cebbbf0997b50914fd86b7a752f0d74eb0c5701291f
 Waiting for Alice to lock Monero
 Alice locked Monero txid=c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
 Waiting for 10 confirmations of Monero transaction txid=c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
 Monero lock tx has 1 out of 10 confirmations txid=c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
 Monero lock tx has 2 out of 10 confirmations txid=c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
 Monero lock tx has 3 out of 10 confirmations txid=c00f4543d9c9831fa892a2180b4dabbc33eff1fcd6f58ac24495dcf1450bbab3
...

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Websocket support for the ASB.
   The ASB is now capable to listen on both TCP and Websocket connections.
   Default websocket listening port is 9940.
+- Tor support as an optional feature.
+  If ASB detects that Tor's control port is open, an hidden service is created for
+  network it is listening on (currently 2).
+  The Tor control port as well as Tor socks5 proxy port is configurable.
 
 ## [0.4.0] - 2021-04-06
 
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -22,6 +22,7 @@ bmrng = "0.5"
 config = { version = "0.11", default-features = false, features = ["toml"] }
 conquer-once = "0.3"
 curve25519-dalek = "3"
+data-encoding = "2.3"
 dialoguer = "0.8"
 directories-next = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", features = ["libsecp_compat", "serde"] }
@@ -36,7 +37,7 @@ pem = "0.8"
 prettytable-rs = "0.8"
 rand = "0.7"
 rand_chacha = "0.2"
-reqwest = { version = "0.11", features = ["rustls-tls", "stream"], default-features = false }
+reqwest = { version = "0.11", features = ["rustls-tls", "stream", "socks"], default-features = false }
 rust_decimal = "1"
 serde = { version = "1", features = ["derive"] }
 serde_cbor = "0.11"
@@ -48,10 +49,12 @@ structopt = "0.3"
 strum = { version = "0.20", features = ["derive"] }
 thiserror = "1"
 time = "0.2"
-tokio = { version = "1", features = ["rt-multi-thread", "time", "macros", "sync", "process", "fs"] }
+tokio = { version = "1", features = ["rt-multi-thread", "time", "macros", "sync", "process", "fs", "net"] }
+tokio-socks = "0.5"
 tokio-tungstenite = { version = "0.14", features = [ "rustls-tls" ] }
 tokio-util = { version = "0.6", features = ["io"] }
 toml = "0.5"
+torut = "0.1"
 tracing = { version = "0.1", features = ["attributes"] }
 tracing-appender = "0.1"
 tracing-futures = { version = "0.2", features = ["std-future", "futures-03"] }
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -54,7 +54,7 @@ tokio-socks = "0.5"
 tokio-tungstenite = { version = "0.14", features = [ "rustls-tls" ] }
 tokio-util = { version = "0.6", features = ["io"] }
 toml = "0.5"
-torut = "0.1"
+torut = { version = "0.1", default-features = false, features = ["v3", "control"] }
 tracing = { version = "0.1", features = ["attributes"] }
 tracing-appender = "0.1"
 tracing-futures = { version = "0.2", features = ["std-future", "futures-03"] }
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -26,6 +26,7 @@ data-encoding = "2.3"
 dialoguer = "0.8"
 directories-next = "2"
 ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", features = ["libsecp_compat", "serde"] }
+ed25519-dalek = "1"
 futures = { version = "0.3", default-features = false }
 itertools = "0.10"
 libp2p = { version = "0.37", default-features = false, features = ["tcp-tokio", "yamux", "mplex", "dns-tokio", "noise", "request-response", "websocket"] }
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -1,4 +1,5 @@
 use crate::fs::{default_data_dir, ensure_directory_exists};
+use crate::tor::{DEFAULT_CONTROL_PORT, DEFAULT_SOCKS5_PORT};
 use anyhow::{Context, Result};
 use config::ConfigError;
 use dialoguer::theme::ColorfulTheme;
@@ -22,6 +23,7 @@ pub struct Config {
     pub network: Network,
     pub bitcoin: Bitcoin,
     pub monero: Monero,
+    pub tor: TorConf,
 }
 
 impl Config {
@@ -59,6 +61,22 @@ pub struct Bitcoin {
 #[serde(deny_unknown_fields)]
 pub struct Monero {
     pub wallet_rpc_url: Url,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct TorConf {
+    pub control_port: u16,
+    pub socks5_port: u16,
+}
+
+impl Default for TorConf {
+    fn default() -> Self {
+        Self {
+            control_port: DEFAULT_CONTROL_PORT,
+            socks5_port: DEFAULT_SOCKS5_PORT,
+        }
+    }
 }
 
 #[derive(thiserror::Error, Debug, Clone, Copy)]
@@ -133,6 +151,17 @@ pub fn query_user_for_initial_testnet_config() -> Result<Config> {
         .default(DEFAULT_MONERO_WALLET_RPC_TESTNET_URL.to_owned())
         .interact_text()?;
     let monero_wallet_rpc_url = monero_wallet_rpc_url.as_str().parse()?;
+
+    let tor_control_port = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Enter Tor control port or hit enter to use default. If Tor is not running on your machine, no hidden service will be created.")
+        .default(DEFAULT_CONTROL_PORT.to_owned())
+        .interact_text()?;
+
+    let tor_socks5_port = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Enter Tor socks5 port or hit enter to use default")
+        .default(DEFAULT_SOCKS5_PORT.to_owned())
+        .interact_text()?;
+
     println!();
 
     Ok(Config {
@@ -143,6 +172,10 @@ pub fn query_user_for_initial_testnet_config() -> Result<Config> {
         bitcoin: Bitcoin { electrum_rpc_url },
         monero: Monero {
             wallet_rpc_url: monero_wallet_rpc_url,
+        },
+        tor: TorConf {
+            control_port: tor_control_port,
+            socks5_port: tor_socks5_port,
         },
     })
 }
@@ -175,6 +208,7 @@ mod tests {
             monero: Monero {
                 wallet_rpc_url: Url::from_str(DEFAULT_MONERO_WALLET_RPC_TESTNET_URL).unwrap(),
             },
+            tor: Default::default(),
         };
 
         initial_setup(config_path.clone(), || Ok(expected.clone())).unwrap();
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -13,8 +13,11 @@
 #![allow(non_snake_case)]
 
 use anyhow::{Context, Result};
+use libp2p::core::multiaddr::Protocol;
+use libp2p::core::Multiaddr;
 use libp2p::Swarm;
 use prettytable::{row, Table};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::sync::Arc;
 use structopt::StructOpt;
 use swap::asb::command::{Arguments, Command};
@@ -29,7 +32,8 @@ use swap::network::swarm;
 use swap::protocol::alice::event_loop::KrakenRate;
 use swap::protocol::alice::{run, EventLoop};
 use swap::seed::Seed;
-use swap::{asb, bitcoin, env, kraken, monero};
+use swap::tor::AuthenticatedClient;
+use swap::{asb, bitcoin, env, kraken, monero, tor};
 use tracing::{info, warn};
 use tracing_subscriber::filter::LevelFilter;
 
@@ -96,6 +100,23 @@ async fn main() -> Result<()> {
             }
 
             let kraken_price_updates = kraken::connect()?;
+
+            // setup Tor hidden services
+            let tor_client =
+                tor::Client::new(config.tor.socks5_port).with_control_port(config.tor.control_port);
+            let _ac = match tor_client.assert_tor_running().await {
+                Ok(_) => {
+                    tracing::info!("Tor found. Setting up hidden service. ");
+                    let ac =
+                        register_tor_services(config.network.clone().listen, tor_client, &seed)
+                            .await?;
+                    Some(ac)
+                }
+                Err(_) => {
+                    tracing::warn!("Tor not found. Running on clear net. ");
+                    None
+                }
+            };
 
             let mut swarm = swarm::alice(&seed)?;
 
@@ -211,4 +232,46 @@ async fn init_monero_wallet(
     .await?;
 
     Ok(wallet)
+}
+
+/// Registers a hidden service for each network.
+/// Note: Once ac goes out of scope, the services will be de-registered.
+async fn register_tor_services(
+    networks: Vec<Multiaddr>,
+    tor_client: tor::Client,
+    seed: &Seed,
+) -> Result<AuthenticatedClient> {
+    let mut ac = tor_client.into_authenticated_client().await?;
+
+    let hidden_services_details = networks
+        .iter()
+        .flat_map(|network| {
+            network.iter().map(|protocol| match protocol {
+                Protocol::Tcp(port) => Some((
+                    port,
+                    SocketAddr::new(IpAddr::from(Ipv4Addr::new(127, 0, 0, 1)), port),
+                )),
+                _ => {
+                    // We only care for Tcp for now.
+                    None
+                }
+            })
+        })
+        .filter_map(|details| details)
+        .collect::<Vec<_>>();
+
+    let key = seed.derive_torv3_key();
+
+    ac.add_services(&hidden_services_details, &key).await?;
+
+    let onion_address = key
+        .public()
+        .get_onion_address()
+        .get_address_without_dot_onion();
+
+    hidden_services_details.iter().for_each(|(port, _)| {
+        tracing::info!("/onion3/{}:{}", onion_address, port);
+    });
+
+    Ok(ac)
 }
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -51,6 +51,7 @@ async fn main() -> Result<()> {
                     monero_daemon_host,
                 },
             electrum_rpc_url,
+            tor_socks5_port,
         } => {
             let swap_id = Uuid::new_v4();
 
@@ -76,7 +77,7 @@ async fn main() -> Result<()> {
                 init_monero_wallet(data_dir, monero_daemon_host, env_config).await?;
             let bitcoin_wallet = Arc::new(bitcoin_wallet);
 
-            let mut swarm = swarm::bob(&seed, alice_peer_id)?;
+            let mut swarm = swarm::bob(&seed, alice_peer_id, tor_socks5_port).await?;
             swarm
                 .behaviour_mut()
                 .add_address(alice_peer_id, alice_multiaddr);
@@ -152,6 +153,7 @@ async fn main() -> Result<()> {
                     monero_daemon_host,
                 },
             electrum_rpc_url,
+            tor_socks5_port,
         } => {
             let data_dir = data.0;
             cli::tracing::init(debug, data_dir.join("logs"), swap_id)?;
@@ -172,7 +174,8 @@ async fn main() -> Result<()> {
             let bitcoin_wallet = Arc::new(bitcoin_wallet);
 
             let alice_peer_id = db.get_peer_id(swap_id)?;
-            let mut swarm = swarm::bob(&seed, alice_peer_id)?;
+
+            let mut swarm = swarm::bob(&seed, alice_peer_id, tor_socks5_port).await?;
             let bob_peer_id = swarm.local_peer_id();
             tracing::debug!("Our peer-id: {}", bob_peer_id);
             swarm
--- a/swap/src/cli/command.rs
+++ b/swap/src/cli/command.rs
@@ -13,6 +13,8 @@ pub const DEFAULT_STAGENET_MONERO_DAEMON_HOST: &str = "monero-stagenet.exan.tech
 pub const DEFAULT_ELECTRUM_HTTP_URL: &str = "https://blockstream.info/testnet/api/";
 const DEFAULT_ELECTRUM_RPC_URL: &str = "ssl://electrum.blockstream.info:60002";
 
+const DEFAULT_TOR_SOCKS5_PORT: &str = "9050";
+
 #[derive(structopt::StructOpt, Debug)]
 #[structopt(name = "swap", about = "CLI for swapping BTC for XMR", author)]
 pub struct Arguments {
@@ -48,6 +50,9 @@ pub enum Command {
 
         #[structopt(flatten)]
         monero_params: MoneroParams,
+
+        #[structopt(long = "tor-socks5-port", help = "Your local Tor socks5 proxy port", default_value = DEFAULT_TOR_SOCKS5_PORT)]
+        tor_socks5_port: u16,
     },
     /// Show a list of past ongoing and completed swaps
     History,
@@ -70,6 +75,9 @@ pub enum Command {
 
         #[structopt(flatten)]
         monero_params: MoneroParams,
+
+        #[structopt(long = "tor-socks5-port", help = "Your local Tor socks5 proxy port", default_value = DEFAULT_TOR_SOCKS5_PORT)]
+        tor_socks5_port: u16,
     },
     /// Try to cancel an ongoing swap (expert users only)
     Cancel {
--- a/swap/src/network/tor_transport.rs
+++ b/swap/src/network/tor_transport.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use data_encoding::BASE32;
 use futures::future::Ready;
 use futures::prelude::*;
@@ -6,7 +7,6 @@ use libp2p::core::transport::TransportError;
 use libp2p::core::Transport;
 use libp2p::tcp::tokio::{Tcp, TcpStream};
 use libp2p::tcp::{GenTcpConfig, TcpListenStream, TokioTcpConfig};
-use std::cmp::Ordering;
 use std::io;
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
 use std::pin::Pin;
@@ -55,13 +55,14 @@ impl Transport for TorTcpConfig {
         }
 
         match to_address_string(addr.clone()) {
-            Some(tor_address_string) => {
+            Ok(tor_address_string) => {
                 Ok(Box::pin(do_tor_dial(self.socks_port, tor_address_string)))
             }
-            _ => {
+            Err(e) => {
                 tracing::warn!(
-                    "Address {} could not be formatted. Dialling via clear net",
-                    addr
+                    "Address {} could not be formatted. Dialling via clear net. Details: {}",
+                    addr,
+                    e
                 );
                 self.inner.dial(addr)
             }
@@ -76,68 +77,41 @@ impl Transport for TorTcpConfig {
 /// Tor expects an address format of ADDR:PORT.
 /// This helper function tries to convert the provided multi-address into this
 /// format. None is returned if an unsupported protocol was provided.
-fn to_address_string(multi: Multiaddr) -> Option<String> {
-    let components = multi.iter();
-    for protocol in components {
-        if let Protocol::Onion3(addr) = protocol {
-            return Some(format!(
+fn to_address_string(multi: Multiaddr) -> anyhow::Result<String> {
+    let mut protocols = multi.iter();
+    let address_string = match protocols.next() {
+        // if it is an Onion address, we have all we need and can return
+        Some(Protocol::Onion3(addr)) => {
+            return Ok(format!(
                 "{}.onion:{}",
                 BASE32.encode(addr.hash()).to_lowercase(),
                 addr.port()
-            ));
+            ))
         }
+        // Deal with non-onion addresses
+        Some(Protocol::Ip4(addr)) => Some(format!("{}", addr)),
+        Some(Protocol::Ip6(addr)) => Some(format!("{}", addr)),
+        Some(Protocol::Dns(addr)) => Some(format!("{}", addr)),
+        Some(Protocol::Dns4(addr)) => Some(format!("{}", addr)),
+        _ => None,
     }
-
-    // Deal with non-onion addresses
-    let protocols = multi.iter().collect::<Vec<_>>();
-    let address_string = protocols
-        .iter()
-        .filter_map(|protocol| match protocol {
-            Protocol::Ip4(addr) => Some(format!("{}", addr)),
-            Protocol::Ip6(addr) => Some(format!("{}", addr)),
-            Protocol::Dns(addr) => Some(format!("{}", addr)),
-            Protocol::Dns4(addr) => Some(format!("{}", addr)),
-            _ => None,
-        })
-        .collect::<Vec<_>>();
-    if address_string.is_empty() {
-        tracing::warn!(
+    .ok_or_else(|| {
+        anyhow!(
             "Could not format address {}. Please consider reporting this issue. ",
             multi
-        );
-        return None;
-    }
+        )
+    })?;
 
-    let address_string = address_string
-        .get(0)
-        .expect("Valid multiaddr consist out of max 1 address")
-        .clone();
+    let port_string = match protocols.next() {
+        Some(Protocol::Tcp(port)) => Some(format!("{}", port)),
+        Some(Protocol::Udp(port)) => Some(format!("{}", port)),
+        _ => None,
+    };
 
-    // check for port
-    let port = protocols
-        .iter()
-        .filter_map(|protocol| match protocol {
-            Protocol::Tcp(port) => Some(format!("{}", port)),
-            Protocol::Udp(port) => Some(format!("{}", port)),
-            _ => None,
-        })
-        .collect::<Vec<_>>();
-
-    match port.len().cmp(&1) {
-        Ordering::Greater => {
-            tracing::warn!(
-                "Did not expect more than 1 port in address {}. Please consider reporting this issue.",
-                multi
-            );
-            Some(address_string)
-        }
-        Ordering::Less => Some(address_string),
-        Ordering::Equal => Some(format!(
-            "{}:{}",
-            address_string,
-            port.get(0)
-                .expect("Already verified the length of the vec.")
-        )),
+    if let Some(port) = port_string {
+        Ok(format!("{}:{}", address_string, port))
+    } else {
+        Ok(address_string)
     }
 }
 
@@ -219,7 +193,7 @@ pub mod test {
     #[test]
     fn dnsaddr_to_address_string_should_be_none() {
         let address = "/dnsaddr/randomdomain.com";
-        let address_string = to_address_string(address.parse().unwrap());
+        let address_string = to_address_string(address.parse().unwrap()).ok();
         assert_eq!(address_string, None);
     }
 }
--- a/swap/src/network/tor_transport.rs
+++ b/swap/src/network/tor_transport.rs
@@ -1,0 +1,238 @@
+use data_encoding::BASE32;
+use futures::future::Ready;
+use futures::prelude::*;
+use libp2p::core::multiaddr::{Multiaddr, Protocol};
+use libp2p::core::transport::TransportError;
+use libp2p::core::Transport;
+use libp2p::tcp::tokio::{Tcp, TcpStream};
+use libp2p::tcp::{GenTcpConfig, TcpListenStream, TokioTcpConfig};
+use std::cmp::Ordering;
+use std::io;
+use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::pin::Pin;
+use tokio_socks::tcp::Socks5Stream;
+use tokio_socks::IntoTargetAddr;
+
+/// Represents the configuration for a TCP/IP transport capability for libp2p.
+#[derive(Clone)]
+pub struct TorTcpConfig {
+    inner: GenTcpConfig<Tcp>,
+    /// Tor SOCKS5 proxy port number.
+    socks_port: u16,
+}
+
+impl TorTcpConfig {
+    pub fn new(tcp: TokioTcpConfig, socks_port: u16) -> Self {
+        Self {
+            inner: tcp,
+            socks_port,
+        }
+    }
+}
+
+impl Transport for TorTcpConfig {
+    type Output = TcpStream;
+    type Error = io::Error;
+    type Listener = TcpListenStream<Tcp>;
+    type ListenerUpgrade = Ready<Result<Self::Output, Self::Error>>;
+    #[allow(clippy::type_complexity)]
+    type Dial = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
+
+    fn listen_on(self, addr: Multiaddr) -> Result<Self::Listener, TransportError<Self::Error>> {
+        self.inner.listen_on(addr)
+    }
+
+    // dials via Tor's socks5 proxy if configured and if the provided address is an
+    // onion address. or it falls back to Tcp dialling
+    fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
+        async fn do_tor_dial(socks_port: u16, dest: String) -> Result<TcpStream, io::Error> {
+            tracing::trace!("Connecting through Tor proxy to address: {}", dest);
+            let stream = connect_to_socks_proxy(dest, socks_port)
+                .await
+                .map_err(|e| io::Error::new(io::ErrorKind::ConnectionRefused, e))?;
+            tracing::trace!("Connection through Tor established");
+            Ok(stream)
+        }
+
+        match to_address_string(addr.clone()) {
+            Some(tor_address_string) => {
+                Ok(Box::pin(do_tor_dial(self.socks_port, tor_address_string)))
+            }
+            _ => {
+                tracing::warn!(
+                    "Address {} could not be formatted. Dialling via clear net",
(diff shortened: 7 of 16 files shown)


swap/src/network/transport.rs
```diff
@@ -1,28 +1,31 @@
...
-
+pub fn build(id_keys: &identity::Keypair, tor_socks5_port: Option<u16>) -> Result<SwapTransport> {
     let dh_keys = noise::Keypair::<X25519Spec>::new().into_authentic(id_keys)?;
     let noise = NoiseConfig::xx(dh_keys).into_authenticated();
 
     let tcp = TokioTcpConfig::new().nodelay(true);
+    let tcp = match tor_socks5_port {
+        None => TorTcpConfig::new(tcp),
+        Some(tor_socks5_port) => TorTcpConfig::new(tcp).with_socks5_port(tor_socks5_port),
+    };
```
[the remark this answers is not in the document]
thomaseizinger: That will allow you to get rid of the `Option` within the Tor transport.

thomaseizinger (commented):
Nice work!

My comments are a bit all over the place, so I'll summarize things here:

I think we should move the creation of the ephemeral services into the `Transport` implementation. This makes everything much more self-contained.

A more self-contained implementation is easier to test and hence allows us to drop the e2e test again (we already have too many IMO). I'd rather not add another one that is identical apart from the transport protocol.

I am not fussed about whether the implementation goes into its own crate (in this workspace) or is just a single module within `swap`. What is important to me is that we can test the Tor transport with a trivial protocol like `ping` instead of running a full swap.

swap/src/bin/asb.rs
```diff
@@ -212,3 +232,46 @@ async fn init_monero_wallet(
...
+
+    ac.add_services(&hidden_services_details, &key).await?;
+
+    let onion_address = key
+        .public()
+        .get_onion_address()
+        .get_address_without_dot_onion();
+
+    hidden_services_details.iter().for_each(|(port, _)| {
+        tracing::info!("/onion3/{}:{}", onion_address, port);
```
thomaseizinger, line 273: Another benefit of moving the creation of ephemeral services into the Transport will be that we can "reuse" the log in the `EventLoop` about `NewListenAddr` as that will be automatically triggered if we report a [`ListenerEvent::NewAddress`](https://docs.rs/libp2p/0.37.1/libp2p/core/transport/enum.ListenerEvent.html#variant.NewAddress).

swap/src/bin/asb.rs
```diff
@@ -212,3 +232,46 @@ async fn init_monero_wallet(
...
+                    // We only care for Tcp for now.
+                    None
+                }
+            })
+        })
+        .filter_map(|details| details)
+        .collect::<Vec<_>>();
+
+    // Note: ideally we have deterministic keys but unfortunately there is no Api
+    // for that yet.
```
thomaseizinger, line 262: Seems like any 64-bytes can be turned into a secret key: https://docs.rs/torut/0.1.8/torut/onion/struct.TorSecretKeyV3.html#impl-From%3C%5Bu8%3B%2064%5D%3E

thomaseizinger: Would be nice to derive these from the seed as well I guess.

swap/src/asb/config.rs
```diff
@@ -61,6 +63,22 @@ pub struct Monero {
     pub wallet_rpc_url: Url,
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct TorConf {
+    pub control_port: u16,
+    pub socks5_port: u16,
```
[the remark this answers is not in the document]
bonomat: Yes, I think we can but it didn't work out of the box. Maybe I misconfigured Tor or I need more rights to read the conf. 
It should be as simple as: `let socks_port = auc.get_conf("SocksPort").await?;`

swap/src/bin/asb.rs
```diff
@@ -73,6 +78,21 @@ async fn main() -> Result<()> {
 
     let env_config = env::Testnet::get_config();
 
+    let uac = tor::UnauthenticatedConnection::default()
+        .with_control_port(config.tor.control_port)
+        .with_socks5_port(config.tor.socks5_port);
+    let (tor_socks5_port, _ac) = match uac.assert_tor_running().await {
+        Ok(_) => {
+            tracing::info!("Tor found. Setting up hidden service. ");
+            let ac = register_tor_services(config.network.clone().listen, uac).await?;
```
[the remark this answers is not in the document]
bonomat: > At this point in the program's lifecycle, you can't actually know whether or not you will be listening on these addresses.

I don't understand why not. You configure your service to listen on these addresses, why would it change?

> This snippet of code should either run:
> Here in the eventloop

The `eventloop` could work as well. However, ` SwarmEvent::NewListenAddr(addr)` will be fired for each local listening address. We should only care about `localhost`/`127.0.0.1`. 

> Like we discussed the other day, the latter would be the cleanest because it makes the whole Tor transport module self-contained.

I played with this thought but disagree with it. I think it would be weird if a network behavior has the side effect of interacting with Tor and adding hidden service. 


Personally I think the cleanest is to add hidden services from the outside as we have more information about the listening addresses we care about. 
Not everyone might be comfortable with handing over the control port to the application and with this approach, a user can configure the hidden service from the outside of the application. 
I used this approach successfully while testing and only added the feature of adding hidden Tor services later on.

swap/src/cli/command.rs
```diff
@@ -48,6 +50,9 @@ pub enum Command {
 
         #[structopt(flatten)]
         monero_params: MoneroParams,
+
+        #[structopt(long = "tor-socks5-port", help = "Your local Tor socks5 proxy port", default_value = DEFAULT_TOR_SOCKS5_PORT)]
+        tor_socks5_port: u16,
```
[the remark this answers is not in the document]
bonomat: He can, if the the port is not reachable, we send everything via clearnet. 
Maybe the an optional field ?

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,132 @@
...
+            (Some(tor_address_string), Some(socks_port)) => {
+                tracing::debug!("Dialing Tor destination address: {}", tor_address_string);
+                Ok(Box::pin(do_tor_dial(socks_port, tor_address_string)))
+            }
+            _ => self.inner.dial(addr),
+        }
+    }
+
+    fn address_translation(&self, listen: &Multiaddr, observed: &Multiaddr) -> Option<Multiaddr> {
+        self.inner.address_translation(listen, observed)
```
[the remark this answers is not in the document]
bonomat: I wasn't sure what implications this will have.

swap/tests/happy_path_tor.rs
```diff
@@ -0,0 +1,29 @@
...
+use harness::SlowCancelConfig;
+use swap::protocol::{alice, bob};
+use tokio::join;
+
+/// Run the following tests with RUST_MIN_STACK=10000000
+
+/// Note: for this test to work, Tor needs to be running with control port 9051
+/// and socks5 proxy port on 9050
+#[tokio::test]
+async fn swap_through_tor() {
```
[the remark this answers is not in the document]
bonomat: A standalone ping-pong example using Tor can be found here: https://github.com/bonomat/libp2p-ping-pong-tor

This test actually helped me finding a problem of converting the provided multi-address to an onion address because we do something different than in the PoC I wrote above. 

I agree that a independent test of the transport itself might be more efficient but I wouldn't know how to start to decouple this from the current implementation.

swap/src/bin/asb.rs
bonomat: Cool, didn't see that from/into is derived.

swap/src/bin/asb.rs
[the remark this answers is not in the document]
thomaseizinger: > I don't understand why not. You configure your service to listen on these addresses, why would it change?

It is not about whether or not it changes. It is about whether or not the _intent_ to listen on these addresses actually succeeds. `Swarm::listen_on` is asynchronous. Calling the function by itself doesn't do anything other than modifying some state. That state is picked up on the next `poll` of the Swarm and the transport is instructed to _try_ and listen on the given multi-address.

You are not actually listening on the interface until the underlying transport emits `ListenerEvent::NewAddress`.

> However, SwarmEvent::NewListenAddr(addr) will be fired for each local listening address. We should only care about localhost/127.0.0.1.

There is `is_loopback()`: https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_loopback

> I played with this thought but disagree with it. I think it would be weird if a network behavior has the side effect of interacting with Tor and adding hidden service.

I agree that a `NetworkBehaviour` would be weird. I am talking about the `Transport` though. A `Transport`'s responsibility is to _dial_ and _listen_ using a certain transport protocol. If you have a Tor-`Transport`, it seems perfectly reasonable that the implementation somehow achieves that it is reachable via Tor. One way of doing this is setting up an ephemeral service. There are other ways as well so we will probably need some configuration at some point.

On the point of side-effects: A `NetworkBehaviour` as well as a `Transport` need to be **explicitly** included by the user in the configuration. It doesn't appear out of nowhere. As part of choosing to include such a module, I would hope that they read the documentation and are aware of what provided functionality does.

> Personally I think the cleanest is to add hidden services from the outside as we have more information about the listening addresses we care about.

There seems to be a misunderstanding here. Like I explained above, you actually have _less_ information about the listen addresses on the outside. Already in-use ports are a problem for example. Another one is wildcard addresses like `0.0.0.0` that only get resolved to concrete interfaces by the kernel once you start listening. A third one might be unsupported multi-addresses like trying to listen on `/ip4/0.0.0.0/udp/3000` although you don't have a Udp transport configured.

To summarize, the `Transport` implementation has absolutely the most knowledge about what is going on in terms of which interfaces are being listened on etc. For example, the `Swarm` can only emit a `SwarmEvent::NewListenAddr` because the `Transport` emitted a `ListenerEvent::NewAddr` event.

> Not everyone might be comfortable with handing over the control port to the application and with this approach, a user can configure the hidden service from the outside of the application.

Happy to make it configurable then. From our discussion, I see at least three possible options:

- Don't listen on Tor at all, i.e. just pass through to the underlying TCP transport
- "Listen" on an existing hidden service. This one is a bit weird because effectively, the only thing you can do is emit a `ListenerEvent::NewAddr` and inform the rest of the application that we are listening on a particular onion address. I guess we could do some validation with the Tor node that the user set up Tor correctly to route to the port that we are actually listening on.
- Automatically set up one or more hidden service(s) for the addresses we are listening on. Can probably make this configurable to use random keys or existing ones.

---

In summary, having this code inside the `Transport` achieves multiple things:

- It is more correct. Unless you react to events of the `Swarm` about successful / failed listening, you don't know whether or not you are actually listening on an interface.
- It makes the functionality reusable because adding the transport is basically plug-and-play.
- Conceptually, a `Transport` is about listening and dialling. Having an implementation do only one of the two is a code-smell. The code-smell occurs again in the code inside `main` that sets up the hidden service that just sits there and doesn't know where it wants to belong.
- It makes the functionality independently testable. We can make a swarm that automatically sets up a hidden service, emits the correct listen address and we have another swarm that connects to it. That is how (to my knowledge) all `Transport`s are tested in rust-libp2p.

Hope this clears things up!

swap/src/network/tor_transport.rs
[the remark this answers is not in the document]
thomaseizinger: Usually, this is used to translate externally observed ports to actual listening ports. With Tor, this should be completely irrelevant and just passing through to `Tcp` certainly won't help much :sweat_smile:

swap/src/cli/command.rs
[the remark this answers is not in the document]
thomaseizinger: > He can, if the the port is not reachable, we send everything via clearnet.

That is what I meant with "providing an invalid port" and I think it is quite weird that I have to choose and invalid port to disable Tor. Ideally, I can say `--no-tor` or something. I didn't want to get into the details of actually designing the CLI here just pointing out the lack of a clear "no tor" option.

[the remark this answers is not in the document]
bonomat: That's a hard one :D 

I played around with it a bit way too long and think the current solution is the best one. 

The issue is that if `alice_multiaddr` is an onion address, we need to have a `socks5_port`. 
If this can be expressed with `structopt`, that'd be awesome. 

- 
I wasn't able to achieve this, so I have the issue that if `tor_socks5_port` is an option and not provided, we will fail eventually when trying to dial to Alice. 

If it is not an option and defaults to a default port, we can fall back to clear net if Tor is not available (current implementation). 

So what I'm doing now is I provide a default port and fail Swap early if an onion address was provided but Tor was not running on the default port. 

This seem to me the best trade-off, e.g. a user who has Tor running on default port can simple do `swap buy-xmr -seller /onion/...` and the the communication will go through the default port. 

Check the latest commit for some details: #1a1b6ae

swap/src/asb/config.rs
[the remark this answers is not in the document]
bonomat: I played around with it and have a bit and got it working with a different command. No idea why `get_conf` does not work. But it always returns an empty vec.
Check the latest commit for some details: b5e4fe0 (#438).

.github/workflows/ci.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         if: matrix.target == 'armv7-unknown-linux-gnueabihf'
         run: |
           sudo apt-get update
-          sudo apt-get install gcc-arm-linux-gnueabihf
+          sudo apt-get install gcc-arm-linux-gnueabihf libssl-dev
```
bonomat, line 67: Should not be needed if vendored-openssl is used.

swap/src/bin/asb.rs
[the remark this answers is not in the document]
thomaseizinger: So I thought about this in more detail and had some insights.

I came around to the view that a `Transport` shouldn't just magically do things. It should only do something once the user gave a clear instruction via `swarm.listen_on`. For example, a `TcpTransport` doesn't do anything unless the user says:

```
swarm.listen_on(/ip4/0.0.0.0/**tcp**/8080)
```

Hence, I think we should design the `TorTransport` in a way that it only does something if the user calls:

```
swarm.listen_on(/onion3/vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd:1234")
```

To make things work in the above case, the `Transport` would have to:

- assume that the hidden service is already set up
- use the control port to read information about the hidden service and figure out what port it forwards to locally
- listen on the local tcp port to make sure we receive the traffic coming in at this onion address

This will cover the usecase where the user wants to set up Tor outside of the application.

To make use of ephemeral services, we could take some inspiration from "wildcard" addresses: `0.0.0.0` is not actually a valid IP address and you can't listen on it. It is merely a placeholder that just shares the same syntax.

To convey the intent that the `TorTransport` should set up an ephemeral service, we could react to a special onion address in the form of:

```
swarm.listen_on(/onion3/00000000000000000000000000000000000000000000000000000000:1234")
```

Given a multi-address like this, the `TorTransport` would:

- listen on a TCP port locally
- generate a new key and set up an ephemeral service that forwards to the local TCP port
- emit a `NewListenAddr` event with the generated onion address to inform the rest of the application about it

There are a few more corner cases in regards to choosing ports etc but I think overall, this would be a much better design:

- We clearly separate ourselves from the `TcpTransport`, i.e. a `TorTransport` should not react to a `/ip4/0.0.0.0/tcp/0` multiaddress being passed to `swarm.listen_on`.
- We leave it completely up to the user to manage the hidden service themselves or having the application set up a hidden service.
- We can support dialling via Tor without having to listen on Tor.
- We don't have to filter addresses for localhost only: In the case of an ephemeral service, we can explicitly listen on `127.0.0.1`.

swap/src/asb/config.rs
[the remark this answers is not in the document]
bonomat: I undid the latest commit. The initial approach resulted in a better UX.

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,124 @@
...
+            Ok(stream)
+        }
+
+        match to_onion_address(addr.clone()) {
+            Some(tor_address_string) => {
+                Ok(Box::pin(do_tor_dial(self.socks_port, tor_address_string)))
+            }
+            _ => self.inner.dial(addr),
+        }
+    }
```
bonomat, line 62: @thomaseizinger/ @da-kami : I thought ab it more more on this particular part: Assuming the user configured a Socks5 proxy port we should assume that she wants all connections to go through that port, i.e. if the user provides a Tor socks5 port, we should always dial through this port.  

As of now, we only dial through the socks5 port if we received an onion address. 
Now, if we always want to do this, we need to manually translate the provided multiaddress into a string of format `ADDR:PORT`. I tried to achieve this on my playground project to send all traffic through Tor but failed so far. It could be that this is because Tor blocks traffic to localhost, so this would need some more investigation.

swap/src/asb/config.rs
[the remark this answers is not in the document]
thomaseizinger: > The initial approach resulted in a better UX.

I am curious, why is that? I had a look at the commit and - as long as it works - it looks like we can successfully reduce the configuration surface?

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Websocket support for the ASB.
   The ASB is now capable to listen on both TCP and Websocket connections.
   Default websocket listening port is 9940.
+- Tor support as an optional feature. If ASB detects that Tor's control port is open, an hidden service is created for
+  network it is listening on (currently 2). The Tor control port as well as Tor socks5 proxy port is configurable.
```
thomaseizinger, line 33: Nit: Needs newlines after the sentences.
```suggestion
- Tor support as an optional feature.
  If ASB detects that Tor's control port is open, an hidden service is created for
  network it is listening on (currently 2).
  The Tor control port as well as Tor socks5 proxy port is configurable.
```

swap/Cargo.toml
```diff
@@ -48,10 +50,12 @@ structopt = "0.3"
 strum = { version = "0.20", features = ["derive"] }
 thiserror = "1"
 time = "0.2"
-tokio = { version = "1", features = ["rt-multi-thread", "time", "macros", "sync", "process", "fs"] }
+tokio = { version = "1", features = ["rt-multi-thread", "time", "macros", "sync", "process", "fs", "net"] }
+tokio-socks = "0.5"
 tokio-tungstenite = { version = "0.14", features = [ "rustls-tls" ] }
 tokio-util = { version = "0.6", features = ["io"] }
 toml = "0.5"
+torut = { version = "0.1", features = ["serialize", "v2", "v3", "control", "vendored_openssl"] }
```
thomaseizinger, line 58: Is the traffic to the socks5 proxy encrypted or what do we need openssl for?

Onion v2 is EOL as far as I know, so if it is only for the RSA keys in Onion v2, then we should drop v2 support IMO. Not worth the added dependency baggage.

swap/src/bin/asb.rs
```diff
@@ -73,6 +77,21 @@ async fn main() -> Result<()> {
 
     let env_config = env::Testnet::get_config();
 
+    let uac = tor::UnauthenticatedConnection::default()
+        .with_control_port(config.tor.control_port)
+        .with_socks5_port(config.tor.socks5_port);
```
thomaseizinger, line 82: A builder pattern is only useful if the `default` instance is useful by itself. I haven't dug into the details but it looks like we always want to configure at least the socks5 port. Hence we should design this in a way so we can't make an instance of `UnauthenticatedConnection` without passing that port in.

I am also not convinced that carrying over the terminology of "Unauthenticated" vs "Authenticated" connection from torut actually helps us here in anyway. A more general "TorClient" might be a better fit.

swap/src/bin/swap.rs
```diff
@@ -76,7 +77,14 @@ async fn main() -> Result<()> {
...
             let bitcoin_wallet = Arc::new(bitcoin_wallet);
 
-            let mut swarm = swarm::bob(&seed, alice_peer_id)?;
+            let tor_socks5_port = {
+                let uac =
+                    tor::UnauthenticatedConnection::default().with_socks5_port(tor_socks5_port);
+                uac.assert_tor_running()
+                    .await
+                    .map_or(None, |_| Some(uac.tor_proxy_port()))
+            };
```
thomaseizinger, line 86: This is duplicated between two match arms. I think this could also just move into the `swarm::bob` function given that it already "knows" about the socks5 port.

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,124 @@
...
+use libp2p::core::Transport;
+use libp2p::tcp::tokio::{Tcp, TcpStream};
+use libp2p::tcp::{GenTcpConfig, TcpListenStream, TokioTcpConfig};
+use std::io;
+use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::pin::Pin;
+use tokio_socks::tcp::Socks5Stream;
+use tokio_socks::IntoTargetAddr;
+
+/// Represents the configuration for a TCP/IP transport capability for libp2p.
```
thomaseizinger, line 15: Seems this wasn't adapted yet.

swap/src/network/tor_transport.rs
thomaseizinger: > if the user provides a Tor socks5 port, we should always dial through this port.

I agree.

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// iterates trhough multi address until we have onion protocol, else return
+/// None Tor expects address in form: ADDR.onion:PORT
+fn to_onion_address(multi: Multiaddr) -> Option<String> {
+    let components = multi.iter();
+    for protocol in components {
+        match protocol {
+            Protocol::Onion(addr, port) => {
+                tracing::warn!("Onion service v2 is being deprecated, consider upgrading to v3");
```
thomaseizinger, line 76: They are already deprecated as of now, and support is being removed from the Tor codebase on July 15th. I don't think we should support v2.

swap/src/bin/asb.rs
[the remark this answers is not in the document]
bonomat: Thanks for writing this up so clearly. Very Interesting idea. 
Some thoughts and questions from my side: 

> [...]
> ```
> swarm.listen_on(/onion3/vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd:1234")
> ```
> 

>   -  assume that the hidden service is already set up
>   -  use the control port to read information about the hidden service and figure out what port it forwards to locally
>   -  listen on the local tcp port to make sure we receive the traffic coming in at this onion address

From my current understand of Tor this feels a bit _weird_ because the application is not listening on this onion address. It's your local Tor instance which does that for you. 

The other thing is that I don't see a UX improvement with telling the application to listen on `/onion3/vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd:1234` vs telling it to listen on `/ipv/127.0.0.1/tcp/1234`. 
The user has to configure Tor in the first place knowing that the there is a service listening already or that he will start a service which listens on this address. To be precise, the user will need to add the following to their torrc file: 

```
HiddenServiceDir /usr/local/etc/tor/hidden_service
HiddenServicePort 3000 127.0.0.1:3000
```

>   -  use the control port to read information about the hidden service and figure out what port it forwards to locally

I would try to avoid any deep interaction with the control port. the control port spec are very verbose and if things fail it's not clear why. 
To give you a feeling, checkout the control port spec Section `3.9. GETINFO` from here https://gitweb.torproject.org/torspec.git/tree/control-spec.txt
In particular, these two requests seem to be relevant: 

```
    "hs/client/desc/id/<ADDR>"
      Prints the content of the hidden service descriptor corresponding to
      the given <ADDR> which is an onion address without the ".onion" part.
      The client's cache is queried to find the descriptor. The format of
      the descriptor is described in section 1.3 of the rend-spec.txt
      document.

      If <ADDR> is unrecognized or if not found in the cache, a 551 error is
      returned.

      [New in Tor 0.2.7.1-alpha]
      [HS v3 support added 0.3.3.1-alpha]

    "hs/service/desc/id/<ADDR>"
      Prints the content of the hidden service descriptor corresponding to
      the given <ADDR> which is an onion address without the ".onion" part.
      The service's local descriptor cache is queried to find the descriptor.
      The format of the descriptor is described in section 1.3 of the
      rend-spec.txt document.

      If <ADDR> is unrecognized or if not found in the cache, a 551 error is
      returned.

      [New in Tor 0.2.7.2-alpha]
      [HS v3 support added 0.3.3.1-alpha]
```

I'm not even sure what the responds will look like 😅


> To make use of ephemeral services, we could take some inspiration from "wildcard" addresses: 0.0.0.0 is not actually a valid IP address and you can't listen on it. It is merely a placeholder that just shares the same syntax.
> 
> To convey the intent that the TorTransport should set up an ephemeral service, we could react to a special onion address in the form of:
> 
> ```
> swarm.listen_on(/onion3/00000000000000000000000000000000000000000000000000000000:1234")
> ```

That's an interesting idea. Unfortunately parsing `/onion3/00000000000000000000000000000000000000000000000000000000` to a multi address will fail because it is not a valid multi-address format. It expects the onion address to be in base32 format, e.g. `oarchy4tamydxcitaki6bc2v4leza6v35iezmu2chg2bap63sv6f2did`. 

I understand that you would like to just have a _wild_card_ address here but I wouldn't know how a _nice_ wild_card address could look like. 

Imho a better UX would result from asking the user to explicitly state if she wants to have an hidden service registered for her or not.

swap/Cargo.toml
bonomat: Interesting, Torout seems to need openssl only for v2 keys. So dropping this should allow us to drop openssl :) Let's see what CI says.

swap/Cargo.toml
```diff
@@ -64,6 +63,13 @@ url = { version = "2", features = ["serde"] }
 uuid = { version = "0.8", features = ["serde", "v4"] }
 void = "1"
 
+[dependencies.torut]
+branch = "feature-flag-tor-secret-keys"
+default-features = false
+features = ["v3", "control"]
+git = "https://github.com/bonomat/torut/"
+version = "0.1"
```
bonomat, line 71: We temporarily depend on this fork until the PR is merged: https://github.com/teawithsand/torut/pull/10

thomaseizinger: You might be interested in learning about the [`[patch]`](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) section inside `Cargo.toml` files. Useful to do things like this without having to modify the original manifest.

The good thing is that `[patch]` is applied across the whole dependency tree and hence also affects transitive dependencies.

bonomat: cool stuff :)

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Websocket support for the ASB.
   The ASB is now capable to listen on both TCP and Websocket connections.
   Default websocket listening port is 9940.
+- Tor support as an optional feature.
+  If ASB detects that Tor's control port is open, an hidden service is created for
```
da-kami, line 33:
```suggestion
  If ASB detects that Tor's control port is open, a hidden service is created for
```

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  network it is listening on (currently 2).
```
da-kami, line 34:
```suggestion
  the network it is listening on (currently 2).
```

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,225 @@
...
+            Ok(stream)
+        }
+
+        match to_address_string(addr.clone()) {
+            Some(tor_address_string) => {
+                Ok(Box::pin(do_tor_dial(self.socks_port, tor_address_string)))
+            }
+            _ => {
+                tracing::warn!(
+                    "Address {} could not be formatted. Dialling via clear net",
```
da-kami, line 63:
```suggestion
                    "Address {} could not be formatted as onion address. Dialling via clear net",
```

This message might be somewhat confusing for a user. Not sure if it should be on `warn` level - but given that someone might want to stay anonymous it is better to warn them when going clearnet in case they dial `dnsaddr`

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,225 @@
...
+            Protocol::Ip4(addr) => Some(format!("{}", addr)),
+            Protocol::Ip6(addr) => Some(format!("{}", addr)),
+            Protocol::Dns(addr) => Some(format!("{}", addr)),
+            Protocol::Dns4(addr) => Some(format!("{}", addr)),
+            _ => None,
+        })
+        .collect::<Vec<_>>();
+    if address_string.is_empty() {
+        tracing::warn!(
+            "Could not format address {}. Please consider reporting this issue. ",
```
da-kami, line 105: to my understanding this concerns `dnsaddr` not being supported. I don't think we should print that as a warning here, given that we have an additional warning on the caller side. Also I am not sure the `Please consider reporting this issue.` is helpful and the message should include what is the problem more clearly (i.e. `Could not format address {} as onion address (...)`.

swap/src/network/tor_transport.rs
```diff
@@ -0,0 +1,225 @@
...
+            Some(address_string)
+        }
+        Ordering::Less => Some(address_string),
+        Ordering::Equal => Some(format!(
+            "{}:{}",
+            address_string,
+            port.get(0)
+                .expect("Already verified the length of the vec.")
+        )),
+    }
```
thomaseizinger, line 141: That looks quite complicated. Any reason you can't just step through the iterator as needed?

```rust
let mut protocols = addr.iter();

let address = match protocols.next() {
	// extract ip4 etc, fail if non-matching
	// do proper error handling on `None`
};

let port = match protocols.next() {
	// extract port for tcp / udp etc, fail if non-matching
	// do proper error handling on `None`
};

format!("{}:{}", address, port)

```

thomaseizinger: Also, a `Result` seems more appropriate if we _fail_ to extract a usable address.

da-kami (approved):
Thanks for putting this up! Good to go in from my side, I don't see any blockers that cannot be fixed in a follow up PR. We might want to adapt some of the messages though as they might be confusing for users.

swap/tests/harness/mod.rs
```diff
@@ -420,8 +423,13 @@ impl BobParams {
...
     }
 
-    pub fn new_eventloop(&self, swap_id: Uuid) -> Result<(bob::EventLoop, bob::EventLoopHandle)> {
-        let mut swarm = swarm::bob(&self.seed, self.alice_peer_id)?;
+    pub async fn new_eventloop(
+        &self,
+        swap_id: Uuid,
+    ) -> Result<(bob::EventLoop, bob::EventLoopHandle)> {
+        let tor_socks5_port = get_port()
+            .expect("We don't care about Tor in the tests so we get a free port to disable it.");
```
thomaseizinger, line 431: Why does it need to be a free port? Is `0` not sufficient?

thomaseizinger (commented):
Looking good.

The other day we discussed (and to my knowledge agreed) that a design where the user primarily uses the `Swarm::listen_on` interface to control Tor would be better.

Do you still plan on doing this?

I would strongly prefer it because it serves as the much better foundation to iterate on as things are more contained. Like we discussed, in an MVP implementation, we don't need to include all the possibilities like user-configured Tor services if what we initially ship always uses ephemeral services anyway.

As such I think it shouldn't be a lot of work. The only thing we need to figure out is how we represent such a "wildcard" onion address.

thomaseizinger:
> The only thing we need to figure out is how we represent such a "wildcard" onion address.

Looking at the base32 alphabet, this should be a valid onion3 address:

```
"/onion3/WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW:65535"
```

I chose `W` for "wildcard" and `65535` because it is the max u16 value. We can't use 0 because the multi-address parser [doesn't allow this](https://github.com/libp2p/rust-libp2p/blob/e3af8b7d03bdf7e4b156481102a6a961c23a8946/misc/multiaddr/src/protocol.rs#L557-L560). I think we could try patch that upstream. Multiaddresses with port 0 are valid for other protocols, I don't see why we shouldn't allow this here as well. For now though, `65535` should be a good enough magic value :)

thomaseizinger:
More options if you don't like `WWWW...`: :laughing: 

- `/onion3/CANIHAZNEWEPHEMERALHIDDENSERVICEPLZZZZZZZZZZZZZZZZZZZZZZ:65535`
- `/onion3/ROUTEALLLLLLLLLLLLLLLLLLLLLLLLLLLLLLTHETRAFFICTHROUGHTOR:65535`

swap/src/network/tor_transport.rs
bonomat: Regarding your change: `to_address_string` is not formatting it to an onion address but formats it to a string of the format `address:port` so that we can dial through Tor. 

I agree that this might not be confusing to some and was deciding between failing, i.e. not dialing at all, or just falling back to clear net. 
I opted for dialing via clear net in the case we cannot format that address string.

swap/src/network/tor_transport.rs
bonomat: Unfortunately there are in theory 15 other Protocol types which we did not consider (see a few lines above where we match on `protocol`). 
That's why I just print a warning and hope that someone reports it so that we can improve on the function. 
As of now, I covered all cases I could think of and thought that `Could not format address {}. Please consider reporting this issue` sounds better than `This should never happen` :D

swap/src/network/tor_transport.rs
bonomat: Agree, it was also cumbersome to write. :) 

Unfortunately I do not know how many protocols in the multi address might or might not be. Hence the iterator.

swap/tests/harness/mod.rs
bonomat: I did not try `0`. I did not want by accident chose a port which exists because we try to connect to this port and fail if it is not reachable.

swap/src/network/tor_transport.rs
bonomat: Atm we do not fail but fall back to the default Tcp dialling.

bonomat: Check out the last commit #00b40c4a8fd92276d89517fcac4ee311f68e4b3e :)

bonomat:
@thomaseizinger : sorry, missed your comment yesterday. 

>Looking good.

> The other day we discussed (and to my knowledge agreed) that a design where the user primarily uses the Swarm::listen_on interface to control Tor would be better.
> 
> Do you still plan on doing this?
> 
> I would strongly prefer it because it serves as the much better foundation to iterate on as things are more contained. Like we discussed, in an MVP implementation, we don't need to include all the possibilities like user-configured Tor services if what we initially ship always uses ephemeral services anyway.
> 
> As such I think it shouldn't be a lot of work. The only thing we need to figure out is how we represent such a "wildcard" onion address.

There might have been a misunderstanding then 😅. I thought we agreed that this would be a nice-to-have and an also elegant solution is to treat an application which is only listening (i.e. the Alice's transport) in a way that it does not know about Tor. This allows us to achieve the scenario where the user does not want the application to create a hidden service without needing "wildcard" addresses. 
That's why I refactored the protocols so that `Alice` uses a transport which does not know about Tor at all. 😅

In regards to the wildcard address, I don't think it adds that much value atm because the only thing which would change is how the user would configure the ASB. 

> Do you still plan on doing this?

With that being said, I'd rather move on and focus on what is more urgent, e.g. dynamic transaction fees.

thomaseizinger:
> This allows us to achieve the scenario where the user does not want the application to create a hidden service without needing "wildcard" addresses.
> That's why I refactored the protocols so that `Alice` uses a transport which does not know about Tor at all. sweat_smile

But from the user's perspective, we do create a hidden service if Tor is available? The user is not going to care about the internal architecture (i.e. whether or not the transport is agnostic to Tor or not), so from their perspective, the application will "make Tor work" if it is there.

I see two paths forward:

1. If an automatically-created hidden service is the feature we want, then I would like to see this added in the already discussed way. The primary reason for my position here is **feature** driven. Consistency within the configuration surface of an application is important and configuring Tor the same way as you'd configure any other transport is a huge plus. It may not yet be as visible but to give some examples of what I mean:
    - If we ever add support for predefined hidden services, simply replacing that wildcard address with the address you want to use is much more intuitive then adding a dedicated configuration section.
    - Disabling Tor support is as simple as not including an onion address in the config. With the current design, the user needs to actively provide an invalid port if for some reason Tor is running on the current machine and they want to disable it.

As an added benefit, the discussed design also results in a much more contained module which I would argue is a sign that this is the way to go.

2. If we don't want the ASB to automatically create a hidden service, then this is fine as well but then we should also not need to change anything in the ASB. The ASB never dials, which means simply only listening on a local port and configuring Tor outside of the application will do the trick. We will need changes for the CLI but that is a different story. We can just create a different `Transport` for each application.

I think it is important to strike a balance between providing features and adding complexity and in the current state of this PR, I see an imbalance here not only in complexity for the user but also in the form of technical debt which results in complexity that we need to deal with when working with the codebase on a daily basis.

> > Do you still plan on doing this?
> 
> With that being said, I'd rather move on and focus on what is more urgent, e.g. dynamic transaction fees.

Okay. How do we proceed with this PR then? Close? Move to draft? Someone else taking over?

bonomat:
This PR is feature complete from both sides, the ASB and the swap cli. It's been reviewed extensively and the feedback and improvements were implemented. 
It's been successfully tested and allows users now to preserve their privacy which is an important requirement.

The main thing which is still being debated here is _where_ the hidden services should be created. 
It is obvious that more discussions are needed for this and hence I created a follow-up ticket in #457 so that we can merge this PR.

Issue #411, opened 2021-04-13
Try save information received for swaps currently not in execution in the database


da-kami:
Follow up of https://github.com/comit-network/xmr-btc-swap/pull/405

Try to save information received for a swap that is currently not in execution into the database so it is not lost (i.e. the other party does not resend it):

For Bob: transfer proofs received from Alice
For Alice: Encryption signatures received from Bob

If no matching swap is found in database the information should be dropped.
Add separate trees in database to allow saving the information.
Check for the information in the database when entering corresponding states of the respective party to allow recovery of information from db.

binarybaron:
Transfer proofs are now being buffered (https://github.com/comit-network/xmr-btc-swap/pull/1669)

PR #451, opened 2021-04-26
Peer id check for transfer proof and encsig messages

### LINKED ISSUE #416: Ensure incoming message for swap is from swap-peer
With #405 we allow concurrent swaps with the same peer and agree on a swap ID during swap setup to be able to identify the swap message is for, messages being transfer proof (Alice -> Bob) and encryption signature (Bob -> Alice). 

In the new implementation the parties do not ensure that an incoming message is coming from the peer-id the swap was started with. A malicious peer could intercept the swap ID and send messages to swaps causing unwanted behavior.

- Alice should ensure that incoming encryption signatures are coming from the `peer-id` that the swap was started with. 
- Bob should ensure that incoming transfer proofs are coming from the `peer-id` of the ASB the swap was started with.

da-kami:
Fixes #416 

The second commit will be relevant for changes planned in #411 

@thomaseizinger with #411 we will need access to Bob's database in the eventloop, this might influence design decisions for #401

### DIFF
--- a/swap/src/protocol/alice/event_loop.rs
+++ b/swap/src/protocol/alice/event_loop.rs
@@ -195,11 +195,32 @@ where
                             }
                         }
                         SwarmEvent::Behaviour(OutEvent::EncryptedSignatureReceived{ msg, channel, peer }) => {
-                            let sender = match self.recv_encrypted_signature.remove(&msg.swap_id) {
+                            let swap_id = msg.swap_id;
+                            let swap_peer = self.db.get_peer_id(swap_id);
+
+                            // Ensure that an incoming encrypted signature is sent by the peer-id associated with the swap
+                            let swap_peer = match swap_peer {
+                                Ok(swap_peer) => swap_peer,
+                                Err(_) => {
+                                    tracing::warn!("Ignoring encrypted signature for unknown swap {} from {}", swap_id, peer);
+                                    continue;
+                                }
+                            };
+
+                            if swap_peer != peer {
+                                tracing::warn!(
+                                    %swap_id,
+                                    "Ignoring malicious encrypted signature from {}, expected to receive it from {}",
+                                    peer,
+                                    swap_peer);
+                                continue;
+                            }
+
+                            let sender = match self.recv_encrypted_signature.remove(&swap_id) {
                                 Some(sender) => sender,
                                 None => {
                                     // TODO: Don't just drop encsig if we currently don't have a running swap for it, save in db
-                                    tracing::warn!(%peer, "No sender for encrypted signature, maybe already handled?");
+                                    tracing::warn!(%swap_id, "No sender for encrypted signature, maybe already handled?");
                                     continue;
                                 }
                             };
@@ -207,7 +228,7 @@ where
                             let mut responder = match sender.send(msg.tx_redeem_encsig).await {
                                 Ok(responder) => responder,
                                 Err(_) => {
-                                    tracing::warn!(%peer, "Failed to relay encrypted signature to swap");
+                                    tracing::warn!(%swap_id, "Failed to relay encrypted signature to swap");
                                     continue;
                                 }
                             };
--- a/swap/src/protocol/bob/event_loop.rs
+++ b/swap/src/protocol/bob/event_loop.rs
@@ -117,11 +117,22 @@ impl EventLoop {
                                 let _ = responder.respond(*response);
                             }
                         }
-                        SwarmEvent::Behaviour(OutEvent::TransferProofReceived { msg, channel }) => {
-                            if msg.swap_id != self.swap_id {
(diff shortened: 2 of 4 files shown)


bonomat (approved):
LGTM. Nice :)

swap/src/protocol/alice/event_loop.rs
```diff
@@ -195,19 +195,40 @@ where
                             }
                         }
                         SwarmEvent::Behaviour(OutEvent::EncryptedSignatureReceived{ msg, channel, peer }) => {
-                            let sender = match self.recv_encrypted_signature.remove(&msg.swap_id) {
+                            let swap_id = msg.swap_id;
+                            let swap_peer = self.db.get_peer_id(swap_id);
+
+                            // Ensure that an incoming encryption signature is sent by the peer-id associated with the swap
```
thomaseizinger, line 201: The message is an **encrypted** signature, not an encryption signature.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -195,19 +195,40 @@ where
...
+                            match swap_peer {
```
thomaseizinger, line 202: I think this would be easier to read if we use two `if let` statements because the current "happy-path" is hidden within the first match arm when the `if` condition **doesn't** evaluate to true. That is quite hard to understand IMO.

swap/src/protocol/alice/event_loop.rs
thomaseizinger: The typo is in a few places (including the commit message).

swap/src/protocol/bob/event_loop.rs
```diff
@@ -117,11 +117,22 @@ impl EventLoop {
...
+                            if peer != self.alice_peer_id {
+                                tracing::warn!(
+                                            %swap_id,
+                                            "Ignoring malicious transfer proof from {}, expected to receive it from {}",
+                                            peer,
+                                            self.alice_peer_id);
+                                        continue;
+                            }
+
+                            if swap_id != self.swap_id {
```
thomaseizinger, line 132: We keep loading up the event loop with these kind of fairly low-level checks when really, we should only be delegating to other components because this functionality in here is quite hard to test (as we can see by the lack of unit tests for this functionality).

Probably about time to create dedicated `NetworkBehaviour`s that encapsulate this kind of logic and only emit events to the `EventLoop` if they actually need processing.

thomaseizinger (changes requested):
Please fix the typo in the first commit, otherwise happy to merge this although I would like to see us _removing_ more of this logic from the eventloop instead of adding it because it bocomes really cluttered and is quite hard to test.

swap/src/protocol/alice/event_loop.rs
da-kami: I don't really understand how using two `if let` would make this section better. Could you give me some small example of what you have in mind. 
This is an early exit on error scenario - I don't understand why the happy path is hidden. There are other early exit scenarios defined in the block, the happy path comes afterwards.

swap/src/protocol/bob/event_loop.rs
da-kami: I totally agree, recorded this ticket https://github.com/comit-network/xmr-btc-swap/issues/458 for now. Planning to do some refactoring after we are mainnet ready :)

swap/src/protocol/alice/event_loop.rs
thomaseizinger: This `match` is a statement, not an expression, meaning the happy path goes **through** this statement instead of around it like we do in other places. Because it is not an expression, I have to evaluate the whole code before I know, what branches there are and under which conditions I am actually continuing with the code.

In particular, the happy path is hidden within the `Ok` branch if `swap_peer != peer` is `false` and this is only the happy path because all other branches use `continue`. In theory, there could be more paths through this statement. This makes it fairly hard to understand.

Compare that to:

```rust
let swap_peer = match swap_peer {
	Ok(swap_peer) => swap_peer,
	Err(_) => {
		tracing::warn!("Ignoring encrypted signature for unknown swap {} from {}", swap_id, peer);
        continue;
	}
};

if swap_peer != peer {
    tracing::warn!(
        %swap_id,
        "Ignoring malicious encrypted signature from {}, expected to receive it from {}",
        peer,
        swap_peer);
    continue;
}
```

da-kami: Ah, sorry, that was a misunderstanding then. Sure, splitting the `match` statement into two distinct parts is a good idea. I think I was confused by 

> two if let statements

because I could not see how I could make that happen :)
Done now.

thomaseizinger: Yes, initially I thought we need two `if let`s. Don't take what I say too literally :)

thomaseizinger (approved):

PR #452, opened 2021-04-27
Asb manual recovery commands

### LINKED ISSUE #377: Manual recovery commands for ASB
Add manual `cancel`, `refund` and `punish` commands for the ASB (similar to the CLI), so that a service provider has the chance to trigger these actions manually. Currently this would mean, that the service provider has to stop the ASB and trigger a different command due to database lock.

`cancel` - construct and send the Bitcoin cancel transaction if cancel timelock has expired (`--force` to overrule timelock check if we want)
`refund` - check for the Bitcoin refund transaction. If found trigger XMR refund scenario, i.e. claim Monero (and transfer to default wallet). Forcing this does not make much sense. 
`punish` - construct and send the Bitcoin punish transaction if the punish timelock uas expired (`--force` to overrule timelock check if we want)

da-kami:
Fixes #377 

Allows us to manually finish swaps if we run into error scenarios during execution.

I opted for putting all these commands under subcommand `manual-recovery` to run them e.g. (cancel):

```
./asb manual-recovery cancel [--force] --swap-id (...)
```

I combined the `cancel` and `refund` command in the same commit, because I only adapted the e2e tests once cancel and refund were implemented (cancel alone is a bit useless anyway...).


da-kami:
I don't have strong feelings about the last commit that moves the commands to a dedicated `recovery` module, just felt it would be a bit cleaner.

da-kami:
CI just ran into an upload error:

```
 With the provided path, there will be 1 files uploaded
Total file count: 1 ---- Processed file #0 (0.0%)
Total file count: 1 ---- Processed file #0 (0.0%)
Total size of all the files uploaded is 71801818 bytes
Error: read ECONNRESET
```

Re-running...

swap/src/protocol/alice/refund.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::protocol::alice::AliceState;
+use anyhow::{bail, Result};
+use libp2p::PeerId;
+use std::sync::Arc;
+use uuid::Uuid;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    // Errors indicating the the swap can *currently* not be refunded but might be later
+    #[error("Cannot refund because swap is not in a cancelled state. Make sure to cancel the swap before trying to refund or use --force.")]
```
thomaseizinger: This error message knows too much about its context. Think about how this will be composed:

A function `fn refund() -> Result<_, Error>` can be called with: `refund().context("Failed to refund swap")?`;

What this error needs to say is **why** it failed to refund. The fact that it failed is already implied by the error being instantiated.

If you compose these errors together, what you want to see is:

```
Failed to refund swap.

Caused by:
	Swap is not in a cancelled state
```

Your current implementation will say:


```
Failed to refund swap.

Caused by:
	Cannot refund because swap is not in a cancelled state.
```

Which is redundant.

thomaseizinger: Another hint that could have told you this is that you are repeating yourself at the beginning of every error message in this enum.

swap/src/protocol/alice/refund.rs
```diff
@@ -0,0 +1,123 @@
...
+            _ => return Ok(Err(Error::SwapNotCancelled)),
+        }
+    };
+
+    tracing::info!(%swap_id, "Trying to manually refund swap");
+
+    let spend_key = if let Ok(published_refund_tx) =
+        state3.check_for_tx_refund(bitcoin_wallet.as_ref()).await
+    {
+        tracing::debug!(%swap_id, "Bitcoin refund transaction found, extracting key to refund Monero");
```
thomaseizinger: Print the txid here?

thomaseizinger: Also, the user passed you the swap ID, seems redundant to include it in the log here.

swap/src/protocol/alice/state.rs
```diff
@@ -403,6 +404,58 @@ impl State3 {
...
+            .create_from(
+                file_name,
+                spend_key,
+                view_key,
+                monero_wallet_restore_blockheight,
+            )
+            .await?;
+
+        Ok(())
+    }
```
thomaseizinger, line 457: NACK adding more async functions to `State`. Ideally, `State` will be pure data transformations at some point.

swap/src/protocol/alice/state.rs
```diff
@@ -403,6 +404,58 @@ impl State3 {
...
+
+    pub async fn check_for_tx_refund(
+        &self,
+        bitcoin_wallet: &bitcoin::Wallet,
+    ) -> Result<Transaction> {
+        let tx_refund = self.tx_refund();
+        let tx = bitcoin_wallet.get_raw_transaction(tx_refund.txid()).await?;
+        Ok(tx)
+    }
```
thomaseizinger, line 423: This is poorly named. A function named `check_` hints that it doesn't return anything because it performs a check inside but this function doesn't check anything, it just returns you the transaction or an error.

swap/src/protocol/alice/state.rs
thomaseizinger: I think these can just be public free functions inside the swap module so you can call them from different places.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -0,0 +1,108 @@
...
+        let alice_swap = ctx.alice_next_swap().await;
+        let alice_swap = tokio::spawn(alice::run_until(alice_swap, is_xmr_lock_transaction_sent));
+
+        let bob_state = bob_swap.await??;
+        assert!(matches!(bob_state, BobState::BtcLocked { .. }));
+
+        let (bob_swap, bob_join_handle) = ctx
+            .stop_and_resume_bob_from_db(bob_join_handle, swap_id)
+            .await;
+        assert!(matches!(bob_swap.state, BobState::BtcLocked { .. }));
```
thomaseizinger, line 26: Are these asserts actually worth anything? Adds a lot of noise to the test for basically no gain (if we are worried that saving something to the DB corrupts the data, we should have more DB tests).

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        // Bob tries but fails to manually refund
+        bob::refund(bob_swap.id, bob_swap.bitcoin_wallet, bob_swap.db, false)
+            .await?
+            .unwrap_err();
```
thomaseizinger, line 72: We are we not asserting the error here?

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -0,0 +1,108 @@
...
+            .stop_and_resume_bob_from_db(bob_join_handle, swap_id)
+            .await;
+        assert!(matches!(bob_swap.state, BobState::BtcLocked { .. }));
+
+        ctx.restart_alice().await;
+        let alice_swap = ctx.alice_next_swap().await;
+        assert!(matches!(
+            alice_swap.state,
+            AliceState::XmrLockTransactionSent { .. }
+        ));
```
thomaseizinger, line 84: Same as above, I think we should get rid of these "pre-check" asserts.

Ideally, a test only contains an assert for things that matter to the test which in this case is about timelocks not being expired yet.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired_force.rs
```diff
@@ -0,0 +1,106 @@
...
+        assert!(matches!(
+            alice_state,
+            AliceState::XmrLockTransactionSent { .. }
+        ));
+
+        // Bob tries but fails to manually cancel
+        let is_outer_err = bob::cancel(bob_swap.id, bob_swap.bitcoin_wallet, bob_swap.db, true)
+            .await
+            .is_err();
+        assert!(is_outer_err);
```
thomaseizinger, line 38: Normal `assert!`s produce pretty useless error messages.

thomaseizinger (commented):
Overall design looks good, couple of nits.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
da-kami: I disagree :) 
This statement makes sure that nothing along the way of restarting the application is causing a side effect that changes the state. 
Granted, we only simulate the restart, but still, I think it is good that we ensure that we are the expected state after the restart.
In these rather complex e2e tests it is good to test all the assumptions on the way - if we don't do that it will be even harder to debug once a test fails, because we only fail in the end and don't know why!

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
da-kami: See my comment below please - These asserts are not just for `ensuring we saved in the db` but that nothing on the way of stopping/starting the application changes the state!
I don't find them noisy, I actually find them helpful in understanding more complex tests setups as in what to expect from the state at a given point in the test.

swap/src/asb/command.rs
```diff
@@ -52,6 +53,42 @@ pub enum Command {
...
     )]
     Balance,
+    #[structopt(about = "Contains sub-commands for recovering swap manually.")]
+    ManualRecovery(ManualRecovery),
+}
+
+#[derive(structopt::StructOpt, Debug)]
+pub enum ManualRecovery {
+    #[structopt(
+        about = "Cancel requires the cancel timelock to be expired. Cancel is a requirement for refund and punish."
```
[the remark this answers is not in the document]
da-kami: These are power user commands. I am not sure how much we want to go into detail here.
We could opt for using [`Arg::long_help`](https://docs.rs/structopt/0.3.21/structopt/#help-messages) but I am not sure if all users will get this. 
I was planning to elaborate more on this in the ASB docs instead of adding really long help texts to the commands. 

Someone who wants to properly use these commands HAS to have a proper understanding of the protocol.
I am not sure how much more education I can and want to pack into the help of these commands. 

I can add a sentence along the lines of `Publishes the Bitcoin cancel transaction.` - but I expect a user to expect that because they understand the protocol prior to using these commands. Do you have some concrete ideas as in what to add?

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Cancel command for the ASB that allows cancelling a specific swap by id.
+  Using the cancel command requires the cancel timelock to be expired, but `--force` can be used to circumvent this check.
```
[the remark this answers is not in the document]
da-kami: I wanted to keep the commands somewhat similar, but that is probably not a good idea :D
I can think about it - the thing is that for some commands `--force` disables not only the timelocks but also allows executing the command from a state that is not directly associated with the command.

I will re-visit this on Monday so I can give it some more thought :)

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
thomaseizinger: Let's agree to disagree then.

I just know from my own experience that tests which are noisy are seldomly checked for correctness and what is the point of a test if I don't verify that it is even correct?

>  In these rather complex e2e tests it is good to test all the assumptions on the way - if we don't do that it will be even harder to debug once a test fails, because we only fail in the end and don't know why!

That sounds more like a case for better logs than littering the test with assertions. Can also always set a breakpoint and inspect the state :man_shrugging:

swap/src/asb/command.rs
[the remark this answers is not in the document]
thomaseizinger: > These are power user commands. I am not sure how much we want to go into detail here.
> We could opt for using [`Arg::long_help`](https://docs.rs/structopt/0.3.21/structopt/#help-messages) but I am not sure if all users will get this.
> I was planning to elaborate more on this in the ASB docs instead of adding really long help texts to the commands.
> 
> Someone who wants to properly use these commands HAS to have a proper understanding of the protocol.
> I am not sure how much more education I can and want to pack into the help of these commands.
> 
> I can add a sentence along the lines of `Publishes the Bitcoin cancel transaction.` - but I expect a user to expect that because they understand the protocol prior to using these commands. Do you have some concrete ideas as in what to add?

Yeah something along those lines just so that it reads better in isolation:

`Publishes the Bitcoin cancel transaction. By default, the cancel timelock will be enforced. A confirmed cancel transaction enables refund and punish.`

bonomat (commented):
A few minor things. 
I think this is a good base for what we eventually want: `./asb manual-recovery cancel all` :)

swap/src/protocol/alice/cancel.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[error("The cancel timelock has not expired yet because the timelock has not expired. Please try again later")]
+    CancelTimelockNotExpiredYet,
+}
+
+pub async fn cancel(
+    swap_id: Uuid,
+    bitcoin_wallet: Arc<Wallet>,
+    db: Arc<Database>,
+    force: bool,
+) -> Result<Result<(Txid, AliceState), Error>> {
```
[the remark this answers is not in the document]
thomaseizinger: Was inspired by this: https://sled.rs/errors

swap/tests/alice_manually_punishes_after_bob_dead.rs
```diff
@@ -0,0 +1,91 @@
...
+                .await
+                .wait_until_confirmed_with(state3.cancel_timelock)
+                .await?;
+        } else {
+            panic!("Alice in unexpected state {}", alice_state);
+        }
+
+        // manual cancel (required to be able to punish)
+
+        ctx.restart_alice().await;
```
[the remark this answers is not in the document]
da-kami: We do not necessarily have to in some scenarios, but it mimics the actual behavior more correctly. 
When writing `e2e` tests for the manual commands I tend to mimic actually running the command, which involves starting the application in a real world scenario. We could handle this different in the tests, but I think the `e2e` test should be as close to the real world as they can be.

swap/src/bin/asb.rs
```diff
@@ -173,6 +174,67 @@ async fn main() -> Result<()> {
...
+                do_not_await_finality,
+            } => {
+                let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+
+                let (txid, _) = alice::redeem(
+                    swap_id,
+                    Arc::new(bitcoin_wallet),
+                    Arc::new(db),
+                    force,
+                    !do_not_await_finality,
```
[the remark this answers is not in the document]
da-kami: yeah... that is not optimal, I will see to model it as `enum` instead of the boolean value to make it more readable.
I will see to keep this as a `bool` flag on the structopt level though.

CHANGELOG.md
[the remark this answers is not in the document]
da-kami: I am not sure about this one yet, I would keep `--force` for now and re-visit it once we get feedback from users. In the end these commands are power²-user commands anyway

IMO adding different flags for each command is not very user-friendly neither. 
The idea is that `--force` circumvents all checks that CAN be circumvented because they are not a hard dependency. And that is the same for all commands. 

Hard dependency are dependencies on e.g. adapter signature publication,  e.g. for Alice's refund, Bob must have published refund as well, that cannot be circumvened)

swap/src/protocol/alice/state.rs
da-kami: I think we should tackle that once we actually tackle cleaning things up. For Bob we still have a lot of these functions around for Alice we don't.

Happy to tackle this in follow ups, but I don't see it as a blocker for merging this PR :)

swap/src/protocol/alice/cancel.rs
```diff
@@ -0,0 +1,76 @@
...
+    let state = db.get_state(swap_id)?.try_into_alice()?.into();
+
+    let (monero_wallet_restore_blockheight, transfer_proof, state3) = match state {
+
+        // In case no XMR has been locked, move to Safely Aborted
+        AliceState::Started { .. }
+        | AliceState::BtcLocked { .. } => bail!("Cannot cancel swap {} because it is in state {} where no XMR was locked. Try aborting instead.", swap_id, state),
```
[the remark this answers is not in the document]
da-kami: The idea was that I hint to use the `safely-abort` command here. But I don't want to literally put the command name here, because it might become outdated. 

I removed the hint `Try aborting instead.` now, because I think it is not really important and no matter what it might become outdated. 
A power user should be aware of this option anyway...

swap/src/protocol/alice/cancel.rs
```diff
@@ -0,0 +1,76 @@
...
+    if !force {
+        tracing::debug!(%swap_id, "Checking if cancel timelock is expired");
+
+        if let ExpiredTimelocks::None = state3.expired_timelocks(bitcoin_wallet.as_ref()).await? {
+            return Ok(Err(Error::CancelTimelockNotExpiredYet));
+        }
+    }
+
+    let txid = if let Ok(tx) = state3.check_for_tx_cancel(bitcoin_wallet.as_ref()).await {
+        tracing::debug!(%swap_id, "Cancel transaction has already been published");
```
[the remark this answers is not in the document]
da-kami: I included the `txid` in this log, note however:
This is not an error path, so it will actually be printed in main anyway. This `debug` print was just to see if we published it or Bob.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
da-kami: Because I could not figure out how to use `machtes!` to assert on a tuple-struct, i.e. this did not compile because the constructor is private:

```
assert!(matches!(result, bob::refund::SwapNotCancelledYet(_)));
```

Fixed it by making the field of the error public :)

```
#[derive(thiserror::Error, Debug, Clone, Copy)]
#[error("Cannot refund because swap {0} was not cancelled yet. Make sure to cancel the swap before trying to refund.")]
pub struct SwapNotCancelledYet(pub Uuid);
```

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired_force.rs
da-kami: I changed it to match on an arbitrary outer error like this:

```
        let result = bob::cancel(bob_swap.id, bob_swap.bitcoin_wallet, bob_swap.db, true)
            .await;
        assert!(matches!(result, Err(_)));
```

Hope that is more descriptive if it fails :)

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
da-kami: I feel this comment is completely independent of work done in this PR but does concern our tests in general. Let's clean this up in a follow up if we deem it important.

da-kami:
@bonomat @thomaseizinger all those separate commit were a bit painful when changing things so I squashed the feature (i.e. the manual commands) into one commit.

da-kami:
@bonomat 

> I think this is a good base for what we eventually want: `./asb manual-recovery cancel all` :)

I'd say that is what https://github.com/comit-network/xmr-btc-swap/pull/461 is for ;)
I see these commands more as last resort if there is a bug in code and then I'd assume it's OK to run them for one specific swap. But let's see how it goes.

CHANGELOG.md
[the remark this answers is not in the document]
thomaseizinger: You can't also really circumvent the timelock right? :)

What this shows is that we should likely not even make this check ourselves but just try and broadcast it and handle the specific errors better.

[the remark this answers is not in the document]
thomaseizinger: And then this flag just goes away for this command, even simpler!

swap/src/protocol/alice/state.rs
```diff
@@ -403,6 +404,55 @@ impl State3 {
         )
     }
 
+    pub async fn check_for_tx_cancel(
```
thomaseizinger, line 407: Happy to leave them for now to move forward but overall, there isn't really any need why this functions need to sit on `State3`. You are getting as much data / functionality off `self` than you do from `bitcoin_wallet`. Putting it on `bitcoin::Wallet` is equally strange, hence I think this could just be a helper function inside the `swap` module.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
thomaseizinger: Agreed, I just pointed it out because these tests make excessive use of these asserts :)

swap/src/protocol/alice/recovery/redeem.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl Finality {
+    pub fn from_bool(do_not_await_finality: bool) -> Self {
+        if do_not_await_finality {
+            Self::NotAwait
+        } else {
+            Self::Await
+        }
+    }
+}
```
thomaseizinger, line 21: :+1:

swap/src/protocol/alice/state.rs
```diff
@@ -453,6 +453,18 @@ impl State3 {
...
+    pub async fn punish_btc(&self, bitcoin_wallet: &bitcoin::Wallet) -> Result<Txid> {
+        let signed_tx_punish = self.signed_punish_transaction()?;
+
+        async {
+            let (txid, subscription) = bitcoin_wallet.broadcast(signed_tx_punish, "punish").await?;
+            subscription.wait_until_final().await?;
+
+            Result::<_, anyhow::Error>::Ok(txid)
+        }
+        .await
```
thomaseizinger, line 465: This async block doesn't add any functional value :)

swap/src/protocol/alice/recovery/punish.rs
```diff
@@ -0,0 +1,98 @@
+use crate::bitcoin::{self, ExpiredTimelocks, Txid};
+use crate::database::{Database, Swap};
+use crate::protocol::alice::AliceState;
+use anyhow::{bail, Result};
+use std::sync::Arc;
+use uuid::Uuid;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    // Errors indicating the the swap can *currently* not be refunded but might be later
```
bonomat, line 10: Nit:
```suggestion
    // Errors indicating the swap can *currently* not be refunded but might be later
```

swap/src/protocol/alice/recovery/punish.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[error("Cannot punish because swap is not in a cancelled state. Make sure to cancel the swap before trying to punish or use --force.")]
+    SwapNotCancelled,
+    #[error("The punish timelock has not expired yet because the timelock has not expired. Please try again later")]
```
bonomat, line 13: I think you wanted to say something different here: 
```suggestion
    #[error("Cannot punish because the punish timelock has not yet expired. Please try again later")]
```

swap/src/protocol/alice/recovery/refund.rs
```diff
@@ -0,0 +1,125 @@
...
+    #[error("Swap is not in a cancelled state. Make sure to cancel the swap before trying to refund or use --force.")]
+    SwapNotCancelled,
+    #[error(
+        "Counterparty {0} did not refund the BTC yet. You can try again later or try to punish."
+    )]
+    RefundTransactionNotPublishedYet(PeerId),
+
+    // Errors indicating that the swap cannot be refunded because because it is in a abort/final
+    // state
+    #[error("Swa is in state {0} where no XMR was locked. Try aborting instead.")]
```
bonomat, line 22:
```suggestion
    #[error("Swap is in state {0} where no XMR was locked. Try aborting instead.")]
```

swap/src/protocol/alice/state.rs
bonomat: Not a recommendation to do, just want to hear your feedback: 

This function is duplicated in `state3` for Alice and `state6` for Bob. 
How about converting this function into a free-floating one which takes as input a wallet and something that implements `.get_cancel()`, that way you can remove one and it can live somewhere else.

bonomat (approved):
good job, let's get this merged. A few nitpicks.

swap/src/protocol/alice/state.rs
da-kami: I would clean this stuff up in a follow up because I think a complete solution should get some more thought. I think we should find a better solution for all these function on both sides. 
If it allows re-using functionality on both sides - even better :)

da-kami:
https://github.com/comit-network/xmr-btc-swap/pull/452/commits/a5f8022f5d6ca682c98b5513e0626b8032d35d6f necessary after rebase on `master` (fee calculation needs the cancel tx to be accessible publicly)

swap/src/protocol/alice/state.rs
thomaseizinger: > How about converting this function into a free-floating one which takes as input a wallet and something that implements .get_cancel(), that way you can remove one and it can live somewhere else.

Abstracting over `get_cancel` would require a trait and the only benefit you are getting from that is that you can type:

```rust
check_for_tx_cancel(state, wallet)
```

instead of 


```rust
check_for_tx_cancel(state.tx_cancel(), wallet)
```

I would say it is probably not worth it to introduce a trait and having to do at least 2 implementations of it (I don't know how many different states we need to call this function from).

As a rule of thumb, abstracting over functionality with a trait is particularly useful if conditionals are involved. Let's say for example your function looks like this:

```rust
fn foo<T: Bar>(bar: &T, baz: Baz) -> u64 {
	let a_or_b = if baz.some_condition() {
		bar.compute_a()
	} else {
		bar.compute_b()
	};

	baz.compute_something(a_or_b)
}
```

In this case, having the trait is particularly useful because whether or not we want to call `compute_a` or `compute_b` depends on `Baz`. Hence, we cannot just "lift" this out and make it the caller's responsibility because we would have to lift the whole `if` block out which kind of makes the whole function pointless.

In our case, we always take `TxCancel` and pass it to the wallet, hence we can just have the caller do that and therefore avoid having to introduce a trait and several implementations at virtually no benefit.

It is usually good to strive for passing around _values_ (like `TxCancel`) and only use traits for real functionality that is not just a getter.

Issue #458, opened 2021-04-29
Remove logic from the eventloop


da-kami:
With https://github.com/comit-network/xmr-btc-swap/pull/451 we added yet another logic element to the event loop. This makes it hard to test and extend the code. 

We should decompose those elements into separate components (i.e. NetworkBehaviours) that handle checks on incoming messages and only bubble up events to the event loop when needed.

This also accounts for loading unfinished swaps upon startup.

da-kami:
Can get inspiration from extracting Alice's `spot_price` behaviour in #461 - this removed a good deal of logic from the eventloop and made it testable with network protocol level tests.

PR #461, opened 2021-04-29
Add resume-only mode for the ASB

### LINKED ISSUE #378: Resume-only mode - Don't allow new swap
In some scenarios we might want to resume swaps to recover funds but don't accept new incoming swaps.

Add a flag `--resume-only` (or similar) that prevents new, incoming swaps from being accepted.

Related tickets:

https://github.com/comit-network/xmr-btc-swap/issues/374
https://github.com/comit-network/xmr-btc-swap/issues/377

da-kami:
Fixes #378 

Resume-only is a maintenance mode where no swaps are accepted but unfinished swaps are resumed.
This is achieve by ignoring incoming spot-price requests (that would lead to execution setup) in the event-loop.

- [x] Refactor `spot_price`, move Alice's decision logic into dedicated `NetworkBehaviour`
- [x] Protocol (network) level tests for the `spot_price` behaviour

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   If you want to access data created by a previous version you will have to rename the data folder or one of the following:
   1. For the CLI you can use `--data-dir` to point to the old directory.
   2. For the ASB you can change the data-dir in the config file of the ASB.
+- The CLI receives proper Error messages if setting up a swap with the ASB fails.
+  This is a breaking change because the spot-price protocol response changed.
+  Expected errors scenarios that are now reported back to the CLI:
+  1. Balance of ASB too low
+  2. Buy amount sent by CLI exceeds maximum buy amount accepted by ASB
+  3. ASB is running in resume-only mode and does not accept incoming swap requests
 
 ## [0.5.0] - 2021-04-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Resume-only mode for the ASB.
+  When started with `--resume-only` the ASB does not accept new, incoming swap requests but only finishes swaps that are resumed upon startup.
+
 ### Fixed
 
 - An issue where both the ASB and the CLI point to the same default directory `xmr-btc-swap` for storing data.
--- a/swap/src/asb/command.rs
+++ b/swap/src/asb/command.rs
@@ -34,6 +34,12 @@ pub enum Command {
             default_value = "0.02"
         )]
         ask_spread: Decimal,
+
+        #[structopt(
+            long = "resume-only",
+            help = "For maintenance only. When set, no new swap requests will be accepted, but existing unfinished swaps will be resumed."
+        )]
+        resume_only: bool,
     },
     History,
     WithdrawBtc {
--- a/swap/src/network/spot_price.rs
+++ b/swap/src/network/spot_price.rs
@@ -41,7 +41,16 @@ pub struct Request {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Response {
-    pub xmr: monero::Amount,
+    pub xmr: Option<monero::Amount>,
+    pub error: Option<Error>,
+}
+
+#[derive(Clone, Debug, thiserror::Error, Serialize, Deserialize)]
+pub enum Error {
+    #[error(
+        "This seller currently does not accept incoming swap requests, please try again later"
+    )]
+    MaintenanceMode,
 }
 
 /// Constructs a new instance of the `spot-price` behaviour to be used by Alice.
--- a/swap/src/network/spot_price.rs
+++ b/swap/src/network/spot_price.rs
@@ -1,17 +1,12 @@
 use crate::monero;
 use crate::network::cbor_request_response::CborCodec;
-use crate::protocol::bob;
 use libp2p::core::ProtocolName;
-use libp2p::request_response::{
-    ProtocolSupport, RequestResponse, RequestResponseConfig, RequestResponseEvent,
-    RequestResponseMessage,
-};
-use libp2p::PeerId;
+use libp2p::request_response::{RequestResponse, RequestResponseEvent, RequestResponseMessage};
 use serde::{Deserialize, Serialize};
 
-const PROTOCOL: &str = "/comit/xmr/btc/spot-price/1.0.0";
+pub const PROTOCOL: &str = "/comit/xmr/btc/spot-price/1.0.0";
 pub type OutEvent = RequestResponseEvent<Request, Response>;
-type Message = RequestResponseMessage<Request, Response>;
+pub type Message = RequestResponseMessage<Request, Response>;
 
 pub type Behaviour = RequestResponse<CborCodec<SpotPriceProtocol, Request, Response>>;
 
@@ -45,55 +40,57 @@ pub enum Response {
     Error(Error),
 }
 
-#[derive(Clone, Debug, thiserror::Error, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum Error {
-    #[error(
-        "This seller currently does not accept incoming swap requests, please try again later"
-    )]
     NoSwapsAccepted,
-    #[error("Seller refused to buy {buy} because the maximum configured buy limit is {max}")]
     MaxBuyAmountExceeded {
         #[serde(with = "::bitcoin::util::amount::serde::as_sat")]
         max: bitcoin::Amount,
         #[serde(with = "::bitcoin::util::amount::serde::as_sat")]
         buy: bitcoin::Amount,
     },
-    #[error("This seller's XMR balance is currently too low to fulfill the swap request to buy {buy}, please try again later")]
     BalanceTooLow {
         #[serde(with = "::bitcoin::util::amount::serde::as_sat")]
         buy: bitcoin::Amount,
     },
-
     /// To be used for errors that cannot be explained on the CLI side (e.g.
     /// rate update problems on the seller side)
-    #[error("The seller encountered a problem, please try again later.")]
     Other,
 }
 
-/// Constructs a new instance of the `spot-price` behaviour to be used by Bob.
-///
-/// Bob only supports outbound connections, i.e. requesting a spot price for a
-/// given amount of BTC in XMR.
-pub fn bob() -> Behaviour {
-    Behaviour::new(
-        CborCodec::default(),
-        vec![(SpotPriceProtocol, ProtocolSupport::Outbound)],
-        RequestResponseConfig::default(),
-    )
-}
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::monero;
 
-impl From<(PeerId, Message)> for bob::OutEvent {
-    fn from((peer, message): (PeerId, Message)) -> Self {
-        match message {
-            Message::Request { .. } => Self::unexpected_request(peer),
-            Message::Response {
-                response,
-                request_id,
-            } => Self::SpotPriceReceived {
-                id: request_id,
-                response,
-            },
-        }
+    #[test]
+    fn snapshot_test_serialize() {
+        let amount = monero::Amount::from_piconero(100_000u64);
+        let xmr = r#"{"Xmr":100000}"#.to_string();
+        let serialized = serde_json::to_string(&Response::Xmr(amount)).unwrap();
+        assert_eq!(xmr, serialized);
+
+        let error = r#"{"Error":"NoSwapsAccepted"}"#.to_string();
+        let serialized = serde_json::to_string(&Response::Error(Error::NoSwapsAccepted)).unwrap();
+        assert_eq!(error, serialized);
+
+        let error = r#"{"Error":{"MaxBuyAmountExceeded":{"max":0,"buy":0}}}"#.to_string();
+        let serialized = serde_json::to_string(&Response::Error(Error::MaxBuyAmountExceeded {
+            max: Default::default(),
+            buy: Default::default(),
+        }))
+        .unwrap();
+        assert_eq!(error, serialized);
+
+        let error = r#"{"Error":{"BalanceTooLow":{"buy":0}}}"#.to_string();
+        let serialized = serde_json::to_string(&Response::Error(Error::BalanceTooLow {
+            buy: Default::default(),
+        }))
+        .unwrap();
+        assert_eq!(error, serialized);
+
+        let error = r#"{"Error":"Other"}"#.to_string();
+        let serialized = serde_json::to_string(&Response::Error(Error::Other)).unwrap();
+        assert_eq!(error, serialized);
     }
 }
-crate::impl_from_rr_event!(OutEvent, bob::OutEvent, PROTOCOL);
--- a/swap/src/network/spot_price.rs
+++ b/swap/src/network/spot_price.rs
@@ -1,6 +1,6 @@
 use crate::monero;
 use crate::network::cbor_request_response::CborCodec;
-use crate::protocol::{alice, bob};
+use crate::protocol::bob;
 use libp2p::core::ProtocolName;
 use libp2p::request_response::{
     ProtocolSupport, RequestResponse, RequestResponseConfig, RequestResponseEvent,
@@ -10,7 +10,7 @@ use libp2p::PeerId;
 use serde::{Deserialize, Serialize};
 
 const PROTOCOL: &str = "/comit/xmr/btc/spot-price/1.0.0";
-type OutEvent = RequestResponseEvent<Request, Response>;
+pub type OutEvent = RequestResponseEvent<Request, Response>;
 type Message = RequestResponseMessage<Request, Response>;
 
 pub type Behaviour = RequestResponse<CborCodec<SpotPriceProtocol, Request, Response>>;
@@ -50,7 +50,7 @@ pub enum Error {
     #[error(
         "This seller currently does not accept incoming swap requests, please try again later"
     )]
-    MaintenanceMode,
+    NoSwapsAccepted,
     #[error("Seller refused to buy {buy} because the maximum configured buy limit is {max}")]
     MaxBuyAmountExceeded {
         #[serde(with = "::bitcoin::util::amount::serde::as_sat")]
@@ -63,18 +63,11 @@ pub enum Error {
         #[serde(with = "::bitcoin::util::amount::serde::as_sat")]
         buy: bitcoin::Amount,
     },
-}
 
-/// Constructs a new instance of the `spot-price` behaviour to be used by Alice.
-///
-/// Alice only supports inbound connections, i.e. providing spot prices for BTC
-/// in XMR.
-pub fn alice() -> Behaviour {
-    Behaviour::new(
-        CborCodec::default(),
-        vec![(SpotPriceProtocol, ProtocolSupport::Inbound)],
-        RequestResponseConfig::default(),
-    )
+    /// To be used for errors that cannot be explained on the CLI side (e.g.
+    /// rate update problems on the seller side)
+    #[error("The seller encountered a problem, please try again later.")]
+    Other,
 }
 
 /// Constructs a new instance of the `spot-price` behaviour to be used by Bob.
@@ -88,22 +81,6 @@ pub fn bob() -> Behaviour {
         RequestResponseConfig::default(),
     )
 }
-
-impl From<(PeerId, Message)> for alice::OutEvent {
-    fn from((peer, message): (PeerId, Message)) -> Self {
-        match message {
-            Message::Request {
-                request, channel, ..
-            } => Self::SpotPriceRequested {
-                request,
-                channel,
-                peer,
-            },
-            Message::Response { .. } => Self::unexpected_response(peer),
-        }
-    }
-}
-crate::impl_from_rr_event!(OutEvent, alice::OutEvent, PROTOCOL);
 
 impl From<(PeerId, Message)> for bob::OutEvent {
     fn from((peer, message): (PeerId, Message)) -> Self {
--- a/swap/src/network/spot_price.rs
+++ b/swap/src/network/spot_price.rs
@@ -1,6 +1,6 @@
+use crate::monero;
 use crate::network::cbor_request_response::CborCodec;
 use crate::protocol::{alice, bob};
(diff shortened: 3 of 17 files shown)


da-kami:
@thomaseizinger I think this is another candidate that falls into https://github.com/comit-network/xmr-btc-swap/issues/458

Would you have some time for a pair programming session these days where we change the `Behavior`s to specific impls for Alice and move the logic from the event loop there?

I started working on it, but some help for ramping things up would be great - should not take too long :)

swap/src/asb/command.rs
```diff
@@ -34,6 +34,12 @@ pub enum Command {
             default_value = "0.02"
         )]
         ask_spread: Decimal,
+
+        #[structopt(
+            long = "resume-only",
+            help = "For maintenance only. When set no new swap requests will be accepted, but existing unfinished swaps will be resumed."
```
thomaseizinger, line 40:
```suggestion
            help = "For maintenance only. When set, no new swap requests will be accepted, but existing unfinished swaps will be resumed."
```

swap/src/protocol/alice/event_loop.rs
```diff
@@ -144,6 +148,11 @@ where
                 swarm_event = self.swarm.next_event() => {
                     match swarm_event {
                         SwarmEvent::Behaviour(OutEvent::SpotPriceRequested { request, channel, peer }) => {
+                            if self.resume_only {
+                                tracing::warn!(%peer, "Ignoring spot price request from {} because ASB started in resume-only mode", peer);
+                                continue;
```
thomaseizinger, line 153: We should return a proper error the CLI here instead of just dropped the channel.

thomaseizinger (commented):
> Would you have some time for a pair programming session these days where we change the Behaviors to specific impls for Alice and move the logic from the event loop there?

> I started working on it, but some help for ramping things up would be great - should not take too long :)

Happy to review a draft of what you've got. Extracting a Behaviour is not that difficult, I am confident you can manage by yourself :)

swap/src/protocol/alice/event_loop.rs
da-kami: Yeah, we should handle all `expected` errors properly and return a proper message, see https://github.com/comit-network/xmr-btc-swap/pull/461/commits/84626d654f8c904a5588a9fdcd2782f4d0a8d4bc

da-kami:
> > Would you have some time for a pair programming session these days where we change the Behaviors to specific impls for Alice and move the logic from the event loop there?
> 
> > I started working on it, but some help for ramping things up would be great - should not take too long :)
> 
> Happy to review a draft of what you've got. Extracting a Behaviour is not that difficult, I am confident you can manage by yourself :)

Update: I'm still focusing on getting features in. I did not keep my trial stuff around from yesterday. 
I think today's changes are an improvement that should merge. Will see to put a Draft PR together for #458 when I find the time.

da-kami:
Do you think it's worth to add more e2e tests that test these error scenarios?

I could make the test setup more configurable so we can 
- start Alice without funds 
- start Alice in maintenance mode
- start Bob with higher Amount than Alice's limit

thomaseizinger:
> @thomaseizinger @bonomat Do you think it's worth to add more e2e tests that test these error scenarios?

Please no more e2e tests unless absolutely necessary. Once more things are extracting into dedicated modules, writing focused tests should be a lot easier and still eliminate a lot of risk.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -146,14 +149,28 @@ where
...
+                                            Ok(_) => {
+                                                continue;
+                                            },
+                                            Err(_) => {
+                                                tracing::debug!(%peer, "Failed to respond with error to spot price request");
+                                                continue;
+                                            }
+                                        }
+                                    }
+                                },
```
bonomat, line 166: I think this can be optimized to improve readability, e.g. 
```Rust
match xmr {
 Ok(Ok(xmr) => ...
 Ok(Err(e)  => ...
 Err(e) => ... 
}
```

swap/src/protocol/alice/event_loop.rs
```diff
@@ -283,30 +300,32 @@ where
         &mut self,
         btc: bitcoin::Amount,
         monero_wallet: Arc<monero::Wallet>,
-    ) -> Result<monero::Amount> {
+    ) -> Result<Result<monero::Amount, spot_price::Error>> {
+        if self.resume_only {
+            return Ok(Err(spot_price::Error::MaintenanceMode));
+        }
```
bonomat, line 306: Why does this have to be an `Result<Result<...>`?

swap/src/protocol/alice/event_loop.rs
```diff
@@ -148,25 +147,25 @@ where
...
+                                    Ok(xmr) => xmr,
+                                    Err(e) => {
+                                        tracing::warn!(%peer, "Ignoring spot price request from {} because: {:#}", peer, e);
+                                        match self.swarm.behaviour_mut().spot_price.send_response(channel, spot_price::Response { xmr: None, error: Some(e) }) {
+                                            Ok(_) => {
+                                                continue;
+                                            },
+                                            Err(_) => {
+                                                tracing::debug!(%peer, "Failed to respond with error to spot price request");
+                                                continue;
```
bonomat, line 162: Maybe this is a bit nicer: 
```Rust
let result = self.swarm.behaviour_mut().spot_price.send_response(channel, spot_price::Response { xmr: None, error: Some(e) });
if result.is_err() {
     tracing::debug!(%peer, "Failed to respond with error to spot price request");
};
continue;
```

swap/src/protocol/alice/event_loop.rs
```diff
@@ -301,30 +300,32 @@ where
         &mut self,
         btc: bitcoin::Amount,
         monero_wallet: Arc<monero::Wallet>,
-    ) -> Result<monero::Amount> {
+    ) -> Result<Result<monero::Amount, spot_price::Error>> {
+        if self.resume_only {
+            return Ok(Err(spot_price::Error::MaintenanceMode));
+        }
```
bonomat, line 306: I find this approach a bit weird, i.e. return an error to indicate that we don't care about the price requests. 
Why did you move the `self.resume_only` out of the loop?

swap/src/protocol/alice/event_loop.rs
```diff
@@ -148,25 +147,25 @@ where
                 swarm_event = self.swarm.next_event() => {
                     match swarm_event {
                         SwarmEvent::Behaviour(OutEvent::SpotPriceRequested { request, channel, peer }) => {
-                            if self.resume_only {
```
bonomat: I'd handle the resume-only case here in the loop instead of further down in call-chain.

bonomat (commented):
Added some ideas of how I would do it differently. 

Regarding your question on an additional e2e tests: I'm with @thomaseizinger : ideally we don't need additional e2e tests 😬

swap/src/protocol/alice/event_loop.rs
thomaseizinger: image

Definitely should get rid of this kamehameha!

swap/src/network/spot_price.rs
```diff
@@ -41,7 +41,28 @@ pub struct Request {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Response {
-    pub xmr: monero::Amount,
+    pub xmr: Option<monero::Amount>,
+    pub error: Option<Error>,
```
thomaseizinger, line 45: These are mutually exclusive, why is this not an enum?

swap/src/network/spot_price.rs
```diff
@@ -41,7 +41,28 @@ pub struct Request {
...
+}
+
+#[derive(Clone, Debug, thiserror::Error, Serialize, Deserialize)]
```
thomaseizinger, line 48: Deriving `Error` on a struct that is part of the wire protocol is kind of weird, can we separate this please so that we have a clearly defined, external API?

swap/src/protocol/alice/event_loop.rs
thomaseizinger: An even better approach would be to extract your own network behaviour for all of this functionality. There is a lot of untested functionality happening here.

swap/src/protocol/bob/event_loop.rs
```diff
@@ -250,11 +250,22 @@ impl EventLoopHandle {
...
+
+        match (response.xmr, response.error) {
+            (Some(xmr), None) => Ok(xmr),
+            (_, Some(error)) => {
+                bail!(error);
+            }
+            (None, None) => {
+                bail!(
+                    "Unexpected response for spot-price request, neither price nor error received"
+                );
```
thomaseizinger, line 266: Does it not bother you that you have to write this code? :D

If you model your response differently, this can be avoided entirely.

swap/tests/harness/mod.rs
```diff
@@ -234,6 +234,7 @@ fn start_alice(
         db,
         FixedRate::default(),
         bitcoin::Amount::ONE_BTC,
+        false,
```
thomaseizinger, line 237: Without IDE type hints, it is practically impossible to know what this boolean does. Can we please model this in a clearer way?

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  When started with `--resume-only` the ASB does not accept new, incoming swap requests but only finishes swaps that are resumed upon startup.
+
+### Fixed
+
+- The ASB returns proper error messages for expected errors upon starting a swap.
+  This is a breaking change because the spot-price protocol response changed.
+  Expected errors scenarios that are now reported back to the CLI:
+  1. Balance of ASB too low
+  2. Buy amount sent by CLI exceeds maximum buy amount accepted by ASB
+  3. ASB is running in maintenance mode and does not accept incoming swap requests
```
thomaseizinger, line 22: What do you think about collapsing these? The added "maintenance mode" motivated adding errors to the network layer so I think we could mentioned them as one change.

thomaseizinger (changes requested):
My main feedback is that we have _a lot_ of untested conditionals in the eventloop now. That, plus the fact that we can't use `?` to simplify this control flow because we are in the eventloop warrants creating your own `NetworkBehaviour` IMO where you can isolate all this logic and test it individually.

swap/src/protocol/alice/event_loop.rs
da-kami: To keep all logic related to `spot-price` requests in a dedicated function and not scatter them over different places (i.e. event loop vs. this function. The function is called `handle_spot_price_request` and this check is part of that I'd say. 
This should help us move this logic into a dedicated NetworkBehaviour at some point.

swap/src/network/spot_price.rs
da-kami: All these errors are related to the `spot_price` protocol that's why I moved them into the protocol. I think the dedicated `NetworkBehaviour` would solve all the problems that came up (then the Error could also be defined there). I will see to work on that.

swap/src/protocol/alice/event_loop.rs
da-kami: > An even better approach would be to extract your own network behaviour for all of this functionality. There is a lot of untested functionality happening here.

Done :)
Swarm tests not added yet, will jump on that next.

swap/src/protocol/alice/event_loop.rs
da-kami: Note: This comment became somewhat outdated with https://github.com/comit-network/xmr-btc-swap/pull/461/commits/24da186be998518eeb94db4ef61eb3f8c6f3b904

swap/src/protocol/alice/event_loop.rs
da-kami: The resume-only case only concerns the spot price protocol and not the complete eventloop.
Now that we have a dedicated `NetworkBehaviour` for Alice the event loop is not even aware of the resume-only mode anymore :)
Note that we can rename the flag to "don't accept swaps` or something similar if you prefer - so the name fits better to the context of the spot-price protocol.

swap/src/protocol/bob/event_loop.rs
da-kami: 😅 it was late at night

swap/src/network/spot_price.rs
```diff
@@ -40,20 +40,34 @@ pub struct Request {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct Response {
-    pub xmr: monero::Amount,
+pub enum Response {
+    Xmr(monero::Amount),
+    Error(Error),
 }
```
thomaseizinger, line 46: Can we have a form of snapshot test for this to make sure we don't accidentially break the wire-format? (i.e. serialize this and test against a hardcoded serialization).

Depending on your familiarity with serde, I would recommend that in general so you have a mental model of what the data you are sending actually looks like.

swap/src/protocol/alice/spot_price.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        let (request, channel) = match message {
+            RequestResponseMessage::Request {
+                request, channel, ..
+            } => (request, channel),
+            RequestResponseMessage::Response { .. } => {
+                tracing::error!("Unexpected message");
+                return;
+            }
+        };
```
thomaseizinger, line 135: This can be collapsed into the above.

swap/src/protocol/alice/spot_price.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        if self.resume_only {
+            tracing::warn!(%peer, "Ignoring spot price request from {} because ASB is running in maintenance mode", peer);
+            self.handle_error_response(peer, channel, spot_price::Error::MaintenanceMode);
```
thomaseizinger, line 139: Half the codebase calls this `resume-only`, the other half "MaintainanceMode". Find a single name for it? :)

swap/src/protocol/alice/spot_price.rs
```diff
@@ -0,0 +1,201 @@
...
+            latest_rate,
+            resume_only,
+        }
+    }
+
+    pub fn update_balance(&mut self, balance: monero::Amount) {
+        self.balance = balance;
+    }
+
+    fn handle_error_response(
```
thomaseizinger, line 77:
```suggestion
    fn send_error_response(
```

swap/src/protocol/alice/spot_price.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        if self.resume_only {
+            tracing::warn!(%peer, "Ignoring spot price request from {} because ASB is running in maintenance mode", peer);
+            self.handle_error_response(peer, channel, spot_price::Error::MaintenanceMode);
+            return;
+        }
+
+        let btc = request.btc;
+        if btc > self.max_buy {
+            tracing::warn!(%peer, "Ignoring spot price request from {} because max muy amount exceeded", peer);
```
thomaseizinger, line 145: If you attach this log message to the error variant, you could have a generic log inside the function.

swap/src/protocol/bob/event_loop.rs
```diff
@@ -250,11 +251,17 @@ impl EventLoopHandle {
...
-            .await?
-            .xmr)
+            .await?;
+
+        match response {
+            Response::Xmr(xmr) => Ok(xmr),
+            Response::Error(error) => {
+                bail!(error);
+            }
+        }
```
thomaseizinger, line 264: That is equivalent to `?` if you would use `Result` instead of `Response`. Not sure if such a good idea though :)

swap/src/network/spot_price.rs
thomaseizinger: I agree with all you said. What I am saying is: we may want to have separate structs for what is _serialized_ vs what is interally used to pass around errors. Usually, you'd only implement `std::error::Error` for things that go into a `Result` and where you use `?`.

thomaseizinger (commented):
Nice work!

Some minor feedback.

swap/src/protocol/alice/spot_price.rs
da-kami: I thought we don't like kamehamehas... 😅

swap/src/protocol/alice/spot_price.rs
da-kami: There is actually some reasoning behind it - from the CLI's perspective the ASB is in maintenance mode that's what one cannot trade with it, but from the ASB provider point of view it is `resume-only` mode - other things could lead to maintenance mode as well. 

But that seems to be confusing, so maybe I just call it resume-only then because that is what it is at the moment...

swap/src/protocol/alice/spot_price.rs
da-kami: I thought about that but could not come up with an easy way to use the `peer` field as context and logging on different levels. Am I missing something, can this be achieved in an easy way?
I did look into the `tracing::Event` and potentially passing events around and then logging them, but I think that makes our life harder than it has to be.

thomaseizinger: > I thought about that but could not come up with an easy way to use the peer field as context and logging on different levels if I move it in. Am I missing something, can this be achieved in an easy way?

You always have a peer right? You can pass it as a standalone parameter together with the error and then always supply it as context and simply log the error with `{:#}`.

swap/src/protocol/alice/spot_price.rs
thomaseizinger: :joy: 

Fair enough :)

swap/src/network/spot_price.rs
da-kami: I would have to write these tests against a `cbor` snapshot given that we use `cbor` encoding, correct?

CHANGELOG.md
da-kami: I gave it a try and find it better separate, mainly because one only concerns the ASB and the other concerns mostly the CLI. 
The fact that we introduced the error messages as part of the `resume-only` mode is a technicality that is irrelevant for users.

swap/src/protocol/alice/spot_price.rs
da-kami: But I would not be able to adapt the level... some logs are `warn` and other are `error` or am I missing something?

swap/src/protocol/alice/spot_price.rs
```diff
@@ -197,3 +207,43 @@ impl From<OutEvent> for alice::OutEvent {
...
+                spot_price::Error::Other
+            }
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error<LR>
+where
+    LR: LatestRate + Debug,
```
thomaseizinger, line 232: As a rule of thumb, try to avoid trait bounds on _type_ declarations. Traits define functionality and hence should be coupled to functionality (`impl` blocks) where possible.

There is two ways around this for this situation:

1. Use a trait object (`Box<dyn std::error::Error>`)
2. Make the `Error` generic over `E` instead of `LR::Error`, i.e.: `LatestRateFetchFailed(E)`

Both approaches will avoid cascading changes like `+ Debug` on the `LR` type parameter.

swap/src/protocol/bob/event_loop.rs
```diff
@@ -270,6 +271,7 @@ impl EventLoopHandle {
         match response {
             Response::Xmr(xmr) => Ok(xmr),
             Response::Error(error) => {
+                let error: bob::spot_price::Error = error.into();
```
thomaseizinger, line 274: I wonder how we could organize our modules such that we don't need to over-qualify the type here given that we already import another module named `spot_price` and we are in a module scoped to `bob` so `bob::spot_price` is pretty redundant.

thomaseizinger (approved):
Thanks!

Left two nits, can easily be tackled in a follow-up though.

swap/src/protocol/bob/event_loop.rs
da-kami: One thing that came to my mind is, that this network related stuff should not be under `protocol` anyway!
I think we should introduce `alice` and `bob` in the network module instead and then decide if something is global, alice or bob on the network level, independent of the protocol.

Will see to give this more thought and tackle this in follow-up(s)

da-kami:
@thomaseizinger @bonomat  network level tests are done with https://github.com/comit-network/xmr-btc-swap/pull/461/commits/36651fd45ee9cc1c48a3f9bbcbb56dfc87e8f76b

Turned out pretty neat! 

If there is nothing mayor it would be great if we could merge this soon, so for the final review please focus on functionality and potential critical flaws. Happy about nits on spelling & co as well, but might fix nits on style in follow ups :)

I am planning to do the module cleanup (the tests just being in `alice::spot_price` is somewhat not optimal neither, but I would fix that together with moving stuff between modules) and the problem with implementing `PartialEq` (unless there is an easy straight-forward solution that I missed) as follow up :)

swap/src/protocol/alice/spot_price.rs
```diff
@@ -244,3 +245,386 @@ impl Error {
...
+        .await;
+
+        let btc_to_swap = bitcoin::Amount::from_btc(0.01).unwrap();
+
+        let request = spot_price::Request { btc: btc_to_swap };
+
+        test.send_request(request);
+        test.assert_error(
+            alice::spot_price::Error::BalanceTooLow { buy: btc_to_swap },
+            bob::spot_price::Error::BalanceTooLow { buy: btc_to_swap },
```
thomaseizinger, line 300: Seeing this now, we may want to think about whether or not we actually want to share this information with the CLI. Could be a privacy concern to let people know that we are running low on balance? In the end the user has to wait and try again later so maybe an error with just "try again later" could be sufficient?

Just some food for thought, I wouldn't change it now :)

thomaseizinger (approved):
Some seriously impressive tests! Very nicely done. Thanks for putting this effort in!

Easy to read due to good structure and extraction of setup and assertions into helpers :clap:

swap/src/protocol/alice/spot_price.rs
da-kami: I had the same thought even when I implemented it already. But I felt that it is actually fine to report it back to the user because I could not find an issue where reporting being low on balance is a privacy concern. 
Could you come up with any real concern for privacy just from knowing that the ASB provider is low on balance? 

IMO the user cannot exploit anything with this information, and given that the user has a communication channel with the ASB provider he can even contact the ASB provider with this information.

thomaseizinger: I can't think of anything concrete now. It is more of a gut-feeling in regards to security. In a targeted attack, any bit of information can be helpful to an attacker. Maybe there is a vulnerability in `monero-wallet-rpc` somewhere that only works on low balances? :D

bonomat (approved):
Awesome. Nothing to add.

PR #466, opened 2021-05-02
Add support to dynamically chose bitcoin transaction fees. 

### LINKED ISSUE #443: Mainnet-ready fees for Bitcoin
As of now, we are using static fees: 
See: https://github.com/comit-network/xmr-btc-swap/blob/52b9a78de2076ec8517b4290f7291678e609d251/swap/src/bitcoin.rs#L41

We should move to dynamic fees depending on the current network. 
It is necessary that both parties agree on the same fee. 

Alice (as the maker) should propose the fees.

bonomat:
Resolves #443 (Bitcoin only). 

@da-kami  /@thomaseizinger: do you know how fees work in Monero? 
If not, I'll read up on it and to come up with a proper strategy. 

Monero Fees will be covered in #470

Note: 
The there is a hardcoded relative and absolute upper bound. I personally feel safer if this is hardcoded for now as it is too easy to get wrong. 
At the time of writing, this equals roughly USD $56. 
Eventually we will want to make this also configurable.

### DIFF
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -37,18 +37,10 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
+pub use crate::bitcoin::cancel::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_CANCEL;
 pub use crate::bitcoin::punish::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_PUNISH;
 pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REDEEM;
 pub use crate::bitcoin::refund::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REFUND;
-
-// TODO: Configurable tx-fee (note: parties have to agree prior to swapping)
-// Current reasoning:
-// tx with largest weight (as determined by get_weight() upon broadcast in e2e
-// test) = 609 assuming segwit and 60 sat/vB:
-// (609 / 4) * 60 (sat/vB) = 9135 sats
-// Recommended: Overpay a bit to ensure we don't have to wait too long for test
-// runs.
-pub const TX_FEE: u64 = 15_000;
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 pub struct SecretKey {
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -37,6 +37,8 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
+pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REDEEM;
+
 // TODO: Configurable tx-fee (note: parties have to agree prior to swapping)
 // Current reasoning:
 // tx with largest weight (as determined by get_weight() upon broadcast in e2e
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -37,10 +37,10 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
-pub use crate::bitcoin::cancel::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_CANCEL;
-pub use crate::bitcoin::punish::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_PUNISH;
-pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REDEEM;
-pub use crate::bitcoin::refund::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REFUND;
+pub use crate::bitcoin::cancel::ESTIMATED_WEIGHT as TX_CANCEL_ESTIMATED_WEIGHT;
+pub use crate::bitcoin::punish::ESTIMATED_WEIGHT as TX_PUNISH_ESTIMATED_WEIGHT;
+pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as TX_REDEEM_ESTIMATED_WEIGHT;
+pub use crate::bitcoin::refund::ESTIMATED_WEIGHT as TX_REFUND_ESTIMATED_WEIGHT;
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 pub struct SecretKey {
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -37,7 +37,9 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
+pub use crate::bitcoin::punish::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_PUNISH;
 pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REDEEM;
+pub use crate::bitcoin::refund::ESTIMATED_WEIGHT as ESTIMATED_WEIGHT_TX_REFUND;
 
 // TODO: Configurable tx-fee (note: parties have to agree prior to swapping)
 // Current reasoning:
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -37,11 +37,6 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use std::str::FromStr;
 
-pub use crate::bitcoin::cancel::ESTIMATED_WEIGHT as TX_CANCEL_ESTIMATED_WEIGHT;
-pub use crate::bitcoin::punish::ESTIMATED_WEIGHT as TX_PUNISH_ESTIMATED_WEIGHT;
-pub use crate::bitcoin::redeem::ESTIMATED_WEIGHT as TX_REDEEM_ESTIMATED_WEIGHT;
-pub use crate::bitcoin::refund::ESTIMATED_WEIGHT as TX_REFUND_ESTIMATED_WEIGHT;
-
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 pub struct SecretKey {
     inner: Scalar,
@@ -259,6 +254,12 @@ pub struct NotThreeWitnesses(usize);
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::bitcoin::wallet::EstimateFeeRate;
+    use crate::env::{GetConfig, Regtest};
+    use crate::protocol::{alice, bob};
+    use bdk::FeeRate;
+    use rand::rngs::OsRng;
+    use uuid::Uuid;
 
     #[test]
     fn lock_confirmations_le_to_cancel_timelock_no_timelock_expired() {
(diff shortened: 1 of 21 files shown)


bonomat:
I didn't run all e2e tests locally, so please hold back with your review until then as this could mean quite some changes.

Converted back to _draft_ to indicate this.

thomaseizinger:
> @da-kami /@thomaseizinger: do you know how fees work in Monero?

I don't, sorry.

swap/src/bitcoin/punish.rs
```diff
@@ -1,12 +1,16 @@
...
+use crate::bitcoin::{self, Address, Amount, PunishTimelock, Transaction, TxCancel, Txid};
 use ::bitcoin::util::bip143::SigHashCache;
 use ::bitcoin::{SigHash, SigHashType};
 use anyhow::{Context, Result};
 use bdk::bitcoin::Script;
 use miniscript::{Descriptor, DescriptorTrait};
 use std::collections::HashMap;
 
+// Taken from https://mempool.space/testnet/tx/ed4d60bc1fd172feca444ed3d06cccb90346b9098c2d28d2d034dac66f608f68
+// The weight might fluctuate slightly in reality.
```
da-kami, line 11: ❓ I find this comment a bit confusing - how would the weight of a transaction fluctuate? 
I thought the weight should always be the same, assuming that we always construct the transaction in the same way.

da-kami:
> > @da-kami /@thomaseizinger: do you know how fees work in Monero?
> 
> I don't, sorry.

I don't have much knowledge about it neither, better read up on it please

swap/src/bitcoin/punish.rs
bonomat: >  I thought the weight should always be the same, assuming that we always construct the transaction in the same way.

That's the point: unfortunately we don't construct our transactions always in the same way. Every byte counts, e.g. a bigger amount, a different timeouts, etc. maybe even spending from a transaction on `vout=0` vs `vout=1` makes a difference :) 

Anyway, I had a chat with @thomaseizinger about it and these constants will go away. 
Wait for these changes please.

bonomat:
@thomaseizinger : testing that the constants do not go out of sync was not that straight forward. Can I have an ACK on the concept of the last commit please?

swap/src/bitcoin/wallet.rs
```diff
@@ -344,13 +345,20 @@ where
...
                 sats_per_vbyte,
                 min_relay_fee.as_sat()
             );
             Ok(min_relay_fee)
+        } else if sats_per_vbyte > MAX_TX_FEE {
+            tracing::warn!(
+                "Hard bound of max transaction fees reached. Falling back to: {} sats",
+                MAX_TX_FEE
+            );
+            Ok(bitcoin::Amount::from_sat(MAX_TX_FEE))
```
bonomat, line 361: Note: upper bound could also be defined as a percentage of the amount being swapped.

swap/src/bitcoin/wallet.rs
```diff
@@ -22,7 +22,12 @@ use std::time::{Duration, Instant};
 use tokio::sync::{watch, Mutex};
 
 const SLED_TREE_NAME: &str = "default_tree";
-const MAX_TX_FEE: u64 = 100_000;
+
+// TODO make these values configurable
+/// Assuming we add a spread of 3% we don't want to pay more than 3% of the
+/// amount for tx fees.
+const MAX_RELATIVE_TX_FEE: f64 = 0.03;
+const MAX_ABSOLUTE_TX_FEE: u64 = 100_000;
```
bonomat, line 30: To keep the configuration small and reduce the risk of errors I propose to keep this hardcoded for the time being. 
It's too easy to get wrong and can lead to losing a lot of money.

swap/src/cli/command.rs
```diff
@@ -53,6 +56,9 @@ pub enum Command {
 
         #[structopt(long = "tor-socks5-port", help = "Your local Tor socks5 proxy port", default_value = DEFAULT_TOR_SOCKS5_PORT)]
         tor_socks5_port: u16,
+
+        #[structopt(long = "bitcoin-target-block", help = "Within how many blocks should the Bitcoin transactions be confirmed.", default_value = DEFAUL_BITCOIN_CONFIRMATION_TARGET)]
```
da-kami, line 60: Since this is suplicated for multiple commands you might want to extract that into a separate struct and use [`flatten`](https://docs.rs/structopt/0.3.21/structopt/#flattening).

da-kami: (same accounts for some other params already in there - we can also clean this up in a follow-up)

swap/src/protocol/alice/event_loop.rs
```diff
@@ -161,8 +161,38 @@ where
...
+                            let (tx_redeem_fee, tx_punish_fee, redeem_address, punish_address) = match (
+                                tx_redeem_fee,
+                                tx_punish_fee,
+                                redeem_address,
+                                punish_address,
+                            ) {
+                                (Ok(tx_redeem_fee), Ok(tx_punish_fee), Ok(redeem_address), Ok(punish_address)) => {
+                                    (tx_redeem_fee, tx_punish_fee, redeem_address, punish_address)
+                                }
+                                _ => { continue; }
```
da-kami, line 183: 🛠 Please **log an error** message! we should not just swallow the error here!

swap/src/protocol/alice/event_loop.rs
```diff
@@ -161,8 +161,38 @@ where
                                     continue;
                                 }
                             }
+                            // TODO: This should be cleaned up.
```
da-kami, line 164: Note: 
In my PR [1] that introduced a dedicated spot_price behaviour for Alice I still construct `State0` outside. We could move that inside the spot_price behaviour and just return `State0` directly. I will see to tackle this once we have one of them in and go for rebase :)

[1] https://github.com/comit-network/xmr-btc-swap/pull/461/files#diff-ac04e254d84094077194f78ec27926a36a2ceb76e65b40d719e7d9ab5b547254

da-kami (approved):
Great work! I did not find flaws in the concept, but added a few suggestions.

One thing that should be fixed before merging is not logging an error in the event loop if we fail to get fees/addresses.

swap/src/bitcoin.rs
```diff
@@ -308,4 +305,126 @@ mod tests {
...
+            TxRefund::weight(),
+            "TxRefund",
+        );
+    }
+
+    // Weights fluctuate -+2 wu assuming because of the length of the signatures.
+    fn assert_weight(is_weight: usize, expected_weight: usize, tx_name: &str) {
+        assert!(
+            is_weight + 1 == expected_weight
+                || is_weight + 2 == expected_weight
```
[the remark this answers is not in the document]
bonomat: My assumption is that it's because we have 2 signatures on the stack and each can fluctuate by 1 byte. 
Can add this comment.

swap/src/bitcoin/wallet.rs
```diff
@@ -662,4 +774,76 @@ mod tests {
 
         assert_eq!(confirmed.depth, 0)
     }
+
+    #[test]
+    fn given_one_BTC_and_100k_sats_per_vb_fees_should_not_hit_max() {
```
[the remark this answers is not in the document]
bonomat: Added some proptests.

swap/src/cli/command.rs
bonomat: Created a follow-up issue: #478

swap/src/protocol/alice/event_loop.rs
da-kami: I gave this a bit more thought. Network behaviours are not async, so I don't think we can actually move that piece of code into . 
Sidenote: We use an async-await network behaviour for the `execution_setup` and we could move the construction of `State0` in there, but I am not sure that is a good design. The `execution_setup` behaviour currently purely focuses on the message exchange and expects `State0` from the outside. I am not sure it should be concerned about the construction of that state.

Thus, at the moment I don't see a good way to move interaction with the wallet out of the eventloop. 
Given the current design I think it will have to stay similar to what it is now, so you could actually remove the ToDo.

bonomat: Interesting. 
Yea, the fee-estimation should not be moved outside the loop because we should only compute it when needed. I'll remove the todo and wait for your PR to introduce the new behavior. Maybe it becomes obvious what can be done here.

da-kami:
Seems you still have some dangling weights there...
Did you actually check that we always use the lower bound in the constant?

test logs 
```
thread 'bitcoin::tests::calculate_transaction_weights' panicked at 'TxPunish to have weight 547, but was 548', swap/src/bitcoin.rs:422:9
```

bonomat:
> Seems you still have some dangling weights there...
> Did you actually check that we always use the lower bound in the constant?
> 
> test logs
> 
> ```
> thread 'bitcoin::tests::calculate_transaction_weights' panicked at 'TxPunish to have weight 547, but was 548', swap/src/bitcoin.rs:422:9
> ```
> 
> would be nice if we can avoid flaky tests 😬

Seems like I missed one: Fixed and force-pushed. I think it was a somehow understandable error: the bound was an uneven number but should have been an even one if it can fluctuate by +2.

swap/src/cli/command.rs
```diff
@@ -15,6 +15,9 @@ const DEFAULT_ELECTRUM_RPC_URL: &str = "ssl://electrum.blockstream.info:60002";
 
 const DEFAULT_TOR_SOCKS5_PORT: &str = "9050";
 
+// Bitcoin transactions should be confirmed within X blocks
+const DEFAUL_BITCOIN_CONFIRMATION_TARGET: &str = "3";
```
da-kami, line 19:
```suggestion
const DEFAULT_BITCOIN_CONFIRMATION_TARGET: &str = "3";
```

missed that one before, can do in follow up cleanup ;)

da-kami (approved):

swap/src/cli/command.rs
bonomat: I'll fix it quickly.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -161,8 +162,26 @@ where
                                     continue;
                                 }
                             }
-
-                            let state0 = match State0::new(btc, xmr, self.env_config, self.bitcoin_wallet.as_ref(), &mut OsRng).await {
+                            // TODO: This should be cleaned up.
+                            let tx_redeem_fee = self.bitcoin_wallet
+                                .estimate_fee(bitcoin::TxRedeem::weight())
+                                .await.unwrap();
```
thomaseizinger: Why are you blowing up the whole event-loop here if this fails? As far as I can see, this can definitely be hit by IO errors happening in the underlying electrum-client!

swap/src/bitcoin.rs
```diff
@@ -304,4 +305,120 @@ mod tests {
...
+                || is_weight == expected_weight,
```
thomaseizinger, line 425: A more readable way would be to take the difference and assert that it is <= 2.

thomaseizinger (commented):
Sorry I didn't get around to a review earlier.

I think you addressed most of my comments in patches anyway so merging it is fine but in general, please work on your patch / commit ordering.

The overall design of this feature changed at least once throughout the whole PR from hard-coded constants to a dynamic estimation. Newly introduced functions and variables were moved around and renamed in later patches which could have easily been avoided.

It might be a bit more work for you but it saves a lot of time for reviewers if they don't have to go through the same thinking process that you did while developing but can just evaluate distinct features in their final form that build on top of each other.

swap/tests/harness/mod.rs
```diff
@@ -636,19 +639,13 @@ impl TestContext {
...
+                - bitcoin::Amount::from_sat(TX_REFUND_FEE);
 
-        let bob_submitted_cancel = btc_balance_after_swap
-            == self.bob_starting_balances.btc
-                - lock_tx_bitcoin_fee
-                - bitcoin::Amount::from_sat(2 * bitcoind::TX_FEE);
-
-        // The cancel tx can be submitted by both Alice and Bob.
-        // Since we cannot be sure who submitted it we have to assert accordingly
-        assert!(alice_submitted_cancel || bob_submitted_cancel);
```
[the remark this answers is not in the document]
bonomat: It shouldn't matter who published the cancel transaction and the assert was actually faulty: The fee is always subtracted from bob's `bob_starting_balances`.

swap/src/cli/command.rs
```diff
@@ -53,6 +56,9 @@ pub enum Command {
 
         #[structopt(long = "tor-socks5-port", help = "Your local Tor socks5 proxy port", default_value = DEFAULT_TOR_SOCKS5_PORT)]
         tor_socks5_port: u16,
+
+        #[structopt(long = "bitcoin-target-block", help = "Within how many blocks should the Bitcoin transactions be confirmed.", default_value = DEFAUL_BITCOIN_CONFIRMATION_TARGET)]
+        bitcoin_target_block: usize,
```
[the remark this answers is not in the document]
bonomat: It's a usize because down in the code we pass it into a function which expect it to be a usize.

swap/src/bitcoin/wallet.rs
```diff
@@ -298,19 +304,57 @@ where
...
+        // Doing some heavy math here :)
+        // `usize` is 32 or 64 bits wide, but `f32`'s mantissa is only 23 bits wide.
+        // This fine because such a big transaction cannot exist and there are also no
+        // negative fees.
+        #[allow(
+            clippy::cast_precision_loss,
+            clippy::cast_possible_truncation,
+            clippy::cast_sign_loss
+        )]
+        let sats_per_vbyte = ((weight as f32) / 4.0 * fee_rate.as_sat_vb()) as u64;
```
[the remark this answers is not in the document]
bonomat: Good point. I did not consider using `Decimal` for this.

bonomat:
@thomaseizinger 

Oh 🦀. I'm sorry, I didn't realize how the commit-history got out of hand :/. 
I fully agree with this feedback and will checkout gitabsob.

swap/tests/harness/mod.rs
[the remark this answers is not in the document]
bonomat: Thinking about this more, I might have misread the intention of this piece of code here. 

While we cannot detect who _really_ published it we can assume that if Bob's balance is not minus `tx_cancel_fee` and minus `tx_refund_fee` that Alice was the one who published it. 

The introduced change asserts this but might not highlight it, i.e. it only asserts that Bob received back his initial money minus `tx_cancel_fee` and `tx_refund_fee`. If this fails, it is not immediately obvious that Alice might have published `tx_cancel`. Either way, he failed publishing `tx_refund`.

PR #474, opened 2021-05-05
Add more log details

### LINKED ISSUE #448: ASB Metrics
As a maker I would like a cli command to see for all swaps: 

- when they were started
- amounts being swapped (incl rate at start time)
- date at each step (fund/redeem/refund) including rate
- taker's id

Hint: currently old states are overwritten in the DB. We could just add a new DB entry for each event including details.

bonomat:
Resolves #448 

1. The first commit adds an additional log statement of the exchange rate for each state-update. This is useful because it allows us to measure profitability easily, i.e. by knowing what was the exchange rate when the swap was started and what was it when it was finalized.
2. The second commit changes a bunch of log messages. 
3. The third commit is adds a new commandline flag to toggle json format.


swap/src/protocol/alice/swap.rs
```diff
@@ -39,15 +50,23 @@ pub async fn run_until(mut swap: Swap, exit_early: fn(&AliceState) -> bool) -> R
...
+    mut rate_service: LR,
+) -> Result<AliceState>
+where
+    LR: LatestRate,
+{
+    let rate = rate_service
+        .latest_rate()
+        .map_or("NaN".to_string(), |rate| format!("{}", rate));
+
+    info!(%state, %rate, "Update");
```
thomaseizinger, line 69: Instead of "Update", maybe "Advancing state" is more descriptive?

swap/src/bin/asb.rs
```diff
@@ -63,8 +63,9 @@ async fn main() -> Result<()> {
     };
 
     info!(
-        "Database and Seed will be stored in directory: {}",
-        config.data.dir.display()
+        db_folder = %config.data.dir.display(),
+        "Database and Seed will be stored in",
+
     );
```
thomaseizinger, line 68: I think for the CLI, we removed that line entirely because it is redundant with what is logged while reading the config.

swap/src/bin/asb.rs
```diff
@@ -83,20 +84,21 @@ async fn main() -> Result<()> {
             ask_spread,
         } => {
             let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
             let bitcoin_balance = bitcoin_wallet.balance().await?;
-            info!("Bitcoin balance: {}", bitcoin_balance);
+            info!(%bitcoin_balance, "Bitcoin balance");
```
thomaseizinger, line 90: The field name is also logged everytime. Would be nice to not be so repetitive here. Maybe the log message can say something more general like:

`Initialized Bitcoin wallet` and then a field `balance` so in the end we have:

```
Initialized Bitcoin wallet balance=0.5 BTC
```

swap/src/bitcoin/wallet.rs
```diff
@@ -143,15 +143,16 @@ impl Wallet {
 
                         let new_status = match client.lock().await.status_of_script(&tx) {
                             Ok(new_status) => new_status,
-                            Err(e) => {
-                                tracing::warn!(%txid, "Failed to get status of script: {:#}", e);
+                            Err(error) => {
+                                tracing::warn!(%txid, %error, "Failed to get status of script.");
```
thomaseizinger, line 147: We usually don't have a dot at the end of log messages.

swap/src/bin/asb.rs
thomaseizinger: This seems to be a recurring theme throughout the change to the log messages in this PR so I am not gonna comment on every occurrence.

swap/src/monero/wallet.rs
```diff
@@ -330,7 +341,12 @@ where
 
         if tx.confirmations > seen_confirmations {
             seen_confirmations = tx.confirmations;
-            tracing::info!(%txid, "Monero lock tx has {} out of {} confirmations", tx.confirmations, conf_target);
+            tracing::info!(
+                %txid,
+                %seen_confirmations,
+                needed_confirmations = %conf_target,
+                "Monero lock transaction update."
+            );
```
thomaseizinger, line 349: Every `info!` within the CLI is user-facing, so these messages should really be handled with care IMO. I am sure we can say something better here other than just "Lock transaction update" which seems to be a pure placeholder so we can log out the values.

swap/src/network/cbor_request_response.rs
```diff
@@ -45,9 +45,9 @@ where
             e => io::Error::new(io::ErrorKind::Other, e),
         })?;
         let mut de = serde_cbor::Deserializer::from_slice(&message);
-        let msg = Req::deserialize(&mut de).map_err(|e| {
-            tracing::debug!("serde read_request error: {:?}", e);
-            io::Error::new(io::ErrorKind::Other, e)
+        let msg = Req::deserialize(&mut de).map_err(|error| {
+            tracing::debug!(%error, "serde read_request error");
```
thomaseizinger, line 49: I think these can just go away, we are passing the error on so we are likely just gonna end up logging them twice.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -164,8 +169,8 @@ where
 
                             let state0 = match State0::new(btc, xmr, self.env_config, self.bitcoin_wallet.as_ref(), &mut OsRng).await {
                                 Ok(state) => state,
-                                Err(e) => {
-                                    tracing::warn!(%peer, "Failed to make State0 for execution setup: {:#}", e);
+                                Err(error) => {
+                                    tracing::warn!(%peer, %error, "Failed to make State0 for execution setup.");
```
thomaseizinger, line 173: This change is not quite functionally equivalent. `{:#}` will log the whole causality chain for an `anyhow::Error` whereas `%` will just invoke regular `Display`. See this open ticket here: https://github.com/tokio-rs/tracing/issues/1311

It is not yet resolved and I think we should really keep the causality chain otherwise debugging might be painful.

swap/src/protocol/alice/swap.rs
```diff
@@ -210,8 +210,10 @@ where
                                 bail!("Waiting for Bitcoin transaction finality failed with {}! The redeem transaction was published, but it is not ensured that the transaction was included! You're screwed.", e)
                             }
                         },
-                        Err(e) => {
-                            error!("Publishing the redeem transaction failed with {}, attempting to wait for cancellation now. If you restart the application before the timelock is expired publishing the redeem transaction will be retried.", e);
+                        Err(error) => {
+                            error!(
+                                %error,
+                                "Publishing the redeem transaction failed. Attempting to wait for cancellation now. If you restart the application before the timelock is expired publishing the redeem transaction will be retried.");
```
thomaseizinger, line 216: I like how this log message knows about entire program's behaviour :upside_down_face:

swap/src/asb/config.rs
```diff
@@ -226,6 +276,9 @@ mod tests {
                 wallet_rpc_url: Url::from_str(DEFAULT_MONERO_WALLET_RPC_TESTNET_URL).unwrap(),
             },
             tor: Default::default(),
+            logging: Logging {
+                format: LoggingFormat::Plain,
+            },
```
thomaseizinger: Does this have to go in the config? I think a command-line switch would be sufficient no?

thomaseizinger (commented):
Overall: Very big fan of utilizing tracing's fields more in the messages, thanks for doing this!

I think all the problems you mentioned in the description go away if you just use a command-line flag for the log output.

I don't think we should be too smart about when we enable / disable JSON. If the app is just running within systemd, JSON logs will be pretty annoying :)

swap/src/bitcoin/wallet.rs
thomaseizinger: Scrolling through my local journal, I also don't see any dots.

swap/src/bin/asb.rs
bonomat: > This seems to be a recurring theme throughout the change to the log messages in this PR so I am not gonna comment on every occurrence.

My reasoning for this was that it simplifies filtering of the logs. 
Assume the following log: 
```json
{
   "bitcoin_balance":"0.1",
   "msg":"Bitcoin balance."
}
```

 you can simply match on the balance directly 

```bash
| jq .bitcoin_balance
```

while if the field is not unique (e.g. balance field shows up for Bitcoin and Monero) you will need to match on the siblings which can get super cumbersome: 

```
select(.msg == "Bitcoin balance.").bitcoin_balance 
```

I find the former one simpler.

swap/src/bitcoin/wallet.rs
bonomat: I was born and raised to use grammatically correct sentences including punctuation in log messages. 😬

Eitherway is fine with me.

swap/src/network/cbor_request_response.rs
bonomat: Do you mean the whole log message should go away? e.g.
```suggestion
```

swap/src/protocol/alice/event_loop.rs
bonomat: > It is not yet resolved and I think we should really keep the causality chain otherwise debugging might be painful.

Good catch, I did not know that.

swap/src/asb/config.rs
bonomat: Happy to do so, that could solve my issue I think.

swap/src/asb/config.rs
```diff
@@ -88,8 +88,8 @@ pub struct ConfigNotInitialized {}
 pub fn read_config(config_path: PathBuf) -> Result<Result<Config, ConfigNotInitialized>> {
     if config_path.exists() {
         info!(
-            "Using config file at default path: {}",
-            config_path.display()
+            default_path = %config_path.display(),
+            "Using config file at default path. ",
```
da-kami, line 92: 🛠 I think this comment is actually incorrect (probably a legacy) because we are not in the context of a `default` path for the config here but this accounts for any config path passed to this function!

swap/src/bitcoin/wallet.rs
da-kami: I noticed that there are some occurrences of `. ` (space after the dot) as well. Would be nice if we can clean this up. I am in favor of no dots, but that can be hard if we have multiple sentences in on message. Might be better to avoid that anyway though.

swap/src/monero/wallet.rs
da-kami: I think it is good to keep in mind that info (and debug if switched on) are currently user-facing on the CLI but I would not waste too much time into making the log messages more user friendly. IMO the above is fine.

I see the current solution as temporary. Eventually we will add a (simple) UI that processes the JSON and the user won't see the logs anymore.

swap/src/network/cbor_request_response.rs
da-kami: You can check if we properly log it upstream, but I assume we are. So yes, the complete log statement can be deleted here.

swap/src/protocol/alice/swap.rs
da-kami: I think it can be stripped down to `Publishing the redeem transaction failed.` - this is a rather delicate scenario where interaction from an ASB provider might be required that's why that much context was added.
Can we assume that an ASB provider will have to have this knowledge anyway...?

da-kami (approved):
Looking good, I added a few minor things that can be adapted.
I think it will be important to run this in production asap to see if it fulfills our needs.

swap/src/bin/asb.rs
da-kami: I think what @bonomat proposes is fine for now.

thomaseizinger: I see the point but I also think there is a better solution here. One where the human-readable logs don't suffer as much.

1. JSON logs are usually not consumed with JQ, but fed into a tool like vector etc. I'd be very surprised if you can't also easily find that specific log line using a simply search for "bitcoin balance" in whatever frontend sits on top of the log database.
2. You can utilize the _target_ of events to give them more context. (Targets can be specified explicitly instead of relying on the implicit target given by the current module name.)
3. In this particular instance, you could make the ledger an field as well if you want to easily match on it.

All I am asking is that we don't make one situation worse in trying to improve another. JSON logs are a good addition and definitely useful. I also very much like the use of structured logs in favor of free-form messages. However, the logs should be of good quality, regardless in how they are formatted.

The other thing that about this particular log event is that the message by itself doesn't make any sense but it should make sense without the attached fields. Hence my suggestion of `Initialized Bitcoin wallet`. The attached key-value pairs should be considered as **additional** information to the message but not as the primary thing we are outputting.

swap/src/monero/wallet.rs
thomaseizinger: Same as noted above, the issue I see here is that we use a tracing event to dump key-value pairs on the user and provide a message that describes what this information is rather than having the message by useful on its own.

A log message provides more value if it contains a clear _action_ of what happened when this particular line executed.

`Monero lock transaction update.` doesn't actually tell me what changed. Okay there is an update, but an update on what? Is the update good or bad? Should I be waiting for more updates?

Compared that to something like (exact wording subject to bike-shedding, also I don't want specify out complete solutions in my review here):

- `Confirmations for Monero lock tx increased`
- `Received new confirmation for Monero lock tx`

Even without the key-value pairs, this provides me with useful information.

swap/src/protocol/alice/swap.rs
thomaseizinger: @bonomat If you resolve conversations without changing the code, can you please add a comment why?

swap/src/bitcoin/wallet.rs
thomaseizinger: I don't think we have any other log messages with dots so I'd be in favor to not introducing them at this stage.

swap/src/protocol/alice/swap.rs
bonomat: I changed it to what you proposed: https://github.com/comit-network/xmr-btc-swap/blob/088ecc47923f04bd348764b7b341ae4261c49739/swap/src/protocol/alice/swap.rs#L69

thomaseizinger: Huh okay, sorry about that. Somehow that didn't show up in my diff.

swap/src/bin/asb.rs
thomaseizinger: To be clear, I am not hung up on the name of the field being `bitcoin_balance` vs `balance`. I just want the log message by itself to be useful, i.e. `info!(%bitcoin_balance, "Initialized Bitcoin wallet");` is totally fine with me. (Not repeating "bitcoin" would obviously be nice but it is much better than a pointless log of "Bitcoin balance").

bonomat: Ah, so the discussion is about the `message`. 
The changes in this PR aim at producing easily machine parsable logs (may it be JQ or other analyzers). I have no strong opinion on the log `message` itself as long as it does not change :D 

In general, if you prefer a different `message`, can you just propose a change please?

swap/src/protocol/alice/event_loop.rs
bonomat: I revert and print the error message inside the `message` and unified _somehow_ across all log messages.

swap/src/monero/wallet.rs
bonomat: Updated to `Received new confirmation for Monero lock tx`

swap/src/protocol/alice/swap.rs
bonomat: Done

swap/src/bin/asb.rs
bonomat: Changed to `Initialized Bitcoin wallet` and `Initialized Monero wallet`

da-kami (approved):
OK to merge - definitely an improvement!
IMO we can fiddle with the cleanup of some message / best practices for logs later so feel free to ignore my comments :)

swap/src/monero/wallet_rpc.rs
```diff
@@ -121,7 +121,10 @@ impl WalletRpc {
             .local_addr()?
             .port();
 
-        tracing::debug!("Starting monero-wallet-rpc on port {}", port);
+        tracing::debug!(
+            monero_wallet_rpc_port = %port,
```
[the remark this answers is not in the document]
bonomat: I thought it's more readable that way. I'll change it to: 
```suggestion
            %port,
```

swap/src/monero/wallet_rpc.rs
```diff
@@ -121,7 +121,10 @@ impl WalletRpc {
...
+            "Starting monero-wallet-rpc on port"
```
bonomat, line 126:
```suggestion
            "Starting monero-wallet-rpc on"
```

bonomat:
docker failed :/ 

bors retry

swap/src/bin/asb.rs
thomaseizinger: > In general, if you prefer a different message, can you just propose a change please?

In general I try not to because I don't see it as the reviewer's job to do the problem solving.

In this particular case, my first comment was literally a suggestion of what the line can be changed to so I am a bit confused of what else you are expecting?

thomaseizinger: Right. Will keep it in mind. It doesn't play very well with our aspiration of atomic commits so there isn't a whole lot of value in it which means I usually use to it only because it is nicely rendered by GitHub, not because of the utility of being able to apply the change directly.

PR #475, opened 2021-05-05
Replace custom "bytes" serde module with serde_with's `TryFromInto`

thomaseizinger:
`serde_with` is a really powerful crate that is basically `serde` on steroids!

The [latest release](https://github.com/jonasbb/serde_with/releases/tag/v1.9.0) adds a feature to (de)serialize data based and automatically convert via `TryFrom` and `TryInto`.

We can use this to get rid of the `bytes` serde module which essentially only called `TryFrom` on a type `T` based on a byte array.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,7 +611,7 @@ dependencies = [
  "ansi_term 0.11.0",
  "atty",
  "bitflags",
- "strsim",
+ "strsim 0.8.0",
  "textwrap",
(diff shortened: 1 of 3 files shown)


thomaseizinger:
I don't think this works because we are not actually testing this code anywhere :)

thomaseizinger:
See: https://github.com/comit-network/monero-epee-bin-serde/pull/4

PR #479, opened 2021-05-07
Use rust_decimal in estimate_fee function.

bonomat:
Follow-up PR from #466

### DIFF
--- a/swap/src/bitcoin/wallet.rs
+++ b/swap/src/bitcoin/wallet.rs
@@ -13,6 +13,9 @@ use bdk::wallet::AddressIndex;
 use bdk::{FeeRate, KeychainKind};
 use bitcoin::{Network, Script};
 use reqwest::Url;
+use rust_decimal::prelude::*;
+use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
 use std::collections::{BTreeMap, HashMap};
 use std::convert::TryFrom;
 use std::fmt;
@@ -25,8 +28,8 @@ const SLED_TREE_NAME: &str = "default_tree";
 
 /// Assuming we add a spread of 3% we don't want to pay more than 3% of the
 /// amount for tx fees.
-const MAX_RELATIVE_TX_FEE: f64 = 0.03;
-const MAX_ABSOLUTE_TX_FEE: u64 = 100_000;
+const MAX_RELATIVE_TX_FEE: Decimal = dec!(0.03);
+const MAX_ABSOLUTE_TX_FEE: Decimal = dec!(100_000);
 
 pub struct Wallet<B = ElectrumBlockchain, D = bdk::sled::Tree, C = Client> {
     client: Arc<Mutex<C>>,
@@ -339,12 +342,7 @@ where
         let min_relay_fee = client.min_relay_fee()?;
         tracing::debug!("Min relay fee: {}", min_relay_fee);
 
-        Ok(estimate_fee(
-            weight,
-            transfer_amount,
-            fee_rate,
-            min_relay_fee,
-        ))
+        estimate_fee(weight, transfer_amount, fee_rate, min_relay_fee)
     }
 }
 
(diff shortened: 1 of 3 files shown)


swap/src/bitcoin/wallet.rs
```diff
@@ -353,55 +350,69 @@ fn estimate_fee(
...
+
+    let min_relay_fee = if min_relay_fee.as_sat() == 0 {
+        // if min_relay_fee is 0 we don't fail, we just set it to 1 satoshi;
+        Amount::ONE_SAT
+    } else {
+        min_relay_fee
+    };
+
+    let weight = Decimal::from(weight);
+    let weight_factor = Decimal::from_f32(4.0).context("Could not parse 0.4 :D")?;
```
thomaseizinger, line 372: Rust-decimal has macros for things like this: https://docs.rs/rust_decimal/1.13.0/rust_decimal/#usage

swap/src/bitcoin/wallet.rs
```diff
@@ -353,55 +350,69 @@ fn estimate_fee(
...
+    let fee_rate = Decimal::from_f32(fee_rate.as_sat_vb()).context("Could not parse fee_rate.")?;
```
thomaseizinger, line 373: Might be worth changing `as_sat_vb` to return you a `Decimal`?

swap/src/bitcoin/wallet.rs
```diff
@@ -353,55 +350,69 @@ fn estimate_fee(
...
-    // couple of times.
-    #[allow(
-        clippy::cast_precision_loss,
-        clippy::cast_possible_truncation,
-        clippy::cast_sign_loss
-    )]
-    let max_allowed_fee = (transfer_amount.as_sat() as f64 * MAX_RELATIVE_TX_FEE).ceil() as u64;
+    let transfer_amount = Decimal::from(transfer_amount.as_sat());
+    let max_relative_tx_fee =
+        Decimal::from_f32(MAX_RELATIVE_TX_FEE).context("Could not parse relative tx fee.")?;
```
thomaseizinger, line 385: Using the `dec!` macro might allow you do declare the `const`s right away as `Decimal` perhaps?

swap/src/bitcoin/wallet.rs
```diff
@@ -896,10 +906,41 @@ mod tests {
...
+            let is_fee = estimate_fee(weight, amount, fee_rate, relay_fee).unwrap();
 
             // weight / 4 * 1_000  is always higher than MAX_ABSOLUTE_TX_FEE
             assert!(is_fee.as_sat() >= MAX_ABSOLUTE_TX_FEE);
         }
     }
+
+    proptest! {
+        #[test]
+        fn given_fee_above_max_should_always_panic(
```
thomaseizinger, line 918: I think you are confusing panic with returning an error?

swap/src/bitcoin/wallet.rs
```diff
@@ -353,55 +350,69 @@ fn estimate_fee(
...
+) -> Result<Amount> {
+    if transfer_amount.as_sat() <= 546 {
+        bail!("Amounts needs to be greater than Bitcoin dust amount.")
+    }
+    if fee_rate.as_sat_vb() <= 0.0 {
+        bail!("Fee rate needs to be > 0")
+    }
+    if fee_rate.as_sat_vb() >= 100_000_000.0 || min_relay_fee.as_sat() > 100_000_000 {
+        bail!("A fee_rate or min_relay_fee of > 1BTC does not make sense")
+    }
```
thomaseizinger, line 362: Did we just not have these checks before or am I missing something?

swap/src/bitcoin/wallet.rs
thomaseizinger: But actually, in this particular instance you could just avoid this by doing `weight / 4`? No need for `Decimal` here I believe.

thomaseizinger (approved):
Thanks for following up with that!

The actually calculation algorithm is quite cluttered with parsing functionality now. I've left some comments with ideas on how we could push this out of the meat of the function to make it easier to read.

There might even be a way of containing the core functionality in an infallible function but I haven't fully thought it through!

Definitely an improvement over what we currently have so ACK!

swap/src/bitcoin/wallet.rs
bonomat: You are not missing anything and there were no checks in place :) I added these when I added proptest-tests as I realized really high values simply do not make sense.

swap/src/bitcoin/wallet.rs
bonomat: It makes it more explicit though.

swap/src/bitcoin/wallet.rs
bonomat: Do you mean changing `fee_rate.as_sat_vb()` to return a Decimal? That's not in our control, it's coming from `bdk`.

swap/src/bitcoin/wallet.rs
```diff
@@ -896,10 +904,41 @@ mod tests {
...
             // weight / 4 * 1_000  is always higher than MAX_ABSOLUTE_TX_FEE
-            assert!(is_fee.as_sat() >= MAX_ABSOLUTE_TX_FEE);
+            assert!(is_fee.as_sat() >= MAX_ABSOLUTE_TX_FEE.to_u64().unwrap());
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn given_fee_above_max_should_always_errors(
+            sat_per_vb in 100_000_001.0f32..,
```
da-kami, line 917: nit: This bound is not 100% correct - because `sat_per_vb` is a float number and any position *after* the point `> 100_000_000.0` should error as well.

I wonder if it would make more sense to convert `sat_per_vb` to `u64` prior to `if sats_per_vbyte < min_relay_fee`. Then this comparison would be a clearer bound than the float one. Eventually you convert `sats_per_vbyte` to u64 anyway.

da-kami (approved):

swap/src/bitcoin/wallet.rs
bonomat: Good catch, I changed the checks inside the function to `>` and use the exact bound here. I prefer this to converting it to u64 rightaway would make the calculations more complex again.

PR #481, opened 2021-05-07
Min buy amount

da-kami:


swap/src/bin/swap.rs
```diff
@@ -94,23 +94,53 @@ async fn main() -> Result<()> {
...
+            debug!("Requesting quote");
+            let bid_quote = event_loop_handle.request_quote().await?;
+            info!("Received quote: 1 XMR ~ {}", bid_quote.price);
+
+            let initial_max_giveable = bitcoin_wallet.max_giveable(TxLock::script_size()).await?;
+            let (send_bitcoin, fees) = determine_btc_to_swap(
+                bid_quote.clone(),
                 bitcoin_wallet.balance(),
                 bitcoin_wallet.new_address(),
                 async {
```
thomaseizinger, line 106: >  If you can come up with a clever way to re-arrange code in/around determine_btc_to_swap let me know please :)

You just need to decompose this future into multiple so you can move the loop inside. I also think you can simplify this implementation quite a bit by going into the direction of what @bonomat was doing in https://github.com/comit-network/xmr-btc-swap/pull/474:

Include relevant information as fields:

```rust
tracing::info!(
    %balance,
    %max_giveable,
	%min_buy,
    "Received BTC",
);
```

Also, I believe you will need to `sync` in the loop, otherwise you never learn about your new balance.

swap/src/bin/swap.rs
```diff
@@ -394,74 +409,144 @@ mod tests {
...
+        let expected_fees = Amount::from_btc(0.0001).unwrap();
+
+        assert_eq!((amount, fees), (expected_amount, expected_fees))
+    }
+
+    #[tokio::test]
+    async fn given_no_balance_and_transfers_less_than_min_keep_waiting_till_enough() {
+        // TODO: We currently cannot depict this test because logic was moved
+        // into waiting for the deposit, but that logic cannot be move into
+        // determine_btc_to_swap because we run into lifetime issues with
```
thomaseizinger, line 532: A future can only be used once, if you need it multiple times, you need to add a layer of indirection that gives you a new future every time, aka a lambda.

swap/src/bin/swap.rs
```diff
@@ -336,51 +337,79 @@ async fn init_monero_wallet(
...
+        let minimum_amount = bid_quote.min_quantity;
+        let maximum_amount = bid_quote.max_quantity;
 
-    let balance = if initial_balance == Amount::ZERO {
         info!(
-            "Please deposit the BTC you want to swap to {} (max {})",
-            get_new_address.await?,
-            bid_quote.max_quantity
+            %deposit_address,
+            %current_maximum_giveable,
```
da-kami, line 365: Not sure we want to expose this to the user, but I see the CLI as a temporary solution so I think it is fine. 
We could also opt for naming it different as in `maximum_swap_amount` and potentially always print `amount` and `fees` together within here.

swap/src/bin/swap.rs
```diff
@@ -332,51 +333,79 @@ async fn init_monero_wallet(
     Ok((monero_wallet, monero_wallet_rpc_process))
 }
 
-async fn determine_btc_to_swap(
-    request_quote: impl Future<Output = Result<BidQuote>>,
-    initial_balance: impl Future<Output = Result<bitcoin::Amount>>,
+async fn determine_btc_to_swap<FB, TB, FMG, TMG, FS, TS>(
```
thomaseizinger, line 336: https://youtu.be/uUV3KvnvT-w?t=23 :joy:

swap/src/bin/swap.rs
```diff
@@ -94,23 +94,24 @@ async fn main() -> Result<()> {
...
 
-                    bitcoin_wallet.balance().await
-                },
-                bitcoin_wallet.max_giveable(TxLock::script_size()),
+            let (send_bitcoin, fees) = determine_btc_to_swap(
+                bid_quote.clone(),
+                max_givable().await?,
+                bitcoin_wallet.new_address(),
+                || bitcoin_wallet.balance(),
+                max_givable,
```
thomaseizinger, line 108: Just pass `max_givable` once.

swap/src/bin/swap.rs
```diff
@@ -94,23 +94,24 @@ async fn main() -> Result<()> {
...
-            let send_bitcoin = determine_btc_to_swap(
-                event_loop_handle.request_quote(),
-                bitcoin_wallet.balance(),
-                bitcoin_wallet.new_address(),
-                async {
-                    while bitcoin_wallet.balance().await? == Amount::ZERO {
-                        bitcoin_wallet.sync().await?;
+            debug!("Requesting quote");
+            let bid_quote = event_loop_handle.request_quote().await?;
+            info!("Received quote: 1 XMR ~ {}", bid_quote.price);
```
thomaseizinger, line 99: Why not leave that inside?

swap/src/bin/swap.rs
thomaseizinger: I guess you did that for the tests? You should be able to have state inbetween the calls to the Fn but might need to change the type to `FnMut`.

thomaseizinger (approved):
Concept ACK, nice tests, good patches!

swap/src/bin/swap.rs
da-kami: I tried to depict more clearly what actually has to be value and what a future / function in the API of `determin_btc_to_swap`. 
We only call `determine_btc_to_swap` once so I did not see any reason not to move this outside. 
Don't have strong feelings about it though, can move it inside again :)

swap/src/bin/swap.rs
da-kami: Yes, I did it for the tests. I think it's fine to leave it with the initial value passed in. I feel otherwise the tests get unnecessary complex. I was trying to put as little "logic" into the tests as possible.

bonomat (commented):
Nice work.

bonomat (approved):

swap/src/bin/swap.rs
thomaseizinger: Fine either way, your position also makes sense.

PR #485, opened 2021-05-11
Update assert_weight test to account for a range up to 8 bytes.

### LINKED ISSUE #482: Flaky test
#443 introduced static weight constant including tests which verify that the constant don't go out of sync. 
We assumed that the transaction weight can only vary by 2 bytes, but it turned out to be wrong. 

```
 ---- bitcoin::tests::calculate_transaction_weights stdout ----
May 10 03:52:20.187  INFO bdk::wallet: Derivation of 100 addresses from 0 took 29 ms    
thread 'bitcoin::tests::calculate_transaction_weights' panicked at 'TxRedeem to have weight 548, but was 545', swap/src/bitcoin.rs:422:9
```

We should investigate why this transaction's weight was so _small_.

bonomat:
Fixes #482 

Weights fluctuate because of the length of the signatures. Valid ecdsa signatures can have 68, 69, 70, 71, or 72 bytes. Since most of our transactions have 2 signatures the weight can be up to 8 bytes less than the static weight (4 bytes per signature). 
Since it is really hard to get these short signatures (<1 in 100), I also include the transaction in the assert message which will help for debugging purposes. 

Source: https://medium.com/coinmonks/on-bitcoin-transaction-sizes-97e31bc9d816

### DIFF
--- a/swap/src/bitcoin.rs
+++ b/swap/src/bitcoin.rs
@@ -393,40 +393,26 @@ mod tests {
             .unwrap();
         let refund_transaction = bob_state6.signed_refund_transaction().unwrap();
 
-        assert_weight(
-            redeem_transaction.get_weight(),
-            TxRedeem::weight(),
-            "TxRedeem",
-        );
-        assert_weight(
-            cancel_transaction.get_weight(),
-            TxCancel::weight(),
-            "TxCancel",
-        );
-        assert_weight(
-            punish_transaction.get_weight(),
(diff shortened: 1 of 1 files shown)


da-kami (approved):
Nice! That should hopefully be the end of that problem. Thanks for investing the time to get to the bottom of it

thomaseizinger (approved):

PR #490, opened 2021-05-11
Mainnet switch

### LINKED ISSUE #360: Test parsing of CLI arguments
Argument  us avoiding mistakes with parameter combinations and defaults.

The basic idea is to add another layer of argument structs in between `structopt` structs and `main` like done before (never merged):

https://github.com/comit-network/xmr-btc-swap/blob/948cc99e9d87dfd710fa3cae15bb777710669dbd/swap/src/cli/command.rs

### LINKED ISSUE #446: Mainnet switch
The ASB and Cli should be configurable to run on mainnet.

### LINKED ISSUE #478: Create separate struct for subcommands in ASB.
Since this is suplicated for multiple commands you might want to extract that into a separate struct and use [](https://docs.rs/structopt/0.3.21/structopt/#flattening).

_Originally posted by @da-kami in https://github.com/comit-network/xmr-btc-swap/pull/466#discussion_r627049333_

### LINKED ISSUE #506: Waiting for multiple confirmations does not work
Triggered `mainnet` swap it is stuck at waiting for additional confirmations after the transaction was included.
For testnet swaps we never ran into this problem, because it was always configured to 1 confirmation.

It appears the `wait_until_final` function for bitcoin does not behave as intended:

https://github.com/comit-network/xmr-btc-swap/blob/7fc5a47619a598846e4bbc14c075654233598ebe/swap/src/bitcoin/wallet.rs#L201-L226

The transaction already has [multiple confirmations](https://mempool.space/tx/fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d) but the application never picks them up.
As far as I can tell `wait_until_final` might only emit the inclusion itself (i.e. the first confirmation) and afterwards **no event** are emitted anymore, so both programs just hang indefinitely waiting for the target confirmations (3 for mainnet). 

This is a critical problem that we will have to address before merging mainnet.

relevant Alice logs:

```
2021-05-14 10:10:19  INFO swap{id=45051582-a5db-428c-882b-63ed2a6407d3}: Advancing state state=started rate=0.00808800 BTC
2021-05-14 10:10:19  INFO swap{id=45051582-a5db-428c-882b-63ed2a6407d3}: Waiting for Bitcoin transaction finality txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d required_confirmation=3
2021-05-14 10:10:25 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=unseen
2021-05-14 10:11:28 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=in mempool
2021-05-14 10:33:34 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=confirmed with 1 blocks
2021-05-14 10:33:34  INFO swap{id=45051582-a5db-428c-882b-63ed2a6407d3}: Waiting for Bitcoin transaction finality txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d seen_confirmations=1 needed_confirmations=3

```

relevant Bob logs:

```
2021-05-14 10:10:11 DEBUG Starting execution setup with 12D3KooWNxQpvdzJeHmt22bvTsLg5AthCum5TS9DTM9nuTreP9ca
2021-05-14 10:10:20  INFO Published Bitcoin transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d kind=lock
2021-05-14 10:10:20  INFO Waiting for Alice to lock Monero
2021-05-14 10:10:24 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=unseen
2021-05-14 10:11:28 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=in mempool
2021-05-14 10:33:33 DEBUG Transaction txid=fc22128ed90cf3b84aeb2befd493f01b1cd62a37737caed2c0f57c8b8ec8ae3d status=confirmed with 1 blocks

```

da-kami:
Fixes  #446
Fixes #360 
Fixes #506 
Fixes #478 

To be precise: It is actually a testnet switch, because I think mainnet should be default.

I took several assumptions on the way (e.g. network support, ...).

At this stage any feedback welcome :)

TODO:

- [ ] successful mainnet swap with this code base before merging :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   1. Balance of ASB too low
   2. Buy amount sent by CLI exceeds maximum buy amount accepted by ASB
   3. ASB is running in resume-only mode and does not accept incoming swap requests
+- An issue where the monero daemon port used by the `monero-wallet-rpc` could not be specified.
+  The CLI parameter `--monero-daemon-host` was changed to `--monero-daemon-address` where host and port have to be specified.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The ASB's `--max-buy` and `ask-spread` parameter were removed in favour of entries in the config file.
   The initial setup includes setting these two values now.
+- From this version on the CLI and ASB run on **mainnet** by default!
+  When running either application with `--testnet` Monero network defaults to `stagenet` and Bitcoin network to `testnet3`.
+  This is a breaking change.
+  It is recommended to run the applications with `--testnet` first and not just run the application on `mainnet` without experience.
 
 ## [0.5.0] - 2021-04-17
 
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -180,12 +180,8 @@ pub fn read_config(config_path: PathBuf) -> Result<Result<Config, ConfigNotIniti
     Ok(Ok(file))
 }
 
-pub fn initial_setup<F>(config_path: PathBuf, config_file: F, testnet: bool) -> Result<()>
-where
-    F: Fn(bool) -> Result<Config>,
-{
-    let initial_config = config_file(testnet)?;
-    let toml = toml::to_string(&initial_config)?;
+pub fn initial_setup(config_path: PathBuf, config: Config) -> Result<()> {
+    let toml = toml::to_string(&config)?;
 
     ensure_directory_exists(config_path.as_path())?;
     fs::write(&config_path, toml)?;
@@ -352,7 +348,7 @@ mod tests {
             },
         };
 
-        initial_setup(config_path.clone(), |_| Ok(expected.clone()), false).unwrap();
+        initial_setup(config_path.clone(), expected.clone()).unwrap();
         let actual = read_config(config_path).unwrap().unwrap();
 
         assert_eq!(expected, actual);
@@ -392,7 +388,7 @@ mod tests {
             },
         };
 
-        initial_setup(config_path.clone(), |_| Ok(expected.clone()), true).unwrap();
+        initial_setup(config_path.clone(), expected.clone()).unwrap();
         let actual = read_config(config_path).unwrap().unwrap();
 
         assert_eq!(expected, actual);
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -1,6 +1,5 @@
 use crate::env::{Mainnet, Testnet};
 use crate::fs::{ensure_directory_exists, system_config_dir, system_data_dir};
-use crate::monero;
 use crate::tor::{DEFAULT_CONTROL_PORT, DEFAULT_SOCKS5_PORT};
 use anyhow::{bail, Context, Result};
 use config::ConfigError;
@@ -121,6 +120,8 @@ pub struct Network {
 pub struct Bitcoin {
     pub electrum_rpc_url: Url,
     pub target_block: usize,
+    pub finality_confirmations: Option<u32>,
+    #[serde(with = "crate::bitcoin::network")]
     pub network: bitcoin::Network,
 }
 
@@ -128,6 +129,8 @@ pub struct Bitcoin {
 #[serde(deny_unknown_fields)]
 pub struct Monero {
     pub wallet_rpc_url: Url,
+    pub finality_confirmations: Option<u64>,
+    #[serde(with = "crate::monero::network")]
     pub network: monero::Network,
 }
 
@@ -335,7 +338,7 @@ mod tests {
 
             monero: Monero {
                 wallet_rpc_url: defaults.monero_wallet_rpc_url,
-                network: monero::Network::Testnet,
+                network: monero::Network::Stagenet,
             },
             tor: Default::default(),
             maker: Maker {
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -33,15 +33,10 @@ pub struct Defaults {
 impl GetDefaults for Testnet {
     fn getConfigFileDefaults() -> Result<Defaults> {
         let defaults = Defaults {
-            config_path: system_config_dir()
-                .map(|dir| Path::join(&dir, "asb"))
-                .map(|dir| Path::join(&dir, "testnet"))
-                .map(|dir| Path::join(&dir, "config.toml"))
-                .context("Could not generate default config file path")?,
-            data_dir: system_data_dir()
-                .map(|proj_dir| Path::join(&proj_dir, "asb"))
-                .map(|proj_dir| Path::join(&proj_dir, "testnet"))
-                .context("Could not generate default data dir")?,
+            config_path: default_asb_config_dir()?
+                .join("testnet")
+                .join("config.toml"),
+            data_dir: default_asb_data_dir()?.join("testnet"),
             listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
             electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:60002")?,
@@ -56,15 +51,10 @@ impl GetDefaults for Testnet {
 impl GetDefaults for Mainnet {
     fn getConfigFileDefaults() -> Result<Defaults> {
         let defaults = Defaults {
-            config_path: system_config_dir()
-                .map(|dir| Path::join(&dir, "asb"))
-                .map(|dir| Path::join(&dir, "mainnet"))
-                .map(|dir| Path::join(&dir, "config.toml"))
-                .context("Could not generate default config file path")?,
-            data_dir: system_data_dir()
-                .map(|proj_dir| Path::join(&proj_dir, "asb"))
-                .map(|proj_dir| Path::join(&proj_dir, "mainnet"))
-                .context("Could not generate default data dir")?,
+            config_path: default_asb_config_dir()?
+                .join("mainnet")
+                .join("config.toml"),
+            data_dir: default_asb_data_dir()?.join("mainnet"),
             listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
             electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:50002")?,
@@ -74,6 +64,18 @@ impl GetDefaults for Mainnet {
 
         Ok(defaults)
     }
+}
+
+fn default_asb_config_dir() -> Result<PathBuf> {
+    system_config_dir()
+        .map(|dir| Path::join(&dir, "asb"))
+        .context("Could not generate default config file path")
+}
+
+fn default_asb_data_dir() -> Result<PathBuf> {
+    system_data_dir()
+        .map(|dir| Path::join(&dir, "asb"))
+        .context("Could not generate default config file path")
 }
 
 const DEFAULT_MIN_BUY_AMOUNT: f64 = 0.002f64;
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -293,10 +293,12 @@ pub fn query_user_for_initial_config(testnet: bool) -> Result<Config> {
         bitcoin: Bitcoin {
             electrum_rpc_url,
             target_block,
+            finality_confirmations: None,
             network: bitcoin_network,
         },
         monero: Monero {
             wallet_rpc_url: monero_wallet_rpc_url,
+            finality_confirmations: None,
             network: monero_network,
         },
         tor: TorConf {
@@ -330,6 +332,7 @@ mod tests {
             bitcoin: Bitcoin {
                 electrum_rpc_url: defaults.electrum_rpc_url,
                 target_block: defaults.bitcoin_confirmation_target,
+                finality_confirmations: None,
                 network: bitcoin::Network::Testnet,
             },
             network: Network {
@@ -338,6 +341,7 @@ mod tests {
 
             monero: Monero {
                 wallet_rpc_url: defaults.monero_wallet_rpc_url,
+                finality_confirmations: None,
                 network: monero::Network::Stagenet,
             },
             tor: Default::default(),
@@ -368,6 +372,7 @@ mod tests {
             bitcoin: Bitcoin {
                 electrum_rpc_url: defaults.electrum_rpc_url,
                 target_block: defaults.bitcoin_confirmation_target,
+                finality_confirmations: None,
                 network: bitcoin::Network::Bitcoin,
             },
             network: Network {
@@ -376,6 +381,7 @@ mod tests {
 
             monero: Monero {
                 wallet_rpc_url: defaults.monero_wallet_rpc_url,
+                finality_confirmations: None,
                 network: monero::Network::Mainnet,
             },
             tor: Default::default(),
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -1,4 +1,6 @@
+use crate::env::{Mainnet, Testnet};
 use crate::fs::{ensure_directory_exists, system_config_dir, system_data_dir};
+use crate::monero;
 use crate::tor::{DEFAULT_CONTROL_PORT, DEFAULT_SOCKS5_PORT};
 use anyhow::{bail, Context, Result};
 use config::ConfigError;
@@ -11,14 +13,69 @@ use serde::{Deserialize, Serialize};
 use std::ffi::OsStr;
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use tracing::info;
 use url::Url;
 
-const DEFAULT_LISTEN_ADDRESS_TCP: &str = "/ip4/0.0.0.0/tcp/9939";
-const DEFAULT_LISTEN_ADDRESS_WS: &str = "/ip4/0.0.0.0/tcp/9940/ws";
-const DEFAULT_ELECTRUM_RPC_URL: &str = "ssl://electrum.blockstream.info:60002";
-const DEFAULT_MONERO_WALLET_RPC_TESTNET_URL: &str = "http://127.0.0.1:38083/json_rpc";
-const DEFAULT_BITCOIN_CONFIRMATION_TARGET: usize = 3;
+pub trait GetDefaults {
+    fn getConfigFileDefaults() -> Result<Defaults>;
+}
+
+pub struct Defaults {
+    pub config_path: PathBuf,
+    data_dir: PathBuf,
+    listen_address_tcp: Multiaddr,
+    listen_address_ws: Multiaddr,
+    electrum_rpc_url: Url,
+    monero_wallet_rpc_url: Url,
+    bitcoin_confirmation_target: usize,
+}
+
+impl GetDefaults for Testnet {
+    fn getConfigFileDefaults() -> Result<Defaults> {
+        let defaults = Defaults {
+            config_path: system_config_dir()
+                .map(|dir| Path::join(&dir, "asb"))
+                .map(|dir| Path::join(&dir, "testnet"))
+                .map(|dir| Path::join(&dir, "config.toml"))
+                .context("Could not generate default config file path")?,
+            data_dir: system_data_dir()
+                .map(|proj_dir| Path::join(&proj_dir, "asb"))
+                .map(|proj_dir| Path::join(&proj_dir, "testnet"))
+                .context("Could not generate default data dir")?,
+            listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
+            listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
+            electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:60002")?,
+            monero_wallet_rpc_url: Url::parse("http://127.0.0.1:38083/json_rpc")?,
+            bitcoin_confirmation_target: 1,
+        };
+
+        Ok(defaults)
+    }
+}
+
+impl GetDefaults for Mainnet {
+    fn getConfigFileDefaults() -> Result<Defaults> {
+        let defaults = Defaults {
+            config_path: system_config_dir()
+                .map(|dir| Path::join(&dir, "asb"))
+                .map(|dir| Path::join(&dir, "mainnet"))
+                .map(|dir| Path::join(&dir, "config.toml"))
+                .context("Could not generate default config file path")?,
+            data_dir: system_data_dir()
+                .map(|proj_dir| Path::join(&proj_dir, "asb"))
+                .map(|proj_dir| Path::join(&proj_dir, "mainnet"))
+                .context("Could not generate default data dir")?,
+            listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
+            listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
+            electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:50002")?,
+            monero_wallet_rpc_url: Url::parse("http://127.0.0.1:18083/json_rpc")?,
+            bitcoin_confirmation_target: 3,
+        };
+
+        Ok(defaults)
+    }
+}
 
 const DEFAULT_MIN_BUY_AMOUNT: f64 = 0.002f64;
 const DEFAULT_MAX_BUY_AMOUNT: f64 = 0.02f64;
@@ -64,12 +121,14 @@ pub struct Network {
 pub struct Bitcoin {
     pub electrum_rpc_url: Url,
     pub target_block: usize,
+    pub network: bitcoin::Network,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct Monero {
     pub wallet_rpc_url: Url,
+    pub network: monero::Network,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
@@ -118,30 +177,11 @@ pub fn read_config(config_path: PathBuf) -> Result<Result<Config, ConfigNotIniti
     Ok(Ok(file))
 }
 
-/// Default location for storing the config file for the ASB
-// Takes the default system config-dir and adds a `/asb/config.toml`
-pub fn default_config_path() -> Result<PathBuf> {
-    system_config_dir()
-        .map(|dir| Path::join(&dir, "asb"))
-        .map(|dir| Path::join(&dir, "config.toml"))
-        .context("Could not generate default config file path")
-}
-
-/// Default location for storing data for the CLI
-// Takes the default system data-dir and adds a `/asb`
-fn default_data_dir() -> Result<PathBuf> {
-    system_data_dir()
-        .map(|proj_dir| Path::join(&proj_dir, "asb"))
-        .context("Could not generate default data dir")
-}
-
-pub fn initial_setup<F>(config_path: PathBuf, config_file: F) -> Result<()>
+pub fn initial_setup<F>(config_path: PathBuf, config_file: F, testnet: bool) -> Result<()>
 where
-    F: Fn() -> Result<Config>,
+    F: Fn(bool) -> Result<Config>,
 {
-    info!("Config file not found, running initial setup...");
-    let initial_config = config_file()?;
-
+    let initial_config = config_file(testnet)?;
     let toml = toml::to_string(&initial_config)?;
 
     ensure_directory_exists(config_path.as_path())?;
@@ -154,13 +194,30 @@ where
     Ok(())
 }
 
-pub fn query_user_for_initial_testnet_config() -> Result<Config> {
+pub fn query_user_for_initial_config(testnet: bool) -> Result<Config> {
+    let (bitcoin_network, monero_network, defaults) = if testnet {
+        tracing::info!("Running initial setup for testnet");
+
+        let bitcoin_network = bitcoin::Network::Testnet;
+        let monero_network = monero::Network::Stagenet;
+        let defaults = Testnet::getConfigFileDefaults()?;
+
+        (bitcoin_network, monero_network, defaults)
+    } else {
+        tracing::info!("Running initial setup for mainnet");
+        let bitcoin_network = bitcoin::Network::Bitcoin;
+        let monero_network = monero::Network::Mainnet;
+        let defaults = Mainnet::getConfigFileDefaults()?;
+
+        (bitcoin_network, monero_network, defaults)
+    };
+
     println!();
     let data_dir = Input::with_theme(&ColorfulTheme::default())
         .with_prompt("Enter data directory for asb or hit return to use default")
         .default(
-            default_data_dir()
-                .context("No default data dir value for this system")?
+            defaults
+                .data_dir
                 .to_str()
                 .context("Unsupported characters in default path")?
                 .to_string(),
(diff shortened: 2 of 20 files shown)


swap/src/protocol/alice/spot_price.rs
```diff
@@ -246,12 +281,15 @@ pub enum Error {
         balance: monero::Amount,
         buy: bitcoin::Amount,
     },
-
     #[error("Failed to fetch latest rate")]
     LatestRateFetchFailed(#[source] Box<dyn std::error::Error + Send + 'static>),
-
     #[error("Failed to calculate quote: {0}")]
     SellQuoteCalculationFailed(#[source] anyhow::Error),
+    #[error("Blockchain networks did not match, we are on {asb:?}, but request from {cli:?}")]
```
da-kami, line 288: OK to use Debug print for the error or will that break the log format?

da-kami:
I think the e2e tests are failing because I don't support `Regtest` over the wire... will have to find a solution for that tomorrow.

swap/src/protocol/alice/spot_price.rs
da-kami:
```
May 12 10:43:42.299 ERROR swap::protocol::alice::spot_price::tests: Blockchain networks did not match, we are on BlockchainNetwork { bitcoin: Testnet, monero: Stagenet }, but request from BlockchainNetwork { bitcoin: Mainnet, monero: Mainnet }

```

thomaseizinger:
I see you still pushing stuff, let me know when this is ready to be reviewed :)

da-kami:
> I see you still pushing stuff, let me know when this is ready to be reviewed :)

Yeah, sorry, my brain spotted some issues overnight :)
Ready now!

swap/src/asb/config.rs
```diff
@@ -11,14 +12,46 @@ use serde::{Deserialize, Serialize};
...
+                .context("Could not generate default config file path")?,
+            default_data_dir: system_data_dir()
+                .map(|proj_dir| Path::join(&proj_dir, "asb"))
+                .map(|proj_dir| Path::join(&proj_dir, "testnet"))
+                .context("Could not generate default data dir")?,
+            default_listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
+            default_listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
+            default_electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:60002")?,
+            default_monero_wallet_rpc_testnet_url: Url::parse("http://127.0.0.1:38083/json_rpc")?,
+            default_bitcoin_confirmation_target: 3,
```
bonomat, line 73: Since fees on testnet don't matter we can safely reduce this to 1.

```suggestion
            default_bitcoin_confirmation_target: 1,
```

swap/src/asb/config.rs
```diff
@@ -11,14 +12,46 @@ use serde::{Deserialize, Serialize};
...
+}
+
+pub struct Defaults {
+    pub default_config_path: PathBuf,
+    default_data_dir: PathBuf,
+    default_listen_address_tcp: Multiaddr,
+    default_listen_address_ws: Multiaddr,
+    default_electrum_rpc_url: Url,
+    default_monero_wallet_rpc_testnet_url: Url,
+    default_bitcoin_confirmation_target: usize,
```
bonomat: Nit: you can probably remove all the leading `default_` part of the variable names.

swap/src/bin/swap.rs
```diff
@@ -39,7 +39,24 @@ extern crate prettytable;
...
+        let data_dir = data.0.join("testnet");
+        let env_config = env::Testnet::get_config();
+
+        (data_dir, env_config)
+    } else {
+        let data_dir = data.0.join("mainnet");
+        let env_config = env::Mainnet::get_config();
+
+        (data_dir, env_config)
+    };
```
bonomat, line 59: Nice 👍

swap/src/bin/asb.rs
```diff
@@ -45,23 +45,49 @@ const DEFAULT_WALLET_NAME: &str = "asb-wallet";
...
 
+    let env_config = if testnet {
+        env::Testnet::get_config()
+    } else {
+        env::Mainnet::get_config()
+    };
+
+    if config.monero.network != env_config.monero_network {
+        bail!(format!(
+            "Expected monero network in config file to be {:?}  but was {:?}",
```
bonomat, line 76: 🙃
```suggestion
            "Expected monero network in config file to be {:?} but was {:?}",
```

swap/src/protocol/alice/spot_price.rs
bonomat: How does it look if you enable json logging? Imho it should still work but it might escape the brackets.

bonomat (approved):
🚀🌕

swap/src/asb/config.rs
da-kami: this is `Mainnet` 😬

For testnet it is set to `1`

da-kami:
Note: Latest `force` push does not change functionality but I squashed the initial ASB prep and the later ASB mainnet work together into one commit to make it easier to review and edit things.

swap/src/protocol/alice/spot_price.rs
da-kami:
```
{
  "timestamp": "May 12 17:32:46.156",
  "level": "ERROR",
  "fields": {
    "message": "Blockchain networks did not match, we are on BlockchainNetwork { bitcoin: Testnet, monero: Stagenet }, but request from BlockchainNetwork { bitcoin: Mainnet, monero: Mainnet }"
  },
  "target": "swap::protocol::alice::spot_price::tests"
}
```

That's fine I guess, the brackets are just within the string, I think that is valid.

da-kami:
This bug is to be fixed on the CLI side:

```
2021-05-12 18:01:46 DEBUG Starting monero-wallet-rpc on port=54873
Failed to parse arguments: unrecognised option '--mainnet'
```

Due to this the `monero-wallet-rpc` is not started correctly. For mainnet the switch just has to be removed (there is a `--stagenet` switch, but no mainnet, that is default...)

swap/src/bin/swap.rs
```diff
@@ -151,8 +151,8 @@ async fn main() -> Result<()> {
             table.printstd();
         }
         Command::Resume {
-            swap_id,
-            alice_multiaddr,
+            swap_id: SwapId { swap_id },
+            seller_addr: SellerAddr { seller_addr },
```
thomaseizinger, line 173: To make this less verbose, we could introduce another level of structs between the actual structopt structs and what is handled by `main.rs`.

This would also be inline with resolving https://github.com/comit-network/xmr-btc-swap/issues/360. We [had that at some point](https://github.com/comit-network/xmr-btc-swap/blob/948cc99e9d87dfd710fa3cae15bb777710669dbd/swap/src/cli/command.rs) but that PR never got merged.

swap/src/bin/swap.rs
thomaseizinger: Ideally this functionality doesn't live inside `main.rs` though because it is hard to test an noisy.

swap/src/monero.rs
```diff
@@ -19,6 +19,33 @@ use std::str::FromStr;
...
+
+impl From<Network> for monero::Network {
+    fn from(network: Network) -> Self {
+        match network {
+            Network::Mainnet => monero::Network::Mainnet,
+            Network::Stagenet => monero::Network::Stagenet,
+            Network::Testnet => monero::Network::Testnet,
+        }
+    }
+}
```
thomaseizinger, line 47: Instead of this indirection, would a serde module also help? If it is just for serde, using a module and `#[serde(with)]` will make things more ergonomic down the line because you don't need to convert back and forth.

Also, a module is easier to upstream. Alternatively, you can also upstream the `derive`. I am actually surprised it is not yet implemented.

swap/src/asb/config.rs
```diff
@@ -118,30 +177,11 @@ pub fn read_config(config_path: PathBuf) -> Result<Result<Config, ConfigNotIniti
...
-pub fn initial_setup<F>(config_path: PathBuf, config_file: F) -> Result<()>
+pub fn initial_setup<F>(config_path: PathBuf, config_file: F, testnet: bool) -> Result<()>
 where
-    F: Fn() -> Result<Config>,
+    F: Fn(bool) -> Result<Config>,
 {
-    info!("Config file not found, running initial setup...");
-    let initial_config = config_file()?;
-
+    let initial_config = config_file(testnet)?;
```
thomaseizinger, line 186: This seems overly complicated. You are passing two parameters in: `config_file` and `testnet` only to combine them as the very first thing to the `initial_config` variable. Why can't the caller just pass in `initial_config`?

swap/src/asb/config.rs
```diff
@@ -11,14 +13,69 @@ use serde::{Deserialize, Serialize};
...
+}
+
+impl GetDefaults for Mainnet {
+    fn getConfigFileDefaults() -> Result<Defaults> {
+        let defaults = Defaults {
+            config_path: system_config_dir()
+                .map(|dir| Path::join(&dir, "asb"))
+                .map(|dir| Path::join(&dir, "mainnet"))
+                .map(|dir| Path::join(&dir, "config.toml"))
+                .context("Could not generate default config file path")?,
```
thomaseizinger, line 64: If you move the `.context` up onto `system_config_dir`, you can just use `.join` for all the remaining calls and don't have to do this noisy `.map`.

swap/src/bin/asb.rs
```diff
@@ -45,23 +45,49 @@ const DEFAULT_WALLET_NAME: &str = "asb-wallet";
...
+            "Expected monero network in config file to be {:?} but was {:?}",
+            env_config.monero_network, config.monero.network
+        ));
+    }
+    if config.bitcoin.network != env_config.bitcoin_network {
+        bail!(format!(
+            "Expected bitcoin network in config file to be {:?} but was {:?}",
+            env_config.bitcoin_network, config.bitcoin.network
+        ));
+    }
```
thomaseizinger: We are overloading `main.rs` quite a bit here. All of this "derivation" of data based of the arguments could easily live in a different function that can be nicely tested even based on actual commandline parameters like we did here (this never merged): https://github.com/comit-network/xmr-btc-swap/blob/948cc99e9d87dfd710fa3cae15bb777710669dbd/swap/src/cli/command.rs#L213-L269

swap/src/network/spot_price.rs
```diff
@@ -59,11 +60,61 @@ pub enum Error {
...
     /// To be used for errors that cannot be explained on the CLI side (e.g.
     /// rate update problems on the seller side)
     Other,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
+pub struct BlockchainNetwork {
+    pub bitcoin: BitcoinNetwork,
+    pub monero: MoneroNetwork,
+}
```
thomaseizinger, line 76: Not quite sure if modelling these out completely isn't just unnecessary complexity? Would be good to give this some proper thought eventually.

swap/src/protocol/alice/spot_price.rs
thomaseizinger: @bonomat 

> Imho it should still work but it might escape the brackets.

Time to learn how [thiserror](https://docs.rs/thiserror/1.0.24/thiserror/) works :)

CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The ASB's `--max-buy` and `ask-spread` parameter were removed in favour of entries in the config file.
   The initial setup includes setting these two values now.
+- From this version on the CLI and ASB run on **mainnet** by default!
+  When running either application with `--testnet` Monero network defaults to `stagenet` and Bitcoin network to `testnet`.
```
thomaseizinger, line 54: Nit: Bitcoin's testnet is actually called `testnet3` :)

CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  This is a breaking change.
+  It is recommended to run the with `--testnet` first and not just run the application on `mainnet` without experience.
```
thomaseizinger, line 56: No shit sherlock :see_no_evil:

thomaseizinger (approved):
Nice patches, was very easy to review!

Left some notes on the way, but overall good effort!

swap/src/network/spot_price.rs
da-kami: Initially I had aggregated Bitcoin and Monero them into `Testnet` vs `Mainnet`, but I felt it somehwhat limits us in what networks are supported. I felt the network protocol should not make assumptions on how the application behaves -  on the application level we only support Bitcoin `testnet3` and Monero `stagenet` for the moment - but the network layer does not necessarily have to know that.
Happy to keep put more thought into this eventually...

swap/src/asb/config.rs
bonomat: upsi 😬

swap/src/bin/asb.rs
da-kami: Will tackle with in follow for both CLI and ASB.

swap/src/network/spot_price.rs
da-kami: The main motivation for doing it this way is, that I wanted to keep the network protocol unaware of application implementation specifics. The fact that the current implementation only supports Monero `stagenet` and Bitcoin `testnet3` should not be a concern of the protocol...?

Initially I had aggregated both Bitcoin and Network networks in either `Testnet` or `Mainnet`, but I found that somewhat incorrect, because Monero is not Testnet but Stagenet :] 

I think the current solution is the most correct, but probably not the simplest

da-kami:
bors ping

da-kami:
Merging blocked until fixed:

- [ ] https://github.com/comit-network/xmr-btc-swap/issues/506
- [ ] CLI testnet params are actually not picked up correctly

swap/src/bin/asb.rs
da-kami: After that bit me: https://github.com/comit-network/xmr-btc-swap/pull/490/commits/9f5a686d4897797f1fd496dd30499ea165334657

swap/src/bin/swap.rs
da-kami: https://github.com/comit-network/xmr-btc-swap/pull/490/commits/9f5a686d4897797f1fd496dd30499ea165334657

da-kami:
Note: 

Previously the CLI setup would *always* create a sub-folder `testnet` / `mainnet` in the provided data-dir. I found that a bit weird and with fixing and testing the CLI params I changed this to just use the `data-dir` provided directly (no creation of sub-folders if dir provided by user) - same behaviour as the ASB. 

(Mentioned because it is not that obvious from code...)

da-kami:
@thomaseizinger @bonomat ready to review the latest changes please!

swap/src/cli/command.rs
```diff
@@ -14,38 +18,219 @@ pub const DEFAULT_STAGENET_MONERO_DAEMON_ADDRESS_TESTNET: &str = "monero-stagene
 // See: https://1209k.com/bitcoin-eye/ele.php?chain=btc
 const DEFAULT_ELECTRUM_RPC_URL: &str = "ssl://electrum.blockstream.info:50002";
 // See: https://1209k.com/bitcoin-eye/ele.php?chain=tbtc
-const DEFAULT_ELECTRUM_RPC_URL_TESTNET: &str = "ssl://electrum.blockstream.info:60002";
+pub const DEFAULT_ELECTRUM_RPC_URL_TESTNET: &str = "ssl://electrum.blockstream.info:60002";
 
-const DEFAULT_BITCOIN_CONFIRMATION_TARGET: &str = "3";
-const DEFAULT_BITCOIN_CONFIRMATION_TARGET_TESTNET: &str = "1";
+const DEFAULT_BITCOIN_CONFIRMATION_TARGET: usize = 3;
+const DEFAULT_BITCOIN_CONFIRMATION_TARGET_TESTNET: usize = 3;
```
bonomat, line 24: I remember from our discussion from yesterday that you plan to align testnet and mainnet configs. 
I just want to point out that this is the target block, i.e. the fees will be estimated in a way so that your transaction will be included within 3 blocks. It does not say that we will be waiting for 3 confirmations.

swap/src/env.rs
```diff
@@ -75,13 +75,13 @@ impl GetConfig for Testnet {
     fn get_config() -> Config {
         Config {
             bitcoin_lock_confirmed_timeout: 12.hours(),
-            bitcoin_finality_confirmations: 1,
-            bitcoin_avg_block_time: 5.minutes(),
+            bitcoin_finality_confirmations: 3,
```
bonomat, line 78: Given we have a max swap limit of USD ~1k I'd be happy to reduce the number of min confirmations to 1 or 2. This would speed up the whole swap experience.

bonomat (approved):
Let's get the mainnet swaps rolling.

swap/src/cli/command.rs
```diff
@@ -7,16 +7,29 @@ use std::str::FromStr;
 use url::Url;
 use uuid::Uuid;
 
-pub const DEFAULT_STAGENET_MONERO_DAEMON_ADDRESS: &str = "monero-stagenet.exan.tech:38081";
+// See: https://moneroworld.com/
+pub const DEFAULT_STAGENET_MONERO_DAEMON_ADDRESS: &str = "node.moneroworld.com:18089";
+pub const DEFAULT_STAGENET_MONERO_DAEMON_ADDRESS_TESTNET: &str = "monero-stagenet.exan.tech:38081";
```
thomaseizinger, line 16: The naming seems to be slightly off here :sweat_smile: 

```suggestion
pub const DEFAULT_MAINNET_MONERO_DAEMON_ADDRESS: &str = "node.moneroworld.com:18089";
pub const DEFAULT_STAGENET_MONERO_DAEMON_ADDRESS: &str = "monero-stagenet.exan.tech:38081";
```

swap/src/bin/asb.rs
```diff
@@ -377,3 +373,24 @@ async fn register_tor_services(
...
+        } else {
+            env_config
+        };
+
+    if let Some(monero_finality_confirmations) = config.monero.finality_confirmations {
+        env_config.with_monero_finality_confirmations(monero_finality_confirmations)
+    } else {
+        env_config
+    }
+}
```
thomaseizinger, line 396: This looks like it would like to live within the `env` module so you can say:

```rust
let env_config = env::new(testnet, config);
```

swap/src/bin/asb.rs
```diff
@@ -69,11 +69,7 @@ async fn main() -> Result<()> {
         }
     };
 
-    let env_config = if testnet {
-        env::Testnet::get_config()
-    } else {
-        env::Mainnet::get_config()
-    };
+    let env_config = get_env_config_from_defaults_and_config(testnet, &config);
```
thomaseizinger, line 72: Nit: `is_testnet` would make the it clearer that this variable is a `bool`.

swap/src/env.rs
```diff
@@ -25,6 +25,22 @@ impl Config {
...
         sync_interval(self.monero_avg_block_time)
     }
+
+    pub fn with_bitcoin_finality_confirmations(
+        mut self,
+        bitcoin_finality_confirmations: u32,
+    ) -> Config {
+        self.bitcoin_finality_confirmations = bitcoin_finality_confirmations;
+        self
+    }
```
thomaseizinger, line 34: The fields are all public which means this function doesn't provide a whole lot of value given that there is no encapsulation or data hiding going on.

Consequently, the main decision on whether or not this is a good design comes down to a trade-off of ergonomics on the caller side vs the amount of code we need to add to make it happen.

Given all fields are public, we can also write (on the caller side):

```rust
let config = Config {
	bitcoin_finality_confirmations: 10,
	..config
};
```

without needing these functions.

Is this readable enough or do the "builder" functions add enough value for the amount code that is being added?

I am tempted to say we don't need the builder functions but I'll leave it up to you :)

swap/src/network/spot_price.rs
```diff
@@ -101,12 +101,12 @@ impl From<bitcoin::Network> for BitcoinNetwork {
...
         match network {
-            monero::Network::Mainnet => MoneroNetwork::Mainnet,
-            monero::Network::Stagenet => MoneroNetwork::Stagenet,
-            monero::Network::Testnet => MoneroNetwork::Testnet,
+            ::monero::Network::Mainnet => MoneroNetwork::Mainnet,
+            ::monero::Network::Stagenet => MoneroNetwork::Stagenet,
+            ::monero::Network::Testnet => MoneroNetwork::Testnet,
         }
     }
 }
```
thomaseizinger: Do we still need this conversion and the enum `MoneroNetwork` now that we have a `serde` module that we can use to serialize an instance of `monero::Network`?

swap/src/monero/wallet_rpc.rs
```diff
@@ -126,15 +126,23 @@ impl WalletRpc {
...
         let mut child = Command::new(self.exec_path())
             .env("LANG", "en_AU.UTF-8")
             .stdout(Stdio::piped())
             .kill_on_drop(true)
-            .arg(match network {
-                Network::Mainnet => "--mainnet",
-                Network::Stagenet => "--stagenet",
-                Network::Testnet => "--testnet",
-            })
+            .args(network_flag)
```
thomaseizinger, line 145: Nice explanation in the commit message for why we change from `arg` to `args` :ok_hand:

swap/src/cli/command.rs
```diff
@@ -208,3 +437,484 @@ fn parse_monero_address(s: &str) -> Result<monero::Address> {
...
+
+        pub fn with_debug(mut self) -> Self {
+            self.debug = true;
+            self
+        }
+    }
+
+    fn data_dir_path_cli() -> PathBuf {
+        system_data_dir().unwrap().join("cli")
+    }
```
thomaseizinger, line 612: It would be nice to have all of these helpers below the tests. When encountering the `tests` module, I am primarily interested in the tests, not the helpers that make the tests easier to read.

swap/src/cli/command.rs
```diff
@@ -208,3 +437,484 @@ fn parse_monero_address(s: &str) -> Result<monero::Address> {
...
+            MONERO_MAINNET_ADDRESS,
+            "--seller-addr",
+            MUTLI_ADDRESS,
+            "--seller-peer-id",
+            PEER_ID,
+        ];
+
+        let expected_args = Arguments::buy_xmr_mainnet_defaults();
+        let args = parse_args_and_apply_defaults(raw_ars).unwrap();
+        assert_eq!(expected_args, args);
```
thomaseizinger, line 629: Style nit: the following structure would follow "arrange, act, assert" more closely:

```suggestion
        let raw_ars = vec![
            BINARY_NAME,
            "buy-xmr",
            "--receive-address",
            MONERO_MAINNET_ADDRESS,
            "--seller-addr",
            MUTLI_ADDRESS,
            "--seller-peer-id",
            PEER_ID,
        ];

        let args = parse_args_and_apply_defaults(raw_ars).unwrap();

        assert_eq!(args, Arguments::buy_xmr_mainnet_defaults());
```

thomaseizinger: It might be even nicer if we don't `.unwrap()` but assert against `Ok()`:

```rust
let args = parse_args_and_apply_defaults(raw_ars);

assert_eq!(args, Ok(Arguments::buy_xmr_mainnet_defaults());
```

That way, the only thing that can fail in this test is the assertion :)

swap/src/cli/command.rs
```diff
@@ -208,3 +437,484 @@ fn parse_monero_address(s: &str) -> Result<monero::Address> {
...
+        let err = parse_args_and_apply_defaults(raw_ars).unwrap_err();
+
+        assert_eq!(
+            err.downcast_ref::<MoneroAddressNetworkMismatch>().unwrap(),
+            &MoneroAddressNetworkMismatch {
+                expected: monero::Network::Mainnet,
+                actual: monero::Network::Stagenet
+            }
+        );
+    }
```
thomaseizinger, line 673: Nice :ok_hand:

swap/src/bitcoin/wallet.rs
```diff
@@ -618,26 +620,32 @@ impl Client {
...
-        let latest_block = std::iter::from_fn(|| self.electrum.block_headers_pop().transpose())
-            .last()
-            .transpose()
-            .context("Failed to pop header notification")?;
+    fn update_latest_block(&mut self) -> Result<()> {
+        // Fetch the latest block for storing the height.
+        // We do not act on this subscription after this call, as we cannot rely on
+        // subscription push notifications because eventually the Electrum server will
+        // close the connection and subscriptions are not automatically renewed
+        // upon renewing the connection.
```
thomaseizinger, line 605: :+1:

swap/src/bitcoin/wallet.rs
```diff
@@ -146,12 +145,13 @@ impl Wallet {
             .or_insert_with(|| {
                 let (sender, receiver) = watch::channel(ScriptStatus::Unseen);
                 let client = self.client.clone();
+                let sync_interval = self.sync_interval;
 
                 tokio::spawn(async move {
                     let mut last_status = None;
 
                     loop {
-                        tokio::time::sleep(Duration::from_secs(5)).await;
```
thomaseizinger, line 154: NACK this "simplification".

This patch removes an essential aspect of the bandwidth optimization because now, every individual subscription will result in a network call whereas before, we only refreshed the local state based on the ping interval, regardless of how often the internal subscription queried the local state.

We need the "barrier" inside `update_state` to ensure we don't scale the number of network requests with the number of subscriptions.

thomaseizinger (changes requested):
Good work! Thank you for adding all the tests!

Given the evolution of this PR, some of the earlier patches are now stale and could be merged with some of the latter work.

There is a bug in the last patch that needs to be fixed, everything else should be good to merge!

swap/src/bin/asb.rs
thomaseizinger: The motivation should always be to have as little code inside `main.rs` as possible.

swap/src/bitcoin/wallet.rs
da-kami: Thanks! I knew that there was a reason that the ping was inside the client but when I looked at it yesterday I just did not see the benefit anymore. Did not think about multiple subscriptions ;)

I find it a bit hard to understand, because the subscription triggers the update (which is effectively only triggered if the current update interval expired). I think it is an efficient design, but it is not that easy to grasp why it is designed this way. I thought about handling the updates in a separate thread, but I am not sure that this complete decoupling would make it easier to understand.

swap/src/env.rs
da-kami: We could change both `testnet` and `mainnet` to `2` finality confirmations by default. That would effect both CLI and ASB. 
I am not sure how big of a security assumption this actually is. Have you done the math? 
Note that for the ASB we allow configuring the `finality_confirmations` in the config as well, but don't expose it in the wizzard, see https://github.com/comit-network/xmr-btc-swap/pull/490/commits/13bbdd5d6e59c7998b4907a5a2bb5467fb197051

swap/src/cli/command.rs
da-kami: I understand that this parameter is different from the `finality_confirmations` and  that for testnet changing this value should not have any effect.
I aligned this param together with the other parameters so one does not have to think why this is different for `mainnet` and `testnet`. Unfortunately I did not align it for the ASB :/

I'll set it back to `1` for testnet - seems it is more logical afterall and the testnet/mainnet alignment caused more confusion than it helped in this case :)

swap/src/bin/asb.rs
da-kami: I did not do that because `env` is share between CLI and ASB and the config is only relevant for the ASB. I can move it there, but I don't find that especially clean neither.

swap/src/network/spot_price.rs
da-kami: This was inspired by what you said in a previous PR about thinking about the wire types. I am treating the wire network types separately so we don't accidently break the wire protocol because an underlying library changes. 
In case the `Network` struct of the Bitcoin and Monero library changes the wire protocol would change. If we assume that these enums never change I can use them directly - shall I change it?

swap/src/cli/command.rs
da-kami: Could not easily make that happen:

```
binary operation `==` cannot be applied to type `std::result::Result<cli::command::Arguments, anyhow::Error>`
```

did not try much more though ;)

swap/src/cli/command.rs
da-kami: Yeah, totally aware that those are different values. After having a look at the commits: this is just a copy paster :D 
Changing it to `1` for testnet :)

swap/src/cli/command.rs
thomaseizinger: Ah yeah, good old anyhow not being PartialEq. Well, worth the attempt I guess :)

swap/src/bin/asb.rs
thomaseizinger: Fair point, I didn't think of that.

swap/src/network/spot_price.rs
thomaseizinger: Good thinking!

The nice thing about using the new serde module is that it is still our code and will not change with updates to the library. Hence any potential breaking change will result in a compile error because the mapping no longer works.

Happy to have dedicated types as well if you find them easier. We are getting the same amount of safety out of a serde-module though :)

swap/src/bitcoin/wallet.rs
thomaseizinger: One thing that could make things easier to understand is moving all the code that is spawned for each Subscription to a dedicated fn on subscription. That might make it more obvious that it happens per subscription whereas the client exists only once as can be seen by the Arc<Mutex>.
Not sure what you are getting at with threads, we have tasks available and already use them to run these things concurrently. `tokio::spawn` is effectively the `thread::spawn` you are thinking of I believe.

swap/src/env.rs
bonomat: >  I am not sure how big of a security assumption this actually is. Have you done the math?

Not really, I would fee safe with 2 confirmations up to a $10k limit. 

It is really hard to calculate but the hardware cost alone for this would be above $25billion. 
Meaning, I doubt we will have a 51% attack. Nevertheless, there is the risk of orphaned blocks which do happen every ~1000blocks. 2 orphaned blocks are extremely rare though. 

note: the calculation in https://www.crypto51.app/ are wrong

da-kami: If we go for Bitcoin  = 2 blocks by default then I would reduce Monero to `10` blocks as well (so we have an alignment on the work input for `20` mins for both). 
Monero enforces `10`by `10` unlock blocks being part of consensus. That's one reason why I would also not go lower with Bitcoin, otherwise it feels unbalanced.

I'll set Bitcoin to `2` and Monero to `10` finality confirmation blocks :)

swap/src/env.rs
```diff
@@ -107,6 +92,33 @@ fn sync_interval(avg_block_time: Duration) -> Duration {
...
+            }
+        } else {
+            env_config
+        };
+
+    if let Some(monero_finality_confirmations) = asb_config.monero.finality_confirmations {
+        Config {
+            monero_finality_confirmations,
+            ..env_config
+        }
```
bonomat, line 116: 📯 
Am I misreading this or is this part wrong? 

If `asb_config.bitcoin.finality_confirmations` is set then you only set the `bitcoin_finality_confirmations` and default for `monero_finality_confirmations`.
 and if it `asb_config.bitcoin.finality_confirmations` none you will set `monero_finality_confirmations`... 

I guess you want to have both set if they are `Some` right?

swap/src/bitcoin/wallet.rs
```diff
@@ -301,11 +305,30 @@ where
...
+            } else {
+                Ordering::Greater
+            }
+        });
+        psbt.global.unsigned_tx.output.sort_by(|a, _| {
+            if a.script_pubkey == script {
+                Ordering::Less
+            } else {
+                Ordering::Greater
+            }
```
bonomat, line 330: I thought I just commented but it seem to have disappeared? 

If I understand you correctly than we won't find the lock transaction later on if the locking script is not on vout 0. 

If this is true: what do you think of asserting here after the sort and fail the function if the script is not on vout 0?

bonomat (commented):
On minor idea. otherwise good to roll from my side.

swap/src/env.rs
bonomat: All good, I was confused about the indentations.

bonomat (approved):

swap/src/bitcoin/wallet.rs
thomaseizinger: We will always find the transaction, just subscribing to a different script but for checking confirmations, that doesn't matter.

This sorting only fixes that we can reuse a previous (program internal) subscription and don't check twice for the same TX.

swap/src/bitcoin/wallet.rs
```diff
@@ -293,6 +294,9 @@ where
         Ok(Amount::from_sat(fees))
     }
 
+    /// Builds a partially signed transaction from the given recipient address
+    /// and amount. Ensures that the address script is at output index `0`
+    /// for the created transaction.
```
thomaseizinger, line 299: Nit: The usual convention for rustdoc is to limit the basic explanation to one sentence and follow up with a more elaborate paragraph further down (basically the same as in a commit message).

Source: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

swap/src/bitcoin/wallet.rs
```diff
@@ -301,11 +305,30 @@ where
...
-        tx_builder.add_recipient(address.script_pubkey(), amount.as_sat());
+        tx_builder.add_recipient(script.clone(), amount.as_sat());
         tx_builder.fee_rate(fee_rate);
         let (psbt, _details) = tx_builder.finish()?;
+        let mut psbt: PartiallySignedTransaction = psbt;
+
+        // When subscribing to transactions we depend on the relevant script being at
+        // output index 0, thus we ensure the relevant output to be at index `0`.
+        psbt.outputs.sort_by(|a, _| {
+            if a.witness_script == Some(script.clone()) {
```
thomaseizinger, line 319: This should allow you to avoid the `.clone()`.

```suggestion
            if a.witness_script.as_ref() == Some(&script) {
```

thomaseizinger (approved):

da-kami:
The latest fore push is fixups and squashing a few commits together. Going for merge.

PR #493, opened 2021-05-12
Add checks to max_giveable amount.

### LINKED ISSUE #487: Swap Cli crashes if Bitcoin balance is 0.0
It looks like #481 revealed (or introduced) a bug which makes the CLI fail if Bitcoin balance is 0.0. 

The issue is that `bitcoin_wallet.max_givable(...)` returns an error if the balance is too low and hence we never get to the nicely designed logic which waits for enough bitcoin balance. 

This is a bit unfortunate because there is no way to _recover_ and deposit more BTC into the address  as the bitcoin address will only be printed later on. 

The line which will fail: 
https://github.com/comit-network/xmr-btc-swap/blob/c011e9506238fc7987310844f77d5ba72d2a6423/swap/src/bin/swap.rs#L100

And further down: 
https://github.com/comit-network/xmr-btc-swap/blob/26e0383e173f6c1d7fd364edf2c4cb9f1c82adc0/swap/src/bitcoin/wallet.rs#L328

Error message: 
```
Error: Failed to build transaction

Caused by:
    InsufficientFunds { needed: 53, available: 0 }
```

bonomat:
We need to check two things:
- balance to be higher than dust amount (546).
- balance to be higher than min-relay fee.

Additionally, the tx_builder might fail if not enough funds are in the wallet to pay for the overall transaction fees. In both cases, we return 0. 

fixes #487

### DIFF
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -343,7 +343,12 @@ where
 {
     debug!("Requesting quote");
     let bid_quote = bid_quote.await?;
-    info!("Received quote: 1 XMR ~ {}", bid_quote.price);
+    info!(
+        minimum_amount = %bid_quote.min_quantity,
+        maximum_amount = %bid_quote.max_quantity,
+        "Received quote: 1 XMR ~ {}",
+        bid_quote.price
+    );
 
     let mut current_maximum_giveable = max_giveable().await?;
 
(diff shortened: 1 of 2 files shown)


da-kami (approved):
👍

bonomat:
bors retry

da-kami:
bors retry

da-kami:
bors cancel

da-kami:
bors ping

da-kami:
bors ping

PR #507, opened 2021-05-15
Introduce read-only wallet for xpubs.

### LINKED ISSUE #489: Allow ASB to move money to a zpub instead of the internal wallet

bonomat:
At the end of a swaps (may it be redeemed, refunded, or punished), the Bitcoin funds should be transferred to an address controlled by the according party.
This wallet should be decoupled from the protocol wallet, i.e. the wallet which will be used to sign internal transactions.
In this PR, we introduce a `personal_wallet` which is meant for exactly this.

Resolves #489 

Some notes: 

1. `zpub-support`: a `zpub` is an `xpub` which implies native segwit support, i.e. an address derived from a `zpub` key should be prefixed with `bc1` (if mainnet). 
Unfortunately [bdk](https://github.com/bitcoindevkit/bdk) does not support `zpub` keys. However, we can achieve the same with `xpub` keys as well. By implying Bip84 descriptors, any key derived from this address will be a `bech32` (native segwit) address. 
3. optional `xpub`-entry for Asb config: A user running the ASB can optionally provide an `xpub` key. We will derive `redeem` and `punish` addresses from this. Since `xpub` keys are read-only, this has the implication that the `withdraw-all` command is useless. If, however, the user does not provide an `xpub` key, we use our internal seed to derive a wallet. In this case, the `withdraw-all` command is still useful :) 
4. `xpub`- support for Swap Cli it wasn't part of the ticket so I only pulled it through to the binary but do not expose it yet to the user. I guess this will need some more thinking, e.g. does an xpub make sense for the Swap Cli or would a single refund address be sufficient? A single address has the disadvantage of making our tests more complicated, i.e. if we have a full wallet, we can just _connect_ it to the blockchain and query for the balance while with an address we would need to build a bit more logic for this. I propose to create a follow-up ticket on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A minimum accepted Bitcoin amount for the ASB similar to the maximum amount already present.
   For the CLI the minimum amount is enforced by waiting until at least the minimum is available as max-giveable amount.
 - Added a new argument to ASB: `--json` or `-j`. If set, log messages will be printed in JSON format.
+- New read-only wallet for ASB which enables users to provide a xpub key to derive redeem/punish address from.
+  If not provided in the config, an internal wallet will be used. Users can still withdraw all their funds from the
+  internal wall.
 
 ### Fixed
 
@@ -48,6 +51,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The ASB's `--max-buy` and `ask-spread` parameter were removed in favour of entries in the config file.
   The initial setup includes setting these two values now.
+- Introduce a new read-only wallet for the ASB. This means, we now have two wallets within the ASB as well as
+  the Swap CLI. Because of this a new folder structure was introduced:
+  - `{data-dir}/wallet` can now be found in `{data-dir}/internal_wallet`. Note: an existing wallet will NOT.
+    automatically be moved to the new folder.
+  - `{data-dir}/personal_wallet` is the new read-only wallet.
+    This is a breaking change!
 
 ## [0.5.0] - 2021-04-17
 
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -1,6 +1,7 @@
 use crate::fs::{ensure_directory_exists, system_config_dir, system_data_dir};
 use crate::tor::{DEFAULT_CONTROL_PORT, DEFAULT_SOCKS5_PORT};
 use anyhow::{bail, Context, Result};
+use bdk::bitcoin::util::bip32::ExtendedPubKey;
 use config::ConfigError;
 use dialoguer::theme::ColorfulTheme;
 use dialoguer::Input;
@@ -11,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::ffi::OsStr;
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use tracing::info;
 use url::Url;
 
@@ -64,6 +66,7 @@ pub struct Network {
 pub struct Bitcoin {
     pub electrum_rpc_url: Url,
     pub target_block: usize,
+    pub xpub: Option<ExtendedPubKey>,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
@@ -224,6 +227,17 @@ pub fn query_user_for_initial_testnet_config() -> Result<Config> {
     }
     let ask_spread = Decimal::from_f64(ask_spread).context("Unable to parse spread")?;
 
+    let key: String = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Provide an xpub key if you want to receive your Bitcoin to an address outside of this application. Hit enter if you want to use an internal wallet.")
+        .default("".to_string())
+        .interact_text()?;
+
+    let xpub = if key.is_empty() {
+        None
+    } else {
+        Some(ExtendedPubKey::from_str(key.as_str()).context("Invalid xpub provided")?)
+    };
+
     println!();
 
     Ok(Config {
@@ -234,6 +248,7 @@ pub fn query_user_for_initial_testnet_config() -> Result<Config> {
         bitcoin: Bitcoin {
             electrum_rpc_url,
             target_block,
+            xpub,
         },
         monero: Monero {
             wallet_rpc_url: monero_wallet_rpc_url,
@@ -268,6 +283,7 @@ mod tests {
             bitcoin: Bitcoin {
                 electrum_rpc_url: Url::from_str(DEFAULT_ELECTRUM_RPC_URL).unwrap(),
                 target_block: DEFAULT_BITCOIN_CONFIRMATION_TARGET,
+                xpub : Some(ExtendedPubKey::from_str("tpubDC2Qwo2TFsaNC4ju8nrUJ9mqVT3eSgdmy1yPqhgkjwmke3PRXutNGRYAUo6RCHTcVQaDR3ohNU9we59brGHuEKPvH1ags2nevW5opEE9Z5Q").unwrap())
             },
             network: Network {
                 listen: vec![
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -297,11 +297,15 @@ async fn init_bitcoin_wallets(
     )
     .context("Failed to initialize protocol Bitcoin wallet")?;
 
+    let xpub = match config.bitcoin.xpub {
+        None => seed.derive_extended_public_key(env_config.bitcoin_network)?,
+        Some(key) => key,
+    };
     let wallet_dir = config.data.dir.join("personal_wallet");
     let personal_wallet = bitcoin::wallet::Wallet::new_readonly(
         config.bitcoin.electrum_rpc_url.clone(),
         &wallet_dir,
-        seed.derive_extended_private_key(env_config.bitcoin_network)?,
+        xpub,
         env_config,
     )
     .context("Failed to initialize personal Bitcoin wallet")?;
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -79,7 +79,7 @@ async fn main() -> Result<()> {
 
     match opt.cmd {
         Command::Start { resume_only } => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
@@ -180,7 +180,7 @@ async fn main() -> Result<()> {
             table.printstd();
         }
         Command::WithdrawBtc { amount, address } => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let amount = match amount {
                 Some(amount) => amount,
@@ -197,7 +197,7 @@ async fn main() -> Result<()> {
             bitcoin_wallet.broadcast(signed_tx, "withdraw").await?;
         }
         Command::Balance => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
             let bitcoin_balance = bitcoin_wallet.balance().await?;
@@ -211,7 +211,7 @@ async fn main() -> Result<()> {
         Command::ManualRecovery(ManualRecovery::Cancel {
             cancel_params: RecoverCommandParams { swap_id, force },
         }) => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let (txid, _) =
                 alice::cancel(swap_id, Arc::new(bitcoin_wallet), Arc::new(db), force).await??;
@@ -221,7 +221,7 @@ async fn main() -> Result<()> {
         Command::ManualRecovery(ManualRecovery::Refund {
             refund_params: RecoverCommandParams { swap_id, force },
         }) => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
             alice::refund(
@@ -238,7 +238,7 @@ async fn main() -> Result<()> {
         Command::ManualRecovery(ManualRecovery::Punish {
             punish_params: RecoverCommandParams { swap_id, force },
         }) => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let (txid, _) =
                 alice::punish(swap_id, Arc::new(bitcoin_wallet), Arc::new(db), force).await??;
@@ -254,7 +254,7 @@ async fn main() -> Result<()> {
             redeem_params: RecoverCommandParams { swap_id, force },
             do_not_await_finality,
         }) => {
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let (txid, _) = alice::redeem(
                 swap_id,
@@ -272,27 +272,36 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
-async fn init_bitcoin_wallet(
+async fn init_bitcoin_wallets(
     config: &Config,
     seed: &Seed,
     env_config: swap::env::Config,
 ) -> Result<bitcoin::Wallet> {
-    debug!("Opening Bitcoin wallet");
-    let wallet_dir = config.data.dir.join("wallet");
-
-    let wallet = bitcoin::Wallet::new(
+    debug!("Opening Bitcoin wallets");
+    let wallet_dir = config.data.dir.join("protocol_wallet");
+    let protocol_wallet = bitcoin::Wallet::new(
         config.bitcoin.electrum_rpc_url.clone(),
         &wallet_dir,
         seed.derive_extended_private_key(env_config.bitcoin_network)?,
         env_config,
         config.bitcoin.target_block,
     )
-    .await
-    .context("Failed to initialize Bitcoin wallet")?;
+    .context("Failed to initialize protocol Bitcoin wallet")?;
 
-    wallet.sync().await?;
+    // todo (phh): clean this up and use it.
+    let wallet_dir = config.data.dir.join("personal_wallet");
+    let _personal_wallet = bitcoin::Wallet::new_readonly(
+        config.bitcoin.electrum_rpc_url.clone(),
+        &wallet_dir,
+        seed.derive_extended_private_key(env_config.bitcoin_network)?,
+        env_config,
+    )
+    .context("Failed to initialize personal Bitcoin wallet")?;
 
-    Ok(wallet)
+    protocol_wallet.sync().await?;
+    _personal_wallet.sync().await?;
+
+    Ok(protocol_wallet)
 }
 
 async fn init_monero_wallet(
--- a/swap/src/bin/asb.rs
+++ b/swap/src/bin/asb.rs
@@ -79,11 +79,12 @@ async fn main() -> Result<()> {
 
     match opt.cmd {
         Command::Start { resume_only } => {
-            let bitcoin_wallet = init_bitcoin_wallets(&config, &seed, env_config).await?;
+            let (internal_wallet, personal_wallet) =
+                init_bitcoin_wallets(&config, &seed, env_config).await?;
 
             let monero_wallet = init_monero_wallet(&config, env_config).await?;
 
-            let bitcoin_balance = bitcoin_wallet.balance().await?;
+            let bitcoin_balance = internal_wallet.balance().await?;
             info!(%bitcoin_balance, "Initialized Bitcoin wallet");
 
             let monero_balance = monero_wallet.get_balance().await?;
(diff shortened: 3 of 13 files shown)


swap/src/protocol/alice/event_loop.rs
```diff
@@ -153,14 +156,14 @@ where
...
                                 .estimate_fee(bitcoin::TxRedeem::weight(), btc)
                                 .await;
-                            let tx_punish_fee = self.bitcoin_wallet
+                            let tx_punish_fee = self.internal_bitcoin_wallet
                                 .estimate_fee(bitcoin::TxPunish::weight(), btc)
                                 .await;
-                            let redeem_address = self.bitcoin_wallet.new_address().await;
-                            let punish_address = self.bitcoin_wallet.new_address().await;
+                            let redeem_address = self.personal_bitcoin_wallet.new_address().await;
+                            let punish_address = self.personal_bitcoin_wallet.new_address().await;
```
thomaseizinger, line 166: Passing around two instead of one wallet is quite the invasive change. What if we instead introduce functions such as `new_redeem_address` and `new_punish_address` and decide internally within `bitcoin::Wallet`, where we derive these addresses from? I think that would be less invasive because it allows us to handle things within `bitcoin::Wallet`.

swap/src/bin/swap.rs
```diff
@@ -296,18 +298,18 @@ async fn init_bitcoin_wallet(
     .context("Failed to initialize Bitcoin wallet")?;
 
     let wallet_dir = data_dir.join("personal_wallet");
-    let _personal_wallet = bitcoin::wallet::Wallet::new_readonly(
+    let personal_wallet = bitcoin::Wallet::new_readonly(
```
thomaseizinger, line 301: Having a "personal wallet" inside the CLI code that is not actually exposed to the user feels like unnecessary complexity to me. I don't want to need to read code that looks like it matter for the user but actually doesn't.

swap/src/seed.rs
```diff
@@ -41,6 +43,19 @@ impl Seed {
...
+    pub fn derive_extended_public_key(&self, network: bitcoin::Network) -> Result<ExtendedPubKey> {
+        let priv_key = self.derive_extended_private_key(network)?;
+        let secp = Secp256k1::new();
+
+        // we take the first child private key to derive a non_hardened xpub key
+        let non_hardened_xprv = priv_key.ckd_priv(&secp, ChildNumber::from_normal_idx(0)?)?;
+
+        let public_key = ExtendedPubKey::from_private(&secp, &non_hardened_xprv);
+        Ok(public_key)
+    }
```
thomaseizinger, line 57: So we are able to spend coins sent to this address because it is a child-key of the xprv that we use for the other wallet?

This is not obvious from my point of view. Can we just use the internal wallet for everything if the user didn't specify an xpub instead of deriving one that we happen to know the xprv for?

I am worried that the invariants that need to be maintained here for things to work are too complicated and we end up deriving addresses that we don't actually have the keys for.

swap/tests/harness/mod.rs
```diff
@@ -734,28 +743,15 @@ impl TestContext {
...
-        } else {
-            bail!("Bob in not in xmr redeemed state: {:?}", state);
-        };
-
-        let lock_tx_bitcoin_fee = self
-            .bob_internal_bitcoin_wallet
-            .transaction_fee(lock_tx_id)
-            .await?;
-
-        Ok(self.bob_starting_balances.btc - self.btc_amount - lock_tx_bitcoin_fee)
```
thomaseizinger: If you are deleting code like this, can we

a) have a dedicated commit for it
b) explain why we can do it without causing problems?

swap/tests/harness/mod.rs
```diff
@@ -675,36 +683,15 @@ impl TestContext {
...
-        let refund_fee = self
-            .alice_internal_bitcoin_wallet
-            .estimate_fee(TxRefund::weight(), self.btc_amount)
-            .await
-            .expect("To estimate fee correctly");
-
-        let bob_cancelled_and_refunded = btc_balance_after_swap
-            == self.bob_starting_balances.btc - lock_tx_bitcoin_fee - cancel_fee - refund_fee;
-
-        assert!(bob_cancelled_and_refunded);
```
thomaseizinger: Same thing here, please explain why it is ok to delete this. Ideally in a dedicated refactoring commit that doesn't involve other functionality.

swap/src/asb/config.rs
```diff
@@ -1,6 +1,7 @@
 use crate::fs::{ensure_directory_exists, system_config_dir, system_data_dir};
 use crate::tor::{DEFAULT_CONTROL_PORT, DEFAULT_SOCKS5_PORT};
 use anyhow::{bail, Context, Result};
+use bdk::bitcoin::util::bip32::ExtendedPubKey;
```
thomaseizinger, line 4: We also depend on `bitcoin`, no need to import things through the re-export. Please prefer a direct import from the `bitcoin` crate unless you want to be uniform across the codebase and import via bdk everywhere. Then we could remove the `bitcoin` dependency but I think we will need it anyway to activate certain features.

thomaseizinger (changes requested):
Thank you for taking this on so quickly!

I'll structure this message in two parts. First, some general feedback on the implementation and second, my personal opinion on the notes that you provided.

1. Your commits aren't particularly atomic.
- There are TODOs and hardcoded constants which are "resolved" in later commits.
- Refactorings and new features often happen in the same commit.
- Things get changed or renamed in later commits after they have been introduced earlier.
- The message often doesn't explain _why_ something happens. I've noted some of these instances inline.
- Personally, I think a PR is easiest to review if the initial patches are pure refactorings that are topped of with the feature we want. The reason being that I can adjust my mental model of the codebase without having to think about change in functionality. I also don't have to focus as much on correctness if I know it is a pure refactoring because the tests will (hopefully) catch a bug there. Finally, all that is "new" is introduced a single unit that just happens to fit in nicely because you've paved the way with your refactorings.

2. It seems to me that passing around a complete separate instance of `bitcoin::Wallet` is very invasive if all we want to do is derive two addresses from a different xpub. Not sure how it pans out but keeping the difference of where we derive the address from _internal_ to `bitcoin::Wallet` might make all of this a lot easier? It would imply making the `bitcoin::Wallet` aware of protocol terms like `refund` and `redeem` but I don't see why it couldn't do that. It is after all a wallet for a swap application. Anything more general should be upstreamed to bdk. Happy to roll with completely different designs as well but the current one is a fairly bad trade-off in terms of complexity vs. feature gain.

3. Like I commented inline, introducing the concept of a personal wallet to the CLI without exposing it to the user increases complexity for no particular gain. Would be nice if we can minimize the code changes to the CLI here unless we actually plan to expose this feature.

Regarding your notes:

> optional xpub-entry for Asb config: A user running the ASB can optionally provide an xpub key. We will derive redeem and punish addresses from this. Since xpub keys are read-only, this has the implication that the withdraw-all command is useless. If, however, the user does not provide an xpub key, we use our internal seed to derive a wallet. In this case, the withdraw-all command is still useful :)

If `withdraw-all` is pointless with an `xpub` set, I think we should tell the user explicitly when they invoke the command.

> xpub- support for Swap Cli it wasn't part of the ticket so I only pulled it through to the binary but do not expose it yet to the user. I guess this will need some more thinking, e.g. does an xpub make sense for the Swap Cli or would a single refund address be sufficient? A single address has the disadvantage of making our tests more complicated, i.e. if we have a full wallet, we can just connect it to the blockchain and query for the balance while with an address we would need to build a bit more logic for this. I propose to create a follow-up ticket on this.

Agreed that we should be dealing with this feature for the CLI in a separate ticket.

swap/src/protocol/alice/event_loop.rs
bonomat: I thought about something similar but decided against it. 
Here are my thoughts: 

1. I thought about adding a new field to `bitcoin::Wallet` i.e. `personal_wallet` so that we can achieve what you proposed. I decided against it because it would destroy the meaning of `bitcoin::Wallet`, because it's not a single `Wallet` anymore but something like `Wallets`... the abstraction did not add-up for me because how do you know from the outside what do you call `balance()`/`sign`/`new_address` on without knowing the internal structure in advance? 
2. I then wanted to have a completely separate wallet implementation for the read-only wallet, i.e. independent from `bitcoin::wallet`. For example, we could have `{crate}::bitcoin::wallets::internal` and `{crate}::bitcoin::wallets::personal`. `personal` could than be an offline-only wallet allowing easier testing key and address derivations. I then ran into problems because I had to assert the balances in the tests. It would also add some duplicated code.
3. I decided to re-use the same wallet which I think turned out quite nice. They are too similar to separate and it allowed me to not have to implement a new wallet for e2e tests which would allow me to assert on the balances. 

Imho this way is best trade-off between asking developers to go yet one level further down.

swap/src/seed.rs
bonomat: Yes, we could also just pass around the _same_ wallet twice, e.g. either clone or instantiate it again from `bitcoin::Wallet`. Imho this is harder to understand for a developer than having two keys, one for each wallet. 

> I am worried that the invariants that need to be maintained here for things to work are too complicated and we end up deriving addresses that we don't actually have the keys for.

I don't follow your concern: if we derive an xpub from the user's seed we should always be able to also derive the private key from it. 
If I misunderstood you, can you rephrase please?

swap/tests/harness/mod.rs
bonomat: > a) have a dedicated commit for it

Not really, it depends on the other change in this commit. 

> b) explain why we can do it without causing problems?

The reason is that the balance of Bob's `personal` bitcoin wallet needs to be 0 after the bitcoin was redeemed. Can add this reasoning into the commit message

swap/src/bitcoin/wallet.rs
```diff
@@ -268,6 +315,7 @@ where
         Ok(Amount::from_sat(balance))
     }
 
+    // TODO(philipp) remove me
```
bonomat, line 318: This todo is wrong 😅

bonomat:
Converted to draft and handed over to @da-kami :) .

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A minimum accepted Bitcoin amount for the ASB similar to the maximum amount already present.
   For the CLI the minimum amount is enforced by waiting until at least the minimum is available as max-giveable amount.
 - Added a new argument to ASB: `--json` or `-j`. If set, log messages will be printed in JSON format.
+- New read-only wallet for ASB which enables users to provide a xpub key to derive redeem/punish address from.
+  If not provided in the config, an internal wallet will be used. Users can still withdraw all their funds from the
+  internal wall.
```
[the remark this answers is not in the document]
thomaseizinger: I highly doubt that :)

a) we don't use `cargo tomlfmt` in this repo
b) it doesn't touch markdown files

swap/src/bitcoin/wallet.rs
```diff
@@ -41,21 +43,66 @@ pub struct Wallet<B = ElectrumBlockchain, D = bdk::sled::Tree, C = Client> {
 }
 
 impl Wallet {
-    pub async fn new(
+    pub fn new(
```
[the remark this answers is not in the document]
thomaseizinger: Seems to be in the works: https://github.com/rust-lang/rust-clippy/issues/7176

CHANGELOG.md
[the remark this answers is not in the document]
bonomat: Then I'll take everything back and claim the opposite. There is also something wrong with the sentence so I blame the internet monster.

swap/src/protocol/alice/event_loop.rs
thomaseizinger: > I thought about adding a new field to bitcoin::Wallet i.e. personal_wallet so that we can achieve what you proposed. I decided against it because it would destroy the meaning of bitcoin::Wallet, because it's not a single Wallet anymore but something like Wallets... the abstraction did not add-up for me because how do you know from the outside what do you call balance()/sign/new_address on without knowing the internal structure in advance?

Do you have to know that? I don't see why you would.

Unless I am missing something, we would only need to remove `new_address` and add new APIs that qualify, what the semantic of this address is:

- `new_redeem_address`
- `new_refund_address`

Like I pointed out in the overall feedback comment above, our `bitcoin::Wallet` is not a general-purpose Bitcoin wallet (unlike `bdk::Wallet`). It is more the wallet component for a swap application. Anything more generic can still live in there obviously but doesn't need to hold us back in creating an abstraction layer that works _better_ for our swap protocol with better being defined as: the protocol doesn't need to care about application semantics like the xpub feature.

We don't have to treat `bitcoin::Wallet` as a single wallet, that is what an instance of `bdk::Wallet` is for. Instead, imagine `bitcoin::Wallet` being a trait and serving as the interface for our protocol to access wallet functionality. What happens under the hood is irrelevant to the protocol as long as the functionality offered is internally consistent, f.e. `balance` doesn't return coins that you can't actually spend.

In the current design, we introduce a lot of complexity by "leaking" into the protocol that some clients (only ASB for now) might want a different wallet for redeem/refund/punish.

swap/src/seed.rs
thomaseizinger: > I don't follow your concern: if we derive an xpub from the user's seed we should always be able to also derive the private key from it.

Yes, as long as the rest of the application follows along here.
What I am saying is, for this to work, several parts of the code need to be kept semantically in sync and that part is not obvious!

To demonstrate what I mean: The "internal" wallet will find coins on the chain that it never handed out an address for. It just happened that a different piece of code, outside the wallet, derived an xpub that belongs to the same seed. Hence, the "ownership" of the xpriv within the wallet is broken.

Having to keep this in mind when modifying the code creates unnecessary complexity.

> Imho this is harder to understand for a developer than having two keys, one for each wallet.

I don't follow how this is harder to understand?
If you trace the code and you see the following:

```rust
let internal_wallet = // ...;

let personal_wallet = if xpub { new_personal_wallet(xpub) } else { personal_wallet.clone() };
```

All you need to keep in your head is:

"So if no xpub is set, I can assume the two wallets are the same."

No Bitcoin specific knowledge required about how derivation paths work etc.
You can simply discard the additional information about two wallets existing and conceptually "merge" them in your head.

I'll claim that it is easier to understand what an application does if you can follow the dataflow of its variables than if you need to have domain-specific knowledge to work out that addresses handed out be one wallet will end up in the balance of the other one. That is just counter-intuitive.

swap/tests/harness/mod.rs
thomaseizinger: > Not really, it depends on the other change in this commit.

It should be part of that commit then.

swap/src/bitcoin/wallet.rs
```diff
@@ -41,21 +43,66 @@ pub struct Wallet<B = ElectrumBlockchain, D = bdk::sled::Tree, C = Client> {
...
+
+    pub fn new_readonly(
+        electrum_rpc_url: Url,
+        wallet_dir: &Path,
+        key: impl DerivableKey<Segwitv0> + Clone,
+        env_config: env::Config,
+    ) -> Result<Self> {
+        // TODO (phh): figure out what exactly this should be. It will influence the
+        // addresses derived from this key
+        let fingerprint = bitcoin::util::bip32::Fingerprint::from_str("2b4497f1")?;
```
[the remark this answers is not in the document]
thomaseizinger: This definitely needs to be resolved before we can merge this.

swap/src/bin/asb.rs
```diff
@@ -272,27 +281,39 @@ async fn main() -> Result<()> {
...
+        config.bitcoin.electrum_rpc_url.clone(),
+        &wallet_dir,
+        xpub,
+        env_config,
+    )
+    .context("Failed to initialize personal Bitcoin wallet")?;
 
-    wallet.sync().await?;
+    internal_wallet.sync().await?;
+    personal_wallet.sync().await?;
```
thomaseizinger, line 314: The user will have to tell us the re-scan height of the key, otherwise we would assume the key hasn't been used before and we will re-use addresses.

swap/tests/harness/mod.rs
bonomat: >(Do the e2e tests even pass with the previous one but without this one?) 

Yes, because we were asserting something else: that the balance of the other wallet is `starting_balance - btc_amount - fee`.

swap/src/bin/asb.rs
bonomat: I disagree with your conclusion: 

If a user provides an xpub key he really should not use this xpub otherwise to get addresses. 
Otherwise we cannot guarantee that we won't re-use addresses. Even if we ask the user for a height. 

Also, this function does not care about the height. 
Does BDK support the feature of only scanning from a particular height?

swap/src/seed.rs
bonomat: > What I am saying is, for this to work, several parts of the code need to be kept semantically in sync and that part is not obvious!

There is the seed generation and the wallet instantiation. I don't see other moving parts which need to keep in sync. 

> "So if no xpub is set, I can assume the two wallets are the same."

I might have misunderstood your initial comment when I answered with : 
> Yes, we could also just pass around the same wallet twice, e.g. either clone or instantiate it again from bitcoin::Wallet. Imho this is harder to understand for a developer than having two keys, one for each wallet.

What I wanted to say with this that I find it harder to understand what you pointed out (We know we have different understandings about code complexity so I won't argue about this any further. :) 
):  

> "So if no xpub is set, I can assume the two wallets are the same."

I'm coming to a different conclusion: the code here is wrong. 
The two wallets should not be the same. 
We should not re-use the same xprv-key to derive the xpub for the personal wallet as there should not be a relation between them. We should just derive a different xprv-key from the seed for the personal wallet.

thomaseizinger: >  The two wallets should not be the same.

But at the moment (without this feature) they are, right? Is your claim that the current code and behaviour is wrong?

My goal would have been: don't "fix" something that is not broken. If we simply clone the original wallet, the xpub functionality is layered on top of what we currently have, without any changes in functionality if it is disabled. Given that we seem to be happy with how the code currently works, I don't see a reason to not do that.

> We should not re-use the same xprv-key to derive the xpub for the personal wallet as there should not be a relation between them. We should just derive a different xprv-key from the seed for the personal wallet.

That would address my concerns about "ownership" of the keys with each wallet. However, I think we care getting off-track here with the feature design in terms of what the original motivation is. I am not sure the concept of a personal vs internal wallet overall makes a whole lot of sense.

1. At the moment, the ASB doesn't need a Bitcoin wallet for anything other than receiving money (the other protocol is still in the works). Hence, even if you want to keep two wallets, there shouldn't be a need to pass a personal-wallet into any Alice / ASB related code.
2. Once we allow trading in both directions, Alice needs to be able to transfer BTC coins into a protocol as well. For that, it will need an xprv to sign transactions (i.e. an xpub is not enough).

If we have a personal and an internal wallet and the xpub feature is not used, which one should be used for (2)? If it is the personal wallet, what job does the internal wallet have? If it is the internal wallet, why can't we reuse the funds from the personal one for swaps in the other direction?

It appears to me that the ASB only has a single wallet and that is an internal one (internal because we control the keys). The only thing we want is to be able to directly send money out of the ASB without an intermediary transaction through the internal wallet. That is an _optimization_ of existing functionality, not a new, orthogonal feature. However, the current code doesn't make it very clear that we are optimization something. Rather, it suggests that there are reasons why we don't to manage to different sets of keys and I don't see why we should be doing that.

Keep in mind that we sort of have two wallets in the code: We have the ephemeral keys that we use for the protocol and an actual wallet that is used to hold all funds that we receive eventually. s Is that what you are trying to express with "internal" and "personal" wallet?

We don't use the wallet to derive these keys so as expressed above, we don't actually need a wallet instance for Alice at the moment for anything other than getting the two final addresses.

swap/src/bin/asb.rs
thomaseizinger: > If a user provides an xpub key he really should not use this xpub otherwise to get addresses.

Why not? It seems perfectly reasonable to wanting to pool together funds into a single wallet from different sources?

> Does BDK support the feature of only scanning from a particular height?

I thought it did but apparently it doesn't. https://docs.rs/bdk/0.7.0/bdk/wallet/struct.Wallet.html#method.sync

The feature of rescan height is actually coupled to the individual blockchain backend in BDK. Since we are using electrum, it doesn't matter.

bonomat: > Why not? It seems perfectly reasonable to wanting to pool together funds into a single wallet from different sources?

It's mostly a privacy concern: Anyone holding your xpub can derive addresses and hence can learn the balances. Meaning, the Asb or Cli can both learn about all your funds and transaction history unrelated to our applications. 

There is also the concern about security of address re-use which we cannot guarantee if addresses are being created outside of our control.

thomaseizinger: Fair. As we don't support any other backends than Electrum for now, it is not a performance concern so we don't need the rescan height actually.

swap/src/seed.rs
bonomat: > But at the moment (without this feature) they are, right? Is your claim that the current code and behaviour is wrong?
> [...]
> Keep in mind that we sort of have two wallets in the code: We have the ephemeral keys that we use for the protocol and an actual wallet that is used to hold all funds that we receive eventually. Is that what you are trying to express with "internal" and "personal" wallet?

Wrong might be a strong word but yes, I think the we should have two different wallets if the 'internal' wallet even qualifies to be called as a wallet as it is only used to sign protocol transactions. 

Anyway, changing this is out of scope of this ticket. 

What I wanted to say with: 

> I'm coming to a different conclusion: the code here is wrong.
The two wallets should not be the same.

Is that the from me newly introduced code is wrong. We should take the chance to decouple the two wallets so that there is no relation between them and not derive the xpub from the same xprv as used for the internal wallet. 

This is also useful when the ASB can trade in both directions, i.e. it has one wallet that holds the money and one _wallet_ which is used to sign protocol transactions.

thomaseizinger: > This is also useful when the ASB can trade in both directions, i.e. it has one wallet that holds the money and one wallet which is used to sign protocol transactions.

Okay, I think we finally came to the bottom of all this!

We don't use `bitcoin::Wallet` to sign protocol transactions. Never. All protocol transactions are signed with very specific private keys that are generated completely randomly during execution setup.

The only transactions signed by `bitcoin::Wallet` are:

- Funding `TxLock`, which by itself does not qualify as a protocol transaction as it spends arbitrary inputs
- Withdrawing to an address

All other transactions:

- spending TxLock as redeem
- spending TxLock as cancel
- spending TxCancel as punish
- spending TxCancel as refund

are not signed by `bitcoin::Wallet` but by randomly generated keys.

As such, I don't see why we need **two** `bitcoin::Wallet` instances.

bonomat: riiight... I don't know what I thought and followed blindly the recommendation in the issue...

Since this is the case, we definitely don't need two wallets. In fact, this ticket is way simpler than this PR proposes and can be reduced to very few changes. 

We let the user define a xpub key (or *pub as defined in [SLIP-132](https://github.com/satoshilabs/slips/blob/master/slip-0132.md)) and instantiate the wallet from this key. If it is an xpub key we have a read-only wallet but this is fine if we only swap in one one direction. 
If no key was provided, we derive one from the internal seed as we used to.

bonomat: cc @da-kami :⬆️ this conclusion is crucial if you take over this PR.

bonomat:
This PR was build under a wrong assumption as discovered in this comment: https://github.com/comit-network/xmr-btc-swap/pull/507#discussion_r633943295 

Closing this as it is easier to start from scratch.

PR #518, opened 2021-05-20
Fix bug that breaks swap ID for logging

da-kami:
Somehow I introduced this really stupid bug. Might have happened during a rebase. 

We create the swap id at the top of main.
Creating another id here breaks the name of the swap's logfile for the CLI, because the actual swap will have another id. 
Should definitely go in before releasing :)


bonomat (approved):
Ups 😬

PR #519, opened 2021-05-20
Avoid application error upon `--help`

da-kami:
Our `preview` release is currently broken because of this issue.

The way we use clap's `get_matches_from_safe` caused parsing errors upon `--help` and `--version` to be bubbled up to the application - which causes the application to exit with an error when running `--help` and `--version`.
This is solved by using `get_matches_from` instead of `get_matches_from_safe` which handles these known clap commands internally and exits early. 

Added smoke tests to CI so we catch such kind of problems in the future. Smoke testing by calling `--help` is cheap and should be OK in CI.

### DIFF
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -20,6 +20,8 @@ use std::future::Future;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
+use structopt::clap;
+use structopt::clap::ErrorKind;
 use swap::bitcoin::TxLock;
 use swap::cli::command::{parse_args_and_apply_defaults, Arguments, Command};
 use swap::database::Database;
@@ -44,7 +46,23 @@ async fn main() -> Result<()> {
         data_dir,
         debug,
         cmd,
-    } = parse_args_and_apply_defaults(env::args_os())?;
+    } = match parse_args_and_apply_defaults(env::args_os()) {
+        Ok(args) => args,
+        Err(e) => {
+            if let Some(clap_err) = e.downcast_ref::<clap::Error>() {
+                match clap_err.kind {
+                    ErrorKind::HelpDisplayed | ErrorKind::VersionDisplayed => {
+                        println!("{}", clap_err.message);
+                        std::process::exit(0);
+                    }
+                    _ => {
+                        bail!(e);
+                    }
+                }
+            }
+            bail!(e);
+        }
+    };
 
     match cmd {
         Command::BuyXmr {


da-kami:
Fun fact: the clap documentation documents this behavior, took some time to figure it out though 😅
https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from_safe

swap/src/cli/command.rs
```diff
@@ -765,6 +771,20 @@ mod tests {
         assert_eq!(args, Arguments::resume_testnet_defaults().with_debug());
     }
 
+    #[test]
+    fn given_help_does_not_panic() {
```
da-kami, line 775: I noticed that these tests may interrupt the test flow when run together with other tests. I guess this is because these new tests get interrupted before the function finishes.

Is there a better way to depict such kind of tests - i.e. check the exit code?

da-kami: Not that easy, maybe better to remove those tests. As far as I can see clap's [get_matches_from](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from) calls [process::exit](https://doc.rust-lang.org/std/process/fn.exit.html) and we cannot check anything against that. Interesting that the test running gets interrupted by that.

thomaseizinger: > Not that easy, maybe better to remove those tests. As far as I can see clap's get_matches_from calls process::exit and we cannot check anything against that. Interesting that the test running gets interrupted by that.

Well, calling `cargo test` builds "just another binary" that ends up being executed so if you call `process::exit` anywhere in there, the test suite just shuts down :)

thomaseizinger: You can actually find that binary in your `target` directory and run it manually as well!

da-kami: You mean from within the rust test? Like run the binary in the test?

thomaseizinger (changes requested):
This feels wrong. The main reason for extracting `parse_args_and_apply_defaults` was that it becomes a pure function and can be tested in isolation.

Calling a function that shuts down the current process is not really in the spirit of that :)

Can we go the other way and (as recommended by the docs) just check at the top of main if the error was `HelpDisplayed` or `VersionDisplayed` and exit ourselves?

swap/src/cli/command.rs
thomaseizinger: No, I was more pointing out that the entire test suite of a crate in Rust just gets compiled to a binary with `fn main` that ends up calling all your test functions in sequence and printing out error reporting in between.

Hence, calling `process::exit` behaves as expected: It shuts down this binary in the middle of the test run.

da-kami:
> This feels wrong. The main reason for extracting `parse_args_and_apply_defaults` was that it becomes a pure function and can be tested in isolation.
> 
> Calling a function that shuts down the current process is not really in the spirit of that :)
> 
> Can we go the other way and (as recommended by the docs) just check at the top of main if the error was `HelpDisplayed` or `VersionDisplayed` and exit ourselves?

If we do that we basically re-implement parts of clap - is that really what we want to do?

thomaseizinger:
> If we do that we basically re-implement parts of clap - is that really what we want to do?

I think that is the price you pay if you want more explicit control over the parsing. Literally killing the process is much more surprising behaviour that may be very hard to find / spot.

Also "re-implement" makes it sound like you will have to write a 100 lines of code. A downcast of the error and a pattern match should sort you.

rishflab:
Been a while since I touched this code so my questions might be a bit dumb. 

1. Don't we want the application to exit with an error when there is a parsing error with `--help` and `--version`?
2. Why is better to exit instead of propagating the error? Doesn't this effect the unit testability of the `parse_raw_args()`

swap/src/cli/command.rs
```diff
@@ -765,6 +771,20 @@ mod tests {
...
+        let raw_ars = vec![BINARY_NAME, "--help"];
```
rishflab, line 776: How does this work? there is no `help` field in `RawArguments`? Doesn't `--help` have to "applied" to another argument?

da-kami: > > If we do that we basically re-implement parts of clap - is that really what we want to do?
> 
> I think that is the price you pay if you want more explicit control over the parsing. Literally killing the process is much more surprising behaviour that may be very hard to find / spot.
> 
> Also "re-implement" makes it sound like you will have to write a 100 lines of code. A downcast of the error and a pattern match should sort you.

unfortunately it is not that simple - If I do this:

```
match parse_args_and_apply_defaults(env::args_os()) {
        Ok(args) => args,
        Err(e) => {
            if let Some(clap_err) = e.downcast_ref::<clap::Error>() {
                match clap_err.kind {
                    ErrorKind::HelpDisplayed | ErrorKind::VersionDisplayed => {
                        std::process::exit(0);
                    }
                    _ => {
                        bail!(e);
                    }
                }
            }
            bail!(e);
        }
    };
```

then `--help` does not print anything. So I would have to somehow manually trigger the help / version display as well.

da-kami: I reckon I can handle this inside the parsing function where I have access to clap - and could trigger `write_help`. But then I don't see much advantages over using `get_matches_from` because we would have to handle it in the function anyway.

da-kami:
> > If we do that we basically re-implement parts of clap - is that really what we want to do?
> 
> I think that is the price you pay if you want more explicit control over the parsing. Literally killing the process is much more surprising behaviour that may be very hard to find / spot.
> 
> Also "re-implement" makes it sound like you will have to write a 100 lines of code. A downcast of the error and a pattern match should sort you.

unfortunately it is not that simple - If I do this:

```
match parse_args_and_apply_defaults(env::args_os()) {
        Ok(args) => args,
        Err(e) => {
            if let Some(clap_err) = e.downcast_ref::<clap::Error>() {
                match clap_err.kind {
                    ErrorKind::HelpDisplayed | ErrorKind::VersionDisplayed => {
                        std::process::exit(0);
                    }
                    _ => {
                        bail!(e);
                    }
                }
            }
            bail!(e);
        }
    };
```

then `--help` does not print anything. So I would have to somehow manually trigger the help / version display as well.

da-kami:
> Been a while since I touched this code so my questions might be a bit dumb.
> 
>     1. Don't we want the application to exit with an error when there is a parsing error with `--help` and `--version`?
> 
>     2. Why is better to exit instead of propagating the error? Doesn't this effect the unit testability of the `parse_raw_args()`

I changed it to the way @thomaseizinger  proposed which makes these questions kind of outdated. The result of the parsing function is not returned to main and we exit if  the error was because help/version was triggered. We also print help/version from the message of the clap::Error returned by the parsing function. (The assumption was that clap would print it either way, but it does not)

rishflab (approved):

thomaseizinger (approved):

PR #520, opened 2021-05-21
Cli json logging

da-kami:
Combining `--json` with the debug file logger was a pita, so I stopped and went for a simpler approach:

If `--json` is given we just log to terminal - **no** logfiles will be created in `{data-dir}/logs`. 
The `--debug` flag applies to `--json` (i.e. if not given it will just print json on info level). We could change that to automatically fallback to debug - could add a `required_if` dependency via strucopt/clap but I did not want to invest more time into thinking about this.

Note on extending binary functionality:
As discussed with @thomaseizinger recently, we will have to think about multiple binaries soon, i.e. a binary that focuses to be used to building on top of it (that always logs json) and potientially keeping a simple CLI that is more user friendly. This also goes towards more clearly separating the application code from re-usable protocol / network code.

### DIFF
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -100,6 +100,8 @@ async fn main() -> Result<()> {
                 .behaviour_mut()
                 .add_address(seller_peer_id, seller_addr);
 
+            let our_peer_id = swarm.local_peer_id();
+            tracing::debug!(peer_id = %our_peer_id, "Initializing network module");
             let (event_loop, mut event_loop_handle) = EventLoop::new(
                 swap_id,
                 swarm,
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):

swap/src/bin/swap.rs
```diff
@@ -175,8 +177,8 @@ async fn main() -> Result<()> {
             let seller_peer_id = db.get_peer_id(swap_id)?;
 
             let mut swarm = swarm::bob(&seed, seller_peer_id, tor_socks5_port).await?;
-            let bob_peer_id = swarm.local_peer_id();
-            tracing::debug!("Our peer-id: {}", bob_peer_id);
+            let our_peer_id = swarm.local_peer_id();
+            tracing::debug!(peer_id = %our_peer_id, "Our");
```
thomaseizinger, line 181: TBH, I find these log messages really weird.

To me, the message should make sense even if the fields wouldn't be there. Can we not say something more general at what just happened in the program where it makes sense to attach the peer-ID to it? Like:

```
tracing::debug!(peer_id = %our_peer_id, "Initializing networking module");
```

thomaseizinger: To be fair, the old log message also wasn't great in that regard :)

PR #521, opened 2021-05-21
Introduce dedicated ParseResult to simplify error handling in main

thomaseizinger:


da-kami (approved):
Fun - I actually had a similar enum `ClapWrapper` as result type of the parsing function at some point yesterday, but I dropped all that in favor of letting clap handle it because I felt we are introducing unnecessary complexity :] 

Happy to handle it like this.

da-kami (approved):

Issue #514, opened 2021-05-19
The ASB advancing incorrectly when waiting too long for redeem tx


da-kami:
## Scenario and Analysis

Both CLI and ASB redeemed successfully, but the ASB's Bitcoin redeem transaction needs significant time to get included because the network overheats (in the current case transaction fees spiked because Bitcoin is crashing).

This eventually results in a timeout on the ASB side:

```
2021-05-19 17:46:30 DEBUG Transaction txid=0b8fdcff62408252762f60c498f68a4ab8ee19dcc0bb6ade3eb95e9bbd4a6c5a status=unseen
2021-05-19 17:46:37 DEBUG Transaction txid=0b8fdcff62408252762f60c498f68a4ab8ee19dcc0bb6ade3eb95e9bbd4a6c5a status=in mempool
2021-05-19 17:47:10  WARN Lost connection. Error Connection error: I/O error: connection is closed peer=12D3KooWL61zVenQs5bgrzXE9tFiqKzRxgdKybViagBk8jUcxj7Y address=/ip4/127.0.0.1/tcp/50617
2021-05-19 18:03:26 DEBUG Got notification for new block block_height=684169
2021-05-19 18:03:26 DEBUG Transaction txid=213da99a0d090ae0065146592ff32d886b5262927c8e754dd0549f3146fb7527 status=confirmed with 4 blocks
2021-05-19 18:08:40 DEBUG Got notification for new block block_height=684170
2021-05-19 18:08:40 DEBUG Transaction txid=213da99a0d090ae0065146592ff32d886b5262927c8e754dd0549f3146fb7527 status=confirmed with 5 blocks
2021-05-20 06:13:36  WARN Failed to get status of script. Error Failed to subscribe to header notifications: Made one or multiple attempts, all errored:
	- Network is down (os error 50)
 txid=0b8fdcff62408252762f60c498f68a4ab8ee19dcc0bb6ade3eb95e9bbd4a6c5a
2021-05-20 06:13:36 ERROR swap{id=dcb9784e-6204-4eb3-98f4-8997768b3889}: error=Waiting for Bitcoin transaction finality failed with Failed while waiting for next status update! The redeem transaction was published, but it is not ensured that the transaction was included! You're screwed.
[...]

PR #525, opened 2021-05-23
Bitcoin transaction published state

da-kami:
This improves the error handling on the ASB.
Once the Bitcoin redeem transaction is seen in mempool, the state machine cannot transition to a cancel scenario anymore because at that point the CLI will have redeemed the Monero.
The additional state then waits for transaction finality and prevents re-publishing the transaction.

### DIFF
--- a/swap/src/protocol/alice/recovery/cancel.rs
+++ b/swap/src/protocol/alice/recovery/cancel.rs
@@ -34,6 +34,9 @@ pub async fn cancel(
             (monero_wallet_restore_blockheight, transfer_proof, state3)
         }
 
+        // The redeem transaction was already published, it is not safe to cancel anymore
+        AliceState::BtcRedeemTransactionPublished { .. } => bail!(" The redeem transaction was already published, it is not safe to cancel anymore"),
+
         // The cancel tx was already published, but Alice not yet in final state
         AliceState::BtcCancelled { .. }
         | AliceState::BtcRefunded { .. }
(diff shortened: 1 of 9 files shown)


da-kami:
This is a quick-fix for https://github.com/comit-network/xmr-btc-swap/issues/514
I think tackling `broadcast` to only publish if the transaction was not included yet can be tackled at a later point.

swap/src/protocol/alice/recovery/cancel.rs
```diff
@@ -43,7 +46,7 @@ pub async fn cancel(
         | AliceState::BtcRedeemed
         | AliceState::XmrRefunded
         | AliceState::BtcPunished
-        | AliceState::SafelyAborted => bail!("Cannot cancel swap {} because it is in state {} which is not cancelable", swap_id, state),
+        | AliceState::SafelyAborted => bail!("Swap it is in state {} which is not cancelable", state),
```
thomaseizinger, line 49:
```suggestion
        | AliceState::SafelyAborted => bail!("Swap is in state {} which is not cancelable", state),
```

thomaseizinger (approved):

da-kami:
Ran into monero container startup problem: https://github.com/comit-network/xmr-btc-swap/runs/2652333552

I recorded a ticket for this now - it would be cool if we could report issues like this one on the ticket to see if it is actually just a problem of the `monero-wallet-rpc` startup:
 
https://github.com/comit-network/xmr-btc-swap/issues/new

Note: Not sure where to add "more logging" (that's what we usually do...) so I felt the issue is a good starting point to track problems like this and then see how we can tackle it.

da-kami:
bors retry

PR #535, opened 2021-05-25
Bitcoin network check when building PSBT

da-kami:
This ensures that funds are not sent to an address on the wrong network.

### DIFF
--- a/swap/src/asb/command.rs
+++ b/swap/src/asb/command.rs
@@ -1,7 +1,202 @@
+use crate::asb::config::GetDefaults;
 use crate::bitcoin::Amount;
+use crate::env;
+use crate::env::GetConfig;
+use anyhow::{bail, Result};
 use bitcoin::Address;
+use serde::Serialize;
+use std::ffi::OsString;
 use std::path::PathBuf;
+use structopt::StructOpt;
 use uuid::Uuid;
+
+pub fn parse_args<I, T>(raw_args: I) -> Result<Arguments>
+where
+    I: IntoIterator<Item = T>,
+    T: Into<OsString> + Clone,
+{
+    let matches = RawArguments::clap().get_matches_from_safe(raw_args)?;
+    let args = RawArguments::from_clap(&matches);
+
+    let is_json = args.json;
+    let is_testnet = args.testnet;
+    let config = args.config;
+    let command: RawCommand = args.cmd;
+
+    let arguments = match command {
+        RawCommand::Start { resume_only } => Arguments {
+            testnet: is_testnet,
+            json: is_json,
+            config_path: config_path(config, is_testnet)?,
+            env_config: env_config(is_testnet),
+            cmd: Command::Start { resume_only },
+        },
+        RawCommand::History => Arguments {
+            testnet: is_testnet,
+            json: is_json,
+            config_path: config_path(config, is_testnet)?,
+            env_config: env_config(is_testnet),
+            cmd: Command::History,
+        },
+        RawCommand::WithdrawBtc { amount, address } => Arguments {
+            testnet: is_testnet,
+            json: is_json,
+            config_path: config_path(config, is_testnet)?,
+            env_config: env_config(is_testnet),
+            cmd: Command::WithdrawBtc {
+                amount,
+                address: bitcoin_address(address, is_testnet)?,
+            },
+        },
+        RawCommand::Balance => Arguments {
+            testnet: is_testnet,
+            json: is_json,
+            config_path: config_path(config, is_testnet)?,
+            env_config: env_config(is_testnet),
+            cmd: Command::Balance,
+        },
+        RawCommand::ManualRecovery(manual_recovery) => match manual_recovery {
+            ManualRecovery::Redeem {
+                redeem_params: RecoverCommandParams { swap_id, force },
+                do_not_await_finality,
+            } => Arguments {
+                testnet: is_testnet,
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):

bonomat (commented):
I propose adding an additional check right after reading the config/input params to terminate early.

thomaseizinger:
> I propose adding an additional check right after reading the config/input params to terminate early.

How much earlier would that actually terminate? Is there anything that happens in-between config reading and the current check that should not happen? 
If so, then we should strive to organize things around in a different way. Duplicating the same check in multiple places is not sustainable.

A check where it matters - like it is introduced here - is going to be more beneficial in the long run (will automatically be re-used with other pieces of the code that call the same functionality).

bonomat:
> How much earlier would that actually terminate?

I'd do it the same way we do it in the CLI as a parameter check. That way it would terminate immediately. 

https://github.com/comit-network/xmr-btc-swap/blob/1de907b1765d99b92354819fd1e3b3ba5939af48/swap/src/cli/command.rs#L426-L444

>  Is there anything that happens in-between config reading and the current check that should not happen?

We initialize and sync the bitcoin wallet before the check which can take a significant amount of time. 

> A check where it matters - like it is introduced here - is going to be more beneficial in the long run (will automatically be re-used with other pieces of the code that call the same functionality).

Makes sense. Is there a reason why we would not want to have two checks? 
- 1 in the library as proposed here
- 1 in the application before anything starts up

thomaseizinger:
> Makes sense. Is there a reason why we would not want to have two checks?

The overall idea is to always minimize complexity as much as possible and duplicated code is complexity.

> We initialize and sync the bitcoin wallet before the check which can take a significant amount of time.

Right that is a good point.
We only use the Electrum backend at the moment though which is always pretty fast, even on the initial setup of the wallet.

> I'd do it the same way we do it in the CLI as a parameter check. That way it would terminate immediately if an invalid/wrong address was provided.

The unfortunate thing with the monero address is that is really only used at the very end of the program. _Ideally_ that check would also not exist in main and only be checked where we actually use it because that is where it belongs semantically. However, that would mean you could do a full swap without being able to take the money in the end. That is inacceptable functionality-wise so the duplication of the check is easier justified.

Unless it drastically improves things, I'd rather cut down on complexity as much as possible.

If wallet sync is the only thing we'd save, then I am not sure the additional complexity is justified because even the wallet sync doesn't hurt really with the Electrum backend.

---

The nice thing about following this mantra rigorously is that you can approach the codebase with a mindset of: "every line of code has an essential reason to exist" which allows you to be mindful about changing it (i.e. if I change this, I break a feature unless I replicate the functionality somewhere else).

Code that doesn't necessarily need to exist to make a feature happen makes it a lot harder to change the codebase because you have to essentially constantly question everything and hunt down the reason for why a line of code exists. That in turn is a very time consuming task and/or requires an amazing memory from when you worked on the codebase the last time.

da-kami:
Concerning this specific instance I actually agree with Philipp that we should add another check.
We should also have application argument tests similar to the ones on the CLI side. Thise should cover validating the address, because it is a better user experience to fail as early as possible. Loading the Bitcoin wallet or Monero wallet takes time. This is a user experience concern!

thomaseizinger:
> Concerning this specific instance I actually agree with Philipp that we should add another check.
> We should also have application argument tests similar to the ones on the CLI side. Thise should cover validating the address, because it is a better user experience to fail as early as possible. Loading the Bitcoin wallet or Monero wallet takes time. This is a user experience concern!

If you think the UX is improved significantly, go for it.

I was just want to stress that the added check is really only for UX and the wallet definitely needs one as well. Hence, if we can make the UX equally good with just the check in the wallet, that solution would be strictly better because it requires less code.

da-kami:
> > Concerning this specific instance I actually agree with Philipp that we should add another check.
> > We should also have application argument tests similar to the ones on the CLI side. Thise should cover validating the address, because it is a better user experience to fail as early as possible. Loading the Bitcoin wallet or Monero wallet takes time. This is a user experience concern!
> 
> If you think the UX is improved significantly, go for it.
> 
> I was just want to stress that the added check is really only for UX and the wallet definitely needs one as well. Hence, if we can make the UX equally good with just the check in the wallet, that solution would be strictly better because it requires less code.

I totally get your line of argument. I had a similar comment written down already but then thought about it again and came to the conclusion that in this case the application should check for sanity at this point. It is a completely different layer. Even though the user experience is not enhanced by much, it does help to see immediately that there is an issue and not just after opening the wallet. That way there can also not be side effects in the future.

swap/src/asb/command.rs
```diff
@@ -123,3 +318,333 @@ pub struct RecoverCommandParams {
...
+            cmd: Command::SafelyAbort {
+                swap_id: Uuid::parse_str(SWAP_ID).unwrap(),
+            },
+        };
+        let args = parse_args(raw_ars).unwrap();
+        assert_eq!(expected_args, args);
+    }
+
+    #[test]
+    fn given_config_path_then_config_path() {
```
thomaseizinger, line 612: A true statement.

thomaseizinger:
```suggestion
    fn given_user_provides_config_path_ignores_testnet_flag() {
```

thomaseizinger (approved):

da-kami:
bors retry

PR #547, opened 2021-05-30
Print the Bitcoin address to the terminal as a QR code

xscd:
It helps to fund the Bitcoin address with your mobile device without a need to copy-paste the receive address. 

Here is how it looks like under different terminals:
1) Gnome Terminal:

2) Alacritty:

3) Kitty:

4) Konsole:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Printing the deposit address to the terminal as a QR code.
+  To not break automated scripts or integrations with other software, this behaviour is disabled if `--json` is passed to the application.
+
 ## [0.7.0] - 2021-05-28
 
 ### Fixed
--- a/swap/Cargo.toml
+++ b/swap/Cargo.toml
@@ -37,6 +37,7 @@ monero-rpc = { path = "../monero-rpc" }
 pem = "0.8"
 prettytable-rs = "0.8"
 proptest = "1"
+qrcode = "0.12"
 rand = "0.8"
 rand_chacha = "0.3"
 reqwest = { version = "0.11", features = [ "rustls-tls", "stream", "socks" ], default-features = false }
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -14,6 +14,8 @@
 
 use anyhow::{bail, Context, Result};
 use prettytable::{row, Table};
+use qrcode::render::unicode;
+use qrcode::QrCode;
 use std::cmp::min;
 use std::env;
 use std::future::Future;
@@ -101,6 +103,7 @@ async fn main() -> Result<()> {
 
             let max_givable = || bitcoin_wallet.max_giveable(TxLock::script_size());
             let (amount, fees) = determine_btc_to_swap(
+                json,
                 event_loop_handle.request_quote(),
                 bitcoin_wallet.new_address(),
                 || bitcoin_wallet.balance(),
@@ -325,7 +328,18 @@ async fn init_monero_wallet(
     Ok((monero_wallet, monero_wallet_rpc_process))
 }
 
+fn qr_code(value: &impl ToString) -> Result<String> {
+    let code = QrCode::new(value.to_string())?;
+    let qr_code = code
+        .render::<unicode::Dense1x2>()
+        .dark_color(unicode::Dense1x2::Light)
+        .light_color(unicode::Dense1x2::Dark)
+        .build();
+    Ok(qr_code)
+}
+
 async fn determine_btc_to_swap<FB, TB, FMG, TMG, FS, TS>(
+    json: bool,
     bid_quote: impl Future<Output = Result<BidQuote>>,
     get_new_address: impl Future<Output = Result<bitcoin::Address>>,
     balance: FB,
@@ -357,6 +371,10 @@ where
         let deposit_address = get_new_address.await?;
         let minimum_amount = bid_quote.min_quantity;
         let maximum_amount = bid_quote.max_quantity;
+
+        if !json {
+            eprintln!("{}", qr_code(&deposit_address)?);
+        }
 
         info!(
             %deposit_address,
@@ -449,6 +467,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_max(0.01)) },
             get_dummy_address(),
             || async { Ok(Amount::from_btc(0.001)?) },
@@ -475,6 +494,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_max(0.01)) },
             get_dummy_address(),
             || async { Ok(Amount::from_btc(0.1001)?) },
@@ -502,6 +522,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_max(0.01)) },
             async { panic!("should not request new address when initial balance  is > 0") },
             || async { Ok(Amount::from_btc(0.005)?) },
@@ -529,6 +550,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_max(0.01)) },
             async { panic!("should not request new address when initial balance is > 0") },
             || async { Ok(Amount::from_btc(0.1001)?) },
@@ -556,6 +578,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_min(0.01)) },
             get_dummy_address(),
             || async { Ok(Amount::from_btc(0.0101)?) },
@@ -583,6 +606,7 @@ mod tests {
         ])));
 
         let (amount, fees) = determine_btc_to_swap(
+            true,
             async { Ok(quote_with_min(0.01)) },
             get_dummy_address(),
             || async { Ok(Amount::from_btc(0.0101)?) },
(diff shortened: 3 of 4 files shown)


rishflab:
Cool!

Would it be possible to make the QR code smaller and shift it to the right?

120103411-265f6180-c158-11eb-95e1-b37c2d16c2b6


We could also change the text formatting to suit such a layout by spreading it across more lines.

swap/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "swap"
 version = "0.7.0"
-authors = [ "The COMIT guys <hello@comit.network>" ]
+authors = ["The COMIT guys <hello@comit.network>"]
```
thomaseizinger, line 4: Please avoid reformatting like this.

swap/src/bin/swap.rs
```diff
@@ -358,11 +360,21 @@ where
...
+                .light_color(unicode::Dense1x2::Dark)
+                .build()
+                .as_str();
+
         info!(
             %deposit_address,
             %current_maximum_giveable,
             %minimum_amount,
             %maximum_amount,
+            %qr_code,
```
thomaseizinger, line 377: Having the QRCode part of the tracing statement is a bit weird as that will completely break the JSON layout of the logs I believe.

Can we just print this to stderr instead and omit it entirely if we are in json format (ie --json has been passed as an argument)?

That should also address @rishflab 's comment on the horizontal alignment.

thomaseizinger (changes requested):
Thanks for the contribution!

Made some comments inline. Concept ACK from my side. In the long run, I'd like to separate the different concerns of the CLI into distinct binaries but that is out of scope for here.

@da-kami what do you think?

xscd:
> Cool!
> 
> Would it be possible to make the QR code smaller and shift it to the right?
> 
> 120103411-265f6180-c158-11eb-95e1-b37c2d16c2b6
> 
> We could also change the text formatting to suit such a layout by spreading it across more lines.

Thanks for the feedback. Unfortunately this is the smallest as we can get (the qrcode crate already tries to print the smallest as possible on Unicode mode) except it may possible to remove the white border which would make scanning a pain especially devices with low quality cameras. I shifted it a little bit to the right so it looks better IMO. 

Screenshot from 2021-06-02 01-50-20

swap/Cargo.toml
xscd: Done. I reverted the formatting. Rustfmt caused this.

swap/src/bin/swap.rs
xscd: Yeah I never thought about the --json use case that would make sense. I moved the QR Code printing to stderr.

swap/Cargo.toml
thomaseizinger: Rustfmt is formatting `Cargo.toml` now? That is news to me!

xscd: Sorry it wasn't rustfmt. It was Even Better TOML extension (https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml#formatting) Interesting that rustfmt doesn't do Cargo.toml files tho. Also found you here :D https://github.com/rust-lang/rustfmt/issues/4091#issuecomment-730801431

swap/src/bin/swap.rs
```diff
@@ -358,6 +360,18 @@ where
...
+        let qr_code = code
+            .render::<unicode::Dense1x2>()
+            .dark_color(unicode::Dense1x2::Light)
+            .light_color(unicode::Dense1x2::Dark)
+            .build();
+
+        eprintln!("QR Code:{}", qr_code.lines().next().unwrap());
+        for line in qr_code.lines().skip(1) {
+            eprintln!("        {}", line)
+        }
```
thomaseizinger, line 373: I would like this to be skipped if the user passed `--json`, otherwise the QR code is going to break parsing the logs. The `--json` flag is meant for users who want to integrate the CLI into existing applications. For those, the expectation is that every line on stderr is valid JSON.

While you are at, can we extract this into a function such that we can write something like this:

```rust
if !json {
	eprintln!("{}", qr_code(deposit_address));
}
```

I would like `qr_code` to not have any side-effects and just return a String. That should be doable if you just join all the lines together with `\n`.

Thank you :)

thomaseizinger (changes requested):
Sorry for taking so long with following up on this!

Couple of things:

1. This needs a changelog entry.
2. CI is currently not passing.
3. I've made a comment inline that needs to be addressed.

@da-kami I assume you are ok with merging this given that you haven't said anything so far?

swap/src/bin/swap.rs
xscd: Done.

xscd:
> Sorry for taking so long with following up on this!

No issues :) I am happy to be able to contribute to the project even if its a minor thing. 

>     1. This needs a changelog entry.

I added a changelog entry but I am not a native speaker so if it sounds wrong or is badly written let me know.


>     2. CI is currently not passing.

I locally used dprint to format the Cargo.toml file I hope it will pass this time. 

>     3. I've made a comment inline that needs to be addressed.

I hope its better now. Let me know if it can be improved further.

swap/src/bin/swap.rs
```diff
@@ -327,6 +327,40 @@ async fn init_monero_wallet(
     Ok((monero_wallet, monero_wallet_rpc_process))
 }
 
+async fn qr_code(deposit_address: &bitcoin::Address) -> Result<String> {
```
thomaseizinger, line 330: This doesn't need to be `async` because there is no `.await` in this function.

swap/src/bin/swap.rs
```diff
@@ -327,6 +327,40 @@ async fn init_monero_wallet(
...
+        check_json.push(flag);
+    }
+
+    if check_json.contains(&"--json".to_string()) {
+        json = true;
+    } else {
+        json = false;
+    };
+    Ok(json)
+}
```
thomaseizinger, line 362: You don't need to check for the args manually, it is already parsed here: https://github.com/comit-network/xmr-btc-swap/blob/d5d0dda6e73c76911edcae513a0333ca11a35330/swap/src/cli/command.rs#L32

You just need to pass the flag through to where you need it.

CHANGELOG.md
```diff
@@ -129,7 +134,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue where Alice would not verify if Bob's Bitcoin lock transaction is semantically correct, i.e. pays the agreed upon amount to an output owned by both of them.
   Fixing this required a **breaking change** on the network layer and hence old versions are not compatible with this version.
 
-[Unreleased]: https://github.com/comit-network/xmr-btc-swap/compare/0.7.0...HEAD
+[unreleased]: https://github.com/comit-network/xmr-btc-swap/compare/0.7.0...HEAD
```
thomaseizinger, line 137: Why was this updated?

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Printing the deposit address to the terminal as a QR code.
+  Users who are doing automated scripts or integrating with other software can use the `--json` flag to avoid this behaviour. It won't print a QR code if the `--json` flag is being passed to the application.
```
thomaseizinger, line 13: From the way you worded this, I think there was a bit of a misunderstanding here. The `--json` flag doesn't exist to avoid the QRCode printing. The `--json` flag toggles whether or not the logs are printed in JSON format. It will make more sense once you try it!

Hence, what we should mention in the changelog is the fact that `--json` is conceptually incompatible with printing a QRCode and hence will be skipped.

Something like:

```suggestion
  To not break automated scripts or integrations with other software, this behaviour is disabled if `--json` is passed to the application.
```

swap/src/bin/swap.rs
```diff
@@ -327,6 +327,40 @@ async fn init_monero_wallet(
...
 }
 
+async fn qr_code(deposit_address: &bitcoin::Address) -> Result<String> {
+    let code = QrCode::new(deposit_address.to_string())?;
+    let qr_code = code
+        .render::<unicode::Dense1x2>()
+        .dark_color(unicode::Dense1x2::Light)
+        .light_color(unicode::Dense1x2::Dark)
+        .build();
+    let first_line = "QR Code:".to_string() + qr_code.lines().next().unwrap() + "\n";
```
thomaseizinger, line 337: Do we have to say "QR Code:"? It is kind of obvious IMO and it would make this implementation a lot simpler because we could just do: `qr_code.lines().join("\n")`

swap/src/bin/swap.rs
thomaseizinger: You could generalize the function-signature to this if you are feeling fancy:

```suggestion
async fn qr_code(value: &impl ToString) -> Result<String> {
```

thomaseizinger (changes requested):
Thanks, couple of more comments on the newly added code.

swap/src/bin/swap.rs
xscd: Done.

swap/src/bin/swap.rs
xscd: Thanks. At first I didn't wanted to change the function signature for this but this makes more sense to re-use already parsed things.

CHANGELOG.md
xscd: Ughh I've been a victimized by Prettier this time :smile: (https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)

CHANGELOG.md
xscd: Yeah I tried the `--json` flag to test it, its prints the logs in JSON format. This was a bad wording choice by me.

swap/src/bin/swap.rs
xscd: Agree. Removed it. We don't need even to join it manually the build method of qrcode crate already returns it as a String.

CHANGELOG.md
thomaseizinger: We use [`dprint`](https://dprint.dev/) to format things in this repo. If you've ever been annoyed by how slow prettier is, give it a go :)

thomaseizinger (approved):
Thanks! Looking good now :)

One last thing, do you mind squashing all your commits down into a single one?

xscd:
> One last thing, do you mind squashing all your commits down into a single one?

Done. It looks way cleaner now :smile:

PR #558, opened 2021-06-09
Remove websocket transport from CLI

thomaseizinger:
I've also incorporated some of the ideas in how to modularize the Tor integration with libp2p. For more details, please see the individual patches.

### DIFF
--- a/swap/src/cli/transport.rs
+++ b/swap/src/cli/transport.rs
@@ -1,0 +1,32 @@
+use crate::network::tor_transport::TorDialOnlyTransport;
+use crate::network::transport::authenticate_and_multiplex;
+use anyhow::Result;
+use libp2p::core::muxing::StreamMuxerBox;
+use libp2p::core::transport::{Boxed, OptionalTransport};
+use libp2p::dns::TokioDnsConfig;
+use libp2p::tcp::TokioTcpConfig;
+use libp2p::{identity, PeerId, Transport};
+
+/// Creates the libp2p transport for the swap CLI.
+///
+/// The CLI's transport needs the following capabilities:
+/// - Establish TCP connections
+/// - Resolve DNS entries
+/// - Dial onion-addresses through a running Tor daemon by connecting to the
+///   socks5 port. If the port is not given, we will fall back to the regular
+///   TCP transport.
+pub fn new(
+    identity: &identity::Keypair,
+    maybe_tor_socks5_port: Option<u16>,
+) -> Result<Boxed<(PeerId, StreamMuxerBox)>> {
+    let tcp = TokioTcpConfig::new().nodelay(true);
+    let tcp_with_dns = TokioDnsConfig::system(tcp)?;
+    let maybe_tor_transport = match maybe_tor_socks5_port {
+        Some(port) => OptionalTransport::some(TorDialOnlyTransport::new(port)),
+        None => OptionalTransport::none(),
+    };
+
+    let transport = maybe_tor_transport.or_transport(tcp_with_dns).boxed();
+
+    authenticate_and_multiplex(transport, identity)
+}
--- a/swap/src/network/tor_transport.rs
+++ b/swap/src/network/tor_transport.rs
@@ -1,4 +1,4 @@
-use anyhow::anyhow;
+use anyhow::{anyhow, Result};
 use data_encoding::BASE32;
 use futures::future::Ready;
 use futures::prelude::*;
@@ -76,7 +76,7 @@ impl Transport for TorTcpConfig {
 /// Tor expects an address format of ADDR:PORT.
 /// This helper function tries to convert the provided multi-address into this
 /// format. None is returned if an unsupported protocol was provided.
-fn to_address_string(multi: Multiaddr) -> anyhow::Result<String> {
+fn to_address_string(multi: Multiaddr) -> Result<String> {
     let mut protocols = multi.iter();
     let address_string = match protocols.next() {
         // if it is an Onion address, we have all we need and can return
--- a/swap/src/network/tor_transport.rs
+++ b/swap/src/network/tor_transport.rs
@@ -7,7 +7,7 @@ use libp2p::core::Transport;
 use libp2p::tcp::tokio::{Tcp, TcpStream};
 use libp2p::tcp::{GenTcpConfig, TcpListenStream, TokioTcpConfig};
 use std::io;
-use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::net::Ipv4Addr;
 use tokio_socks::tcp::Socks5Stream;
 
 /// Represents the configuration for a Tor transport for libp2p.
@@ -45,10 +45,12 @@ impl Transport for TorTcpConfig {
             Ok(tor_address_string) => Ok(async move {
                 tracing::trace!("Connecting through Tor proxy to address: {}", addr);
 
-                let sock = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, self.socks_port));
-                let stream = Socks5Stream::connect(sock, tor_address_string)
-                    .await
-                    .map_err(|e| io::Error::new(io::ErrorKind::ConnectionRefused, e))?;
+                let stream = Socks5Stream::connect(
+                    (Ipv4Addr::LOCALHOST, self.socks_port),
+                    tor_address_string,
+                )
+                .await
+                .map_err(|e| io::Error::new(io::ErrorKind::ConnectionRefused, e))?;
 
                 tracing::trace!("Connection through Tor established");
 
--- a/swap/src/network/tor_transport.rs
+++ b/swap/src/network/tor_transport.rs
@@ -1,7 +1,6 @@
 use anyhow::{anyhow, Result};
 use data_encoding::BASE32;
-use futures::future::Ready;
-use futures::prelude::*;
+use futures::future::{BoxFuture, Ready};
 use libp2p::core::multiaddr::{Multiaddr, Protocol};
 use libp2p::core::transport::TransportError;
 use libp2p::core::Transport;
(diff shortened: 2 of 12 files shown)


swap/src/cli/transport.rs
```diff
@@ -0,0 +1,28 @@
...
+use libp2p::{identity, PeerId, Transport};
+
+/// Creates the libp2p transport for the swap CLI.
+///
+/// The CLI's tranport needs the following capabilities:
+/// - Establish TCP connections
+/// - Resolve DNS entries
+/// - Dial onion-addresses through a running Tor daemon (by connecting to the
+///   socks5 port).
+pub fn new(
```
bonomat, line 17: I understand that by composing the transports in this way, it means that if `tor_dial_only` returns `MultiaddrNotSupported` it will fallback to `tcp_with_dns`.

If this is true (and if this is the wanted behavior), I would appreciate if we could rephrase the description here.

swap/src/network/tor_transport.rs
```diff
@@ -87,43 +72,33 @@ fn to_address_string(multi: Multiaddr) -> Result<String> {
...
-
-    let port_string = match protocols.next() {
-        Some(Protocol::Tcp(port)) => Some(format!("{}", port)),
-        Some(Protocol::Udp(port)) => Some(format!("{}", port)),
-        _ => None,
+        Some(Protocol::Ip4(addr)) => format!("{}", addr),
+        Some(Protocol::Ip6(addr)) => format!("{}", addr),
+        Some(Protocol::Dns(addr)) => format!("{}", addr),
+        Some(Protocol::Dns4(addr)) => format!("{}", addr),
+        _ => return Err(TransportError::MultiaddrNotSupported(multi)),
```
bonomat, line 79: I'm curious if this error will show up somewhere in the logs or if libp2p _silently_ falls back to the next transport? 
I'm curious because when this function initially was written I could not think of more or other protocols we want to support here.

bonomat (approved):
Looks good 👍

swap/src/network/tor_transport.rs
thomaseizinger: Yep, the fallback is silent, it happens here: https://github.com/libp2p/rust-libp2p/blob/762f4e35e72dbca4ce82dd5099dad7caf3016650/core/src/transport/choice.rs#L46-L60

Are you ok with the fallback being silent?

swap/src/network/tor_transport.rs
```diff
@@ -1,82 +1,67 @@
...
+    fn address_translation(&self, _: &Multiaddr, _: &Multiaddr) -> Option<Multiaddr> {
+        None
     }
 }
 
-/// Tor expects an address format of ADDR:PORT.
-/// This helper function tries to convert the provided multi-address into this
-/// format. None is returned if an unsupported protocol was provided.
-fn to_address_string(multi: Multiaddr) -> Result<String> {
+/// Extracts a Tor address string in the form of `{ADDR}.onion:{PORT}` from the
```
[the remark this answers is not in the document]
thomaseizinger: Right, I missed that. Will update the docs!

swap/src/network/tor_transport.rs
bonomat: Sure, it just means that we won't learn about other protocols we might have missed here.

swap/src/cli/transport.rs
thomaseizinger: Yep, you understand correctly. Do note that this was already the case behaviour, it was just implemented inside the tor transport instead of achieving it through composition here.

bonomat:
📯 
I tested this PR locally (both on the same machine) and swapping works. 
However, I'm not sure if was doing something wrong, but it looks like the ASB is not listening on the websocket anymore. 

logs: 
```bash
{"timestamp":"2021-06-09 13:20:45","level":"INFO","fields":{"onion_address":"/onion3/ktyp4lrhpayvlcxnvuc6a6fgiqod2ltkdr54iar7g7ig4bzpvpkg4pad:9939"}}
{"timestamp":"2021-06-09 13:20:47","level":"INFO","fields":{"message":"Our peer-id","peer_id":"12D3KooWDUrppfpALAVSpNnqDeeriT9rciXnfjdB1Ve2XzD3ffhb"}}
{"timestamp":"2021-06-09 13:20:47","level":"INFO","fields":{"message":"Listening on","address":"/ip4/192.168.1.63/tcp/9939"}}
{"timestamp":"2021-06-09 13:20:47","level":"INFO","fields":{"message":"Listening on","address":"/ip4/127.0.0.1/tcp/9939"}}
```

thomaseizinger:
>  However, I'm not sure if was doing something wrong, but it looks like the ASB is not listening on the websocket anymore.

Are you using a custom config file?

thomaseizinger:
It works for me using the default config file:

```
❯ cargo run --bin asb -- --testnet start
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `/home/thomas/.cache/cargo/target/debug/asb --testnet start`
2021-06-09 14:11:10  INFO Initialized tracing level=debug
2021-06-09 14:11:10  INFO Using config file at path=/home/thomas/.config/xmr-btc-swap/asb/testnet/config.toml
2021-06-09 14:11:10  INFO Database and Seed will be stored in db_folder=/home/thomas/.local/share/xmr-btc-swap/asb/testnet
2021-06-09 14:11:10 DEBUG Opening database at /home/thomas/.local/share/xmr-btc-swap/asb/testnet/database
2021-06-09 14:11:10 DEBUG Reading in seed from /home/thomas/.local/share/xmr-btc-swap/asb/testnet/seed.pem
2021-06-09 14:11:10 DEBUG Opening Bitcoin wallet
2021-06-09 14:11:13 DEBUG Opening Monero wallet
2021-06-09 14:11:18 DEBUG Created Monero wallet monero_wallet_name=asb-wallet
2021-06-09 14:11:19  INFO Initialized Bitcoin wallet bitcoin_balance=0.00000000 BTC
2021-06-09 14:11:19  WARN The Monero balance is 0, make sure to deposit funds at monero_address=478Uqx77ELxQhNQagh6Max26EDCU2SzLAgpCoL8z1ZX8RwmLnw1PFq9VYP5e2iVJdA9J1QM4LQxVVcBNrNPtmmRdTjBcAJG
2021-06-09 14:11:19  WARN Tor not found. Running on clear net
2021-06-09 14:11:19  INFO Network layer initialized peer_id=12D3KooWKD4nnoPBMyPUTQiwYj5FXvwVM7t62kqTQnLwkWDmoogS
2021-06-09 14:11:19  INFO Our peer-id peer_id=12D3KooWKD4nnoPBMyPUTQiwYj5FXvwVM7t62kqTQnLwkWDmoogS
2021-06-09 14:11:19  INFO Listening on address=/ip4/10.1.23.174/tcp/9940/ws
2021-06-09 14:11:19  INFO Listening on address=/ip4/10.1.23.174/tcp/9939
2021-06-09 14:11:19  INFO Listening on address=/ip4/127.0.0.1/tcp/9940/ws
2021-06-09 14:11:19  INFO Listening on address=/ip4/127.0.0.1/tcp/9939
2021-06-09 14:11:20 DEBUG Connected to Kraken websocket API
2021-06-09 14:11:20 DEBUG Subscribed to updates for ticker
```

swap/src/network/tor_transport.rs
thomaseizinger: We can add more logging here if that is what you are asking. Is your concern that a user might expect things to run via tor but because of us not handling some esoteric thing here it actually ends up being on clearnet?

That sounds more like we should either have a switch or a dedicated binary that enforces the Tor transport and removes the regular tcp transport entirely. That will result in an unrecoverable error if our Tor transport cannot handle some kind of address.

bonomat: > We can add more logging here if that is what you are asking. Is your concern that a user might expect things to run via tor but because of us not handling some esoteric thing here it actually ends up being on clearnet?

Yes, kinda. When I wrote the initial transport my thinking was: 
`If Tor is enabled, everything should go through Tor. But we don't want to application to be unusable if we have missed a protocol here. In that case, an error message should be printed which is hopefully reported to us. We then fall-back to TCP and continue.`

Maybe this wasn't the best behavior neither. 
Anyway: We can say we only support these protocols for Tor, everything else will go via clearnet. I still wouldn't know which one we are missing :)

bonomat:
> > However, I'm not sure if was doing something wrong, but it looks like the ASB is not listening on the websocket anymore.
> 
> Are you using a custom config file? Listening on WS will only happen if the application is instructed to do so. The default config file does that though.

Yes, can confirm. The custom config was the problem.

swap/src/network/tor_transport.rs
thomaseizinger: To be honest, I wouldn't count on users reporting warnings like this ...

For now, falling back silently errs on the side of being as compatible as possible. If we see risks there, I would suggest to default to Tor only and only provide a switch the user can actively trigger with which they give consent to fall back to clearnet if Tor doesn't work for some reason.

thomaseizinger:
Found a bug, please re-review the commits I pushed on top!

swap/src/network/tor_transport.rs
da-kami: I think the switch would actually be a good idea - that could even be part of the setup wizzard, something along the lines of "Do you want to allow connection over clearnet (or Tor only)?"

ASB providers do care about anonymity; if the application silently decides to go over clearnet, that is not optimal.

da-kami (approved):

thomaseizinger:
> I think the switch would actually be a good idea - that could even be part of the setup wizzard, something along the lines of "Do you want to allow connection over clearnet (or Tor only)?"
> ASB providers do care about anonymity; if the application silently decides to go over clearnet, that is not optimal. (I'm ok with merging this and tackle it in a follow up.)

Yep totally valid point. Keep in mind though that you are only looking at the CLI's transport here! At this stage, the ASB doesn't even know that it is running over Tor.

swap/src/network/tor_transport.rs
thomaseizinger: Yep totally valid point. Keep in mind though that you are only looking at the CLI's transport here! At this stage, the ASB doesn't even know that it is running over Tor.

thomaseizinger:
Rebased to resolve merge conflicts.

thomaseizinger:
bors retry

PR #582, opened 2021-06-22
Asb price guarantees

### LINKED ISSUE #557: Abuse of fixed XMR/BTC price
First of all, sorry to bother you again with this issue :D
I think Bob can abuse the fixed XMR/BTC price that both parties agree on when Bob connects to Alice. I think to fix this they need to agree on a timeout. If the BTC payment from Bob has not arrived after a certain time, the swap will be cancelled or they will agree on a new more current price.

Because the way it currently works Bob can:
1. connect to the ASB and agree on a price (for example: 0.1 BTC/XMR).
2. Bob can now decide for himself if and when he sends the BTC from his wallet. He now waits 1 hour until the BTC/XMR price reaches 0.15. If after an hour the price dropped Bob simply does not proceed with the swap because if he did he would lose money.
3. Since the two agreed on the price an hour ago, Alice now sells Bob Monero for a price far below the current market price. Bob can sell the Monero again somewhere else and make a profit. Alice has to deal with the implications.

da-kami:
Fixes #557 

Tested on test-/stagenet setup with Raspi.
The change log entry in the last commit sums up the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Printing the deposit address to the terminal as a QR code.
   To not break automated scripts or integrations with other software, this behaviour is disabled if `--json` is passed to the application.
 
+### Fixed
+
+- An issue where the ASB gives long price guarantees when setting up a swap.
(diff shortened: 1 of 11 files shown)


thomaseizinger (approved):
Looks good to me, base of this PR should be #584.

PR #585, opened 2021-06-23
Configurable kraken websocket url via the ASB config

cimble-code:
- Allows the ASB operator to configure a custom kraken websocket url via the ASB config. 
- Addresses the issue of price control first brought up [here](https://github.com/comit-network/xmr-btc-swap/discussions/571)

>  Gotya. 
There is a relatively easy to implement (but temporary) solution for that. We could let the user configure the kraken websocket url via the ASB config. That way you can plug in your own service. The only requirement is that your service publishes prices updates in the same format as [kraken](https://docs.kraken.com/websockets/), e.g. : 

_Originally posted by @bonomat in https://github.com/comit-network/xmr-btc-swap/discussions/571#discussioncomment-885535_

### DIFF
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -27,6 +27,7 @@ pub struct Defaults {
     listen_address_ws: Multiaddr,
     electrum_rpc_url: Url,
     monero_wallet_rpc_url: Url,
+    price_ticker_ws_url: Url,
     bitcoin_confirmation_target: usize,
 }
 
@@ -41,6 +42,7 @@ impl GetDefaults for Testnet {
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
             electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:60002")?,
             monero_wallet_rpc_url: Url::parse("http://127.0.0.1:38083/json_rpc")?,
+            price_ticker_ws_url: Url::parse("wss://ws.kraken.com")?,
             bitcoin_confirmation_target: 1,
         };
 
@@ -59,6 +61,7 @@ impl GetDefaults for Mainnet {
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
             electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:50002")?,
             monero_wallet_rpc_url: Url::parse("http://127.0.0.1:18083/json_rpc")?,
+            price_ticker_ws_url: Url::parse("wss://ws.kraken.com")?,
             bitcoin_confirmation_target: 3,
         };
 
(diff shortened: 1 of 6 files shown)


swap/src/asb/config.rs
```diff
@@ -250,6 +254,11 @@ pub fn query_user_for_initial_config(testnet: bool) -> Result<Config> {
         .default(defaults.monero_wallet_rpc_url)
         .interact_text()?;
 
+    let kraken_url = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Enter Kraken Price Ticker WebSocket URL or hit enter to use default")
+        .default(defaults.kraken_ws_url)
+        .interact_text()?;
+
```
bonomat, line 261: Personally I would prefer this to not to be in the wizard. 
If it is here, an ASB provider gets confronted with this setup not knowing what to do. 
I think it's more of an expert setup and configuring it via the config file directly is enough.

thomaseizinger: Agree with this, it is unlikely this needs to be configured on the first start.

swap/src/asb/config.rs
```diff
@@ -151,6 +154,7 @@ pub struct Maker {
     #[serde(with = "::bitcoin::util::amount::serde::as_btc")]
     pub max_buy_btc: bitcoin::Amount,
     pub ask_spread: Decimal,
+    pub kraken_ws_url: Url
```
thomaseizinger, line 157: Shall we name this something that doesn't include `kraken`?

My thinking is: there is only one correct `kraken_ws_url`: the one that was previously hardcoded. As such, a config option for that one doesn't make any sense as there is only one valid value.

What we are really configuring here should IMO be named something like `price_ticker_ws_url` and it needs to be documented that we are expecting the Kraken's price ticker protocol on this websocket.

thomaseizinger (changes requested):
I would like to discuss the name of the config value.

swap/src/asb/config.rs
cimble-code: https://github.com/comit-network/xmr-btc-swap/pull/585/commits/305f02a672b071582786f29ab1e8dac576f4fe4a

swap/src/asb/config.rs
cimble-code: https://github.com/comit-network/xmr-btc-swap/pull/585/commits/a222f4c64a4aa16b1adbe9703df58c64925625de

cimble-code:
> I approved CI to run. Please check the build failures :)

Should be fixed with https://github.com/comit-network/xmr-btc-swap/pull/585/commits/5b6ffae2715ea498648f4f218eede414f4116da0 and https://github.com/comit-network/xmr-btc-swap/pull/585/commits/b4f5587fe66dc6a54bc77980f19ba22510349baa

thomaseizinger (approved):
Looks good from my side, thank you!

Can we squash all commits into a single one please?

cimble-code:
> Can we squash all commits into a single one please?

Done :)

rishflab (approved):

thomaseizinger:
Sorry to bother you again but do you mind also adding a changelog entry under `Added`?

cimble-code:
> 
> 
> Sorry to bother you again but do you mind also adding a changelog entry under `Added`?
> 
> Squashing into the existing commit is preferable.

Sure, I edited CHANGELOG.md to include the changes.

thomaseizinger:
Thank you @cimble-code !

bors r+

PR #586, opened 2021-06-25
`swap_setup` instead of `spot_price` and `execution_setup`

da-kami:
Having `spot_price` and `execution_setup` as separate protocols did not bring any advantages, but was problematic because we had to ensure that `execution_setup` would be triggered after `spot_price`. Because of this dependency it is better to combine the protocols into one.

Combining the protocols also allows a refactoring to get rid of the `libp2p-async-await` dependency.

Alice always listens for the `swap_setup` protocol. When Bob opens a substream on that protocol the spot price is communicated, and then all execution setup messages (swap-id and signature exchange).


swap/src/protocol/alice/behaviour.rs
```diff
@@ -1,34 +1,34 @@
-use crate::network::quote::BidQuote;
-use crate::network::{encrypted_signature, quote, transfer_proof};
-use crate::protocol::alice::event_loop::LatestRate;
-use crate::protocol::alice::{execution_setup, spot_price, State3};
-use crate::{env, monero};
 use anyhow::{anyhow, Error};
 use libp2p::ping::{Ping, PingEvent};
 use libp2p::request_response::{RequestId, ResponseChannel};
 use libp2p::{NetworkBehaviour, PeerId};
 use uuid::Uuid;
```
thomaseizinger, line 6: With this newline, the formatter doesn't merge the statements.

swap/src/protocol/alice/event_loop.rs
```diff
@@ -150,77 +150,28 @@ where
...
-                            ) {
-                                (Ok(tx_redeem_fee), Ok(tx_punish_fee)) => {
-                                    (tx_redeem_fee, tx_punish_fee)
-                                }
-                                _ => {
-                                    tracing::error!(%peer, "Failed to calculate transaction fees during execution setup");
-                                    continue;
-                                }
-                            };
+                            let (btc, responder) = send_wallet_snapshot.recv().await.expect("TODO: handle error");
```
thomaseizinger, line 155: We should tackle this TODO.

swap/src/protocol/alice.rs
```diff
@@ -1,10 +1,12 @@
 //! Run an XMR/BTC swap in the role of Alice.
 //! Alice holds XMR and wishes receive BTC.
+use std::sync::Arc;
+
+use uuid::Uuid;
+
```
thomaseizinger, line 6: Same here with the newlines.

thomaseizinger (approved):
Feels good!

thomaseizinger (approved):

thomaseizinger:
bors retry

PR #587, opened 2021-06-25
Move files from `protocol` to appropriate module

da-kami:
Some network and application specific code does not belong in the protocol module and was moved.
Eventloop, recovery and the outside behaviour were moved to the respective application module because they are application specific.

The `swap_setup` was moved into the network module because upon change both sides will have to be changed and should thus stay close together.


swap/src/asb.rs
```diff
@@ -1,7 +1,19 @@
+pub use rate::Rate;
+
+pub mod behaviour;
 pub mod command;
 pub mod config;
+pub mod event_loop;
 mod rate;
+pub mod recovery;
```
thomaseizinger, line 8: If we re-export, can these be private now?

swap/src/asb.rs
```diff
@@ -1,7 +1,19 @@
+pub use rate::Rate;
```
thomaseizinger, line 1: Should be part of the other `pub use` statements.

swap/src/bin/asb.rs
```diff
@@ -26,12 +26,13 @@ use swap::asb::command::{parse_args, Arguments, Command};
 use swap::asb::config::{
     initial_setup, query_user_for_initial_config, read_config, Config, ConfigNotInitialized,
 };
+use swap::asb::event_loop::{EventLoop, KrakenRate};
+use swap::asb::recovery::redeem;
```
thomaseizinger, line 30: You are importing from the submodule even though we re-export. Making these modules private should allow us to be more consistent here.

swap/src/bin/swap.rs
```diff
@@ -12,29 +12,32 @@
...
-use prettytable::{row, Table};
-use qrcode::render::unicode;
-use qrcode::QrCode;
 use std::cmp::min;
 use std::env;
 use std::future::Future;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
+
```
thomaseizinger, line 21: Can we not have these newlines inbetween the use statements?

swap/src/bin/swap.rs
```diff
@@ -428,12 +431,15 @@ where
...
-    use crate::determine_btc_to_swap;
-    use ::bitcoin::Amount;
     use std::sync::Mutex;
+
+    use ::bitcoin::Amount;
     use tracing::subscriber;
 
+    use crate::determine_btc_to_swap;
+
+    use super::*;
```
thomaseizinger, line 441: Same here. Are you using "organize imports" from CLion? I saw that causing this at some point.

swap/src/cli/behaviour.rs
```diff
@@ -1,14 +1,17 @@
...
+use std::time::Duration;
+
 use anyhow::{anyhow, Error, Result};
 use libp2p::core::Multiaddr;
 use libp2p::ping::{Ping, PingEvent};
 use libp2p::request_response::{RequestId, ResponseChannel};
 use libp2p::{NetworkBehaviour, PeerId};
-use std::sync::Arc;
-use std::time::Duration;
+
```
thomaseizinger, line 9: Again here.

swap/src/network/redial.rs
```diff
@@ -7,12 +10,11 @@ use libp2p::core::Multiaddr;
 use libp2p::swarm::protocols_handler::DummyProtocolsHandler;
 use libp2p::swarm::{DialPeerCondition, NetworkBehaviour, NetworkBehaviourAction, PollParameters};
 use libp2p::PeerId;
-use std::pin::Pin;
-use std::task::{Context, Poll};
-use std::time::Duration;
 use tokio::time::{Instant, Sleep};
 use void::Void;
```
thomaseizinger, line 15: Also here.

swap/src/network/swap_setup.rs
```diff
@@ -1,10 +1,14 @@
-use crate::monero;
 use anyhow::{Context, Result};
 use libp2p::core::upgrade;
 use libp2p::swarm::NegotiatedSubstream;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
```
thomaseizinger, line 6: And here.

thomaseizinger (approved):
Looks good overall, I think making these modules private would help with consistency.

da-kami:
I managed to make some of the modules private. For the `cli`'s `cancel` and `refund` it did not make much sense, because I would have to re-export the complete module again (because we have `Error` in both that are used within the same file - could work around that, but not sure it's worth it...)

da-kami:
I'd merge this into the underlying branch first, then squash the fixups and then merge the underlying branch. Will do that Monday morning :)

thomaseizinger (approved):

da-kami:
changes merged into #586

PR #589, opened 2021-06-28
Create CONTRIBUTING.md

thomaseizinger:


CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+# Contribution guidelines
+
+Thank you for wanting to contribute to this project!
+
+## Contributing code
+
+There are a couple of things we are going to look out for in PRs and knowing them upfront is going to reduce the number of times we will be going back and forth, making things more efficient.
+
+1. We have CI checks in place that validate formatting and code style.
+   Make sure `dprint check` and `cargo clippy` both finish without any warnings or errors.
```
bonomat, line 10: You could add a reference to `dprint` since it is not a standard tool everyone has installed.

bonomat (approved):
👍

CONTRIBUTING.md
thomaseizinger: Good idea, will do!

da-kami (approved):
👍 thanks for putting this together.

We could also mention that ideas can be discussed via GitHub discussions. So if people have an idea rather than a full feature request they could open discussion threads. not sure it's worth mentioning this.

PR #591, opened 2021-06-28
Rendezvous cli

da-kami:
Very rough overview of the integration of rendezvous into the CLI to discover ASBs. 
Code is full of ToDos that are still to be done, but the rough idea of the integration should be clear.


swap/src/asb/event_loop.rs
```diff
@@ -149,9 +149,9 @@ where
 
         loop {
             tokio::select! {
-                swarm_event = self.swarm.next_event() => {
+                swarm_event = self.swarm.next() => {
```
thomaseizinger, line 152: `select_next_some` would do the unwrapping of the `Option` for you.

swap/src/bin/swap.rs
```diff
@@ -268,6 +273,86 @@ async fn main() -> Result<()> {
...
+            tor_socks5_port,
+        } => {
+            // TODO: Fix tracing to be able to operate without swap ID
+            cli::tracing::init(debug, json, data_dir.join("logs"), Uuid::new_v4())?;
+            let seed = Seed::from_file_or_generate(data_dir.as_path())
+                .context("Failed to read in seed file")?;
+
+            let identity = seed.derive_libp2p_identity();
+
+            // TODO: Compose behaviour that allows to requests Quotes as well
```
thomaseizinger, line 289: Very interesting idea, I like it.

swap/src/bin/swap.rs
```diff
@@ -268,6 +273,86 @@ async fn main() -> Result<()> {
...
+                    SwarmEvent::ConnectionEstablished { peer_id, .. }
+                        if peer_id == rendezvous_node_peer_id =>
+                    {
+                        tracing::info!(
+                            "Connected to rendezvous point, discovering nodes in '{}' namespace ...",
+                            namespace
+                        );
+
+                        swarm.behaviour_mut().rendezvous.discover(
+                            Some(Namespace::new(namespace.clone()).unwrap()),
```
thomaseizinger, line 309: I would hardcode the namespace here.

swap/src/bin/swap.rs
```diff
@@ -268,6 +273,86 @@ async fn main() -> Result<()> {
...
+                                {
+                                    address.clone().with(p2p_suffix)
+                                } else {
+                                    address.clone()
+                                };
+
+                                todo!("Dial Alice and request quote!")
+                            }
+                        }
+                    }
```
thomaseizinger, line 346: Some thoughts:

1. There could be more than one address per peer but we only want to display one quote.
2. We need to account for the fact that a connection might be unsuccessful, we shouldn't wait for that "forever". i.e. we can't just naively wait for 5 quote responses if we discovered 5 ASBs.

thomaseizinger (approved):
Looking good so far :)

swap/src/bin/swap.rs
da-kami: Maybe I misunderstand what you mean...
We can't really hardcode the namespace here, because it is blockchain network dependent. But we decide that when constructing the command already.

swap/src/bin/swap.rs
da-kami: I think this one was already outdated when I pushed it. I think the behaviour that composes `Rendezvous` and `Quote` together is fine for now.

swap/src/bin/swap.rs
da-kami: I did not think about the multiple addresses that could be available for one peer. Thanks for mentioning that. 

I propose to store a relation of 

```
PeerId -> Multiaddress(es) -> Quote status
```

So upon discovery we trigger a quote request for each address (I think we must do that, because otherwise it gets complicated to decide when to trigger the next address...).
Once we get a quote we record it for the peer-id and mark it as done. If we record another quote coming in for the same peer we ignore it. 
If we run into an unreachable we store that for the specific address of the peer. If all addresses are marked as "request failed" we know that this peer was generally unreachable. 

So the break conditions for the loop are: 
For each peer, either quote received for peer (don't care about the address for now), or all quote requests for all addresses failed. 

I think this should be complete, so we don't need an additional timeout around the complete loop.

swap/src/bin/swap.rs
thomaseizinger: I meant, that you should decide the namespace here, rather than passing it through to the command, i.e. as late as possible to make it obvious that this is not globally configurable.

swap/src/bin/swap.rs
thomaseizinger: > So upon discovery we trigger a quote request for each address (I think we must do that, because otherwise it gets complicated to decide when to trigger the next address...).

You will need to `dial` first, before you can request a quote. I think the best way to handle multiple addresses is to add all the addresses you discover to the quote protocol via `add_address`.

Then, you can send a quote request which will initiate a dial attempt (this is `libp2p-request-reponse` behaviour).

Eventually, you will either get one successful or failed event from `Quote` per Peer that you tried.

PR #593, opened 2021-06-29
Integrate rendezvous protocol

rishflab:
Open things:

- [x] Default parameters for CLI
- [x] Print a human-readable table of sellers if `--json` is not passed
- [x] Unresolved comments of review


swap/tests/harness/mod.rs
```diff
@@ -239,6 +243,8 @@ async fn start_alice(
     )
     .unwrap();
     swarm.listen_on(listen_address).unwrap();
+    swarm.add_external_address(listen_address, AddressScore::Infinite);
+    swarm.dial_addr(config.rendezvous_node.addr).unwrap();
```
thomaseizinger, line 247: Shouldn't need that in the test.

swap/src/bin/asb.rs
```diff
@@ -156,10 +157,19 @@ async fn main() -> Result<()> {
             for listen in config.network.listen {
                 Swarm::listen_on(&mut swarm, listen.clone())
                     .with_context(|| format!("Failed to listen on network interface {}", listen))?;
+                Swarm::add_external_address(&mut swarm, listen.clone(), AddressScore::Infinite)
+                    .with_context(|| format!("Failed to add external address {}", listen))?;
```
thomaseizinger, line 161: The address we are adding here should come from either a config or a commandline parameter.

swap/src/asb/event_loop.rs
```diff
@@ -262,6 +263,16 @@ where
...
                                 }
                             }
+
+                            if peer == self.rendezvous_node {
+                                // todo: remove unwraps
+                                let _ = self
+                                    .swarm
+                                    .behaviour_mut()
+                                    .rendezvous
+                                    .register(Namespace::new("xmr-btc-swap".to_string()).unwrap(), self.rendezvous_node, None)
```
thomaseizinger, line 273: Namespace to be chosen based on the enum that @da-kami is defining.

swap/src/asb/behaviour.rs
```diff
@@ -42,6 +44,13 @@ pub enum OutEvent {
         peer: PeerId,
         error: Error,
     },
+    Registered {
+        rendezvous_node: PeerId,
+        ttl: u64,
+        namespace: Namespace,
+    },
+    RegisterFailed(RegisterError),
+    RegisterExpired(Registration),
```
thomaseizinger, line 53: We will need our own timer for when to fire this. We might actually want to implement this as part of the rendezvous protocol itself. Even auto-renew should be possible with some config knobs.

Let's see how it pans out here.

thomaseizinger (commented):
Just some things I noticed.

swap/src/asb/behaviour.rs
da-kami: Given that this logic for now will be in the ASB, where do you think this logic should live?
I reckon for now we can just do this in the eventloop (out of code change simplicity), but in the long run should this logic be in the behaviour?

swap/src/asb/config.rs
```diff
@@ -299,18 +292,6 @@ pub fn query_user_for_initial_config(testnet: bool) -> Result<Config> {
     }
     let ask_spread = Decimal::from_f64(ask_spread).context("Unable to parse spread")?;
 
-    let rendezvous_peer_id_str = Input::with_theme(&ColorfulTheme::default())
```
da-kami: Removing this means that the user does not have a chance to configure this during the startup wizard - i.e. the registration with a rendezvous node would be a hidden feature that the user has to activate later by manually filling in config values.
Initially we said that rendezvous should be active by default. I think we should configure default values for it (at least once they are known) and not just omit it.

swap/src/asb/rendezvous.rs
```diff
@@ -0,0 +1,155 @@
...
+            self.events.push(NetworkBehaviourAction::DialAddress {
+                address: address_with_p2p,
+            })
+        }
+        Ok(())
+    }
+}
+
+type BehaviourInEvent =
+<<<libp2p::rendezvous::Rendezvous as NetworkBehaviour>::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::InEvent;
```
da-kami, line 78: Looks like there was a merge conflict here?

swap/src/asb/config.rs
rishflab: Ok I will add the defaults back

swap/src/asb/rendezvous.rs
rishflab: nope don't think so. It builds

da-kami:
> Ok, I will add the defaults back

Thanks - I agree that the current setup is a bit weird. Especially the combination of rendezvous configuration in the config but the external address as param. Maybe we should add the external address to be registered to the config as well. The problem is, that this address may be subject to change - or may not be known when we initially start the application. 

We could change the current setup to:

1. Ensure that seed exists before config is created
2. Come up with the libp2p identity prior to creating the initial config
3. Check if Tor is running, if so derive our onion address from the seed
4. Use the identities (libp2p, onion) for the default external address to be added to the config

What do you think?

swap/src/asb/rendezvous.rs
da-kami: sorry, may bad :)

rishflab:
> > Ok, I will add the defaults back
> 
> Thanks - I agree that the current setup is a bit weird. Especially the combination of rendezvous configuration in the config but the external address as param. Maybe we should add the external address to be registered to the config as well. The problem is, that this address may be subject to change - or may not be known when we initially start the application.
> 
> We could change the current setup to:
> 
> 1. Ensure that seed exists before config is created
> 2. ~Come up with the libp2p identity prior to creating the initial config~
> 3. Check if Tor is running, if so derive our onion address from the seed
> 4. Use the identities (~libp2p~, onion) for the default external address to be added to the config - if Tor the default suggested is the onion address. If clearnet we can't suggest an external address and will have to ask the user for it without a sane default 😅
> 
> What do you think?

How do we check if tor is running. Do we "inspect" the transport?

swap/src/asb/rendezvous.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+impl NetworkBehaviourEventProcess<libp2p::rendezvous::Event> for Behaviour {
+    fn inject_event(&mut self, event: Event) {
+        match event {
+            Event::RegisterFailed(error) => {
+                tracing::error!(rendezvous_node=%self.rendezvous_point_peer_id, "Registration with rendezvous node failed: {:#}", error);
+            }
+            Event::RegistrationExpired(registration) => {
```
da-kami, line 131: Why was this re-added? As I stated before, this is only relevant for the rendezvous node, not for the registration side. Any particular reason why it was added?

swap/src/asb/rendezvous.rs
```diff
@@ -0,0 +1,155 @@
...
+            rendezvous_point_peer_id: peer_id,
+            rendezvous_point_addr: addr,
+            rendezvous_namespace: namespace,
+            rendezvous_reregister_timestamp: None,
+            is_connected: false,
+            events: vec![],
+        }
+    }
+
+    pub fn refresh_registration(&mut self) -> Result<()> {
```
da-kami, line 48: Is there a way to encapsulate the complete refresh within the Behaviour?
I know the behaviour is not async, but I wonder if there is a way to still handle the registrations internally instead of needing the refresh from the outside?

rishflab: we could move it into the poll function, that might be a good idea

swap/src/asb/config.rs
```diff
@@ -81,18 +83,27 @@ fn default_asb_data_dir() -> Result<PathBuf> {
         .context("Could not generate default config file path")
 }
 
+// TODO: update this to the actual deployed rendezvous server
+//  Currently set to Staging ASB on raspi
+const DEFAULT_RENDEZVOUS_PEER_ID: &str = "12D3KooWPZ69DRp4wbGB3wJsxxsg1XW1EVZ2evtVwcARCF3a1nrx";
+// TODO: update this to the actual deployed rendezvous server
+//  Port still to be opened once running rendezvous node
+const DEFAULT_RENDEZVOUS_ADDR: &str = "/ip4/141.168.172.35/tcp/7654";
```
thomaseizinger, line 91: This will be `/dnsaddr/rendezvous.coblox.tech`.

swap/src/bin/swap.rs
```diff
@@ -280,6 +270,33 @@ async fn main() -> Result<()> {
...
+                namespace,
+                tor_socks5_port,
+                identity,
+            )
+            .await?;
+
+            for seller in sellers {
+                // TODO: Do not use tracing for printing this information
+                // TODO: Print on stdout
+                // TODO: Decide if the json flag should still be honoured
```
thomaseizinger, line 296: These need to be fixed before we can merge this.

swap/src/cli/command.rs
```diff
@@ -324,6 +347,24 @@ pub enum RawCommand {
...
+            help = "The peer-id of a rendezvous node that sellers register with"
+        )]
+        rendezvous_node_peer_id: PeerId,
+
+        // TODO: sane default value
+        #[structopt(
+            long,
+            help = "The multiaddr of a rendezvous node that sellers register with"
+        )]
+        rendezvous_node_addr: Multiaddr,
```
thomaseizinger, line 363: A single multi-addr should be good enough here.

For now we can assume that dialling the `/dnsaddr` will only return a single result. We will deal with that returning multiple results at another time.

swap/src/cli/event_loop.rs
```diff
@@ -94,19 +94,19 @@ impl EventLoop {
         loop {
             // Note: We are making very elaborate use of `select!` macro's feature here. Make sure to read the documentation thoroughly: https://docs.rs/tokio/1.4.0/tokio/macro.select.html
             tokio::select! {
-                swarm_event = self.swarm.next_event().fuse() => {
+                swarm_event = self.swarm.next().fuse() => {
```
thomaseizinger, line 97: Same here, use `select_next_some`.

swap/src/cli/list_sellers.rs
```diff
@@ -0,0 +1,270 @@
...
+        loop {
+            tokio::select! {
+                swarm_event = self.swarm.select_next_some() => {
+                    match swarm_event {
+                        SwarmEvent::ConnectionEstablished { peer_id, endpoint, .. } => {
+                            if peer_id == self.rendezvous_peer_id{
+                                tracing::info!(
+                                    "Connected to rendezvous point, discovering nodes in '{}' namespace ...",
+                                    self.namespace
+                                );
```
thomaseizinger, line 129: This feels more like a `debug` to me.

swap/src/cli/list_sellers.rs
```diff
@@ -0,0 +1,270 @@
...
+                        }
+                        SwarmEvent::Behaviour(OutEvent::Rendezvous(
+                                                  rendezvous::Event::Discovered { registrations, .. },
+                                              )) => {
+                            self.state = State::WaitForQuoteCompletion;
+
+                            for registration in registrations {
+                                let peer = registration.record.peer_id();
+                                for address in registration.record.addresses() {
+                                    tracing::info!("Discovered peer {} at {}", peer, address);
```
thomaseizinger, line 171: Same here, move this to `debug`.

swap/src/cli/list_sellers.rs
```diff
@@ -0,0 +1,270 @@
...
+                                                self.asb_quote_status.remove(&peer);
+                                            }
+                                        }
+                                        RequestResponseMessage::Request { .. } => unreachable!()
+                                    }
+                                }
+                                RequestResponseEvent::OutboundFailure { peer, error, .. } => {
+                                    if peer == self.rendezvous_peer_id {
+                                        tracing::debug!(%peer, "Outbound failure when communicating with rendezvous node: {:#}", error);
+                                    } else {
```
thomaseizinger, line 208: Can this actually happen?

thomaseizinger: I think we can just always call `self.asb_quote_status.remove` here.

swap/src/rendezvous.rs
```diff
@@ -0,0 +1,16 @@
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug, PartialEq)]
+pub enum XmrBtcNamespace {
+    Mainnet,
+    Testnet,
+}
+
+impl Display for XmrBtcNamespace {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
thomaseizinger, line 10: Inconsistent use of import vs fully-qualified. I always import `std::fmt` and qualify from there. Strikes a good balance between number of import lines, noise in the implementation and clarity of what is going on.

swap/Cargo.toml
```diff
@@ -29,7 +29,7 @@ ecdsa_fun = { git = "https://github.com/LLFourn/secp256kfun", default-features =
 ed25519-dalek = "1"
 futures = { version = "0.3", default-features = false }
 itertools = "0.10"
-libp2p = { version = "0.38", default-features = false, features = [ "tcp-tokio", "yamux", "mplex", "dns-tokio", "noise", "request-response", "websocket", "ping" ] }
+libp2p = { git = "https://github.com/comit-network/rust-libp2p", rev = "4cacaf215b334d447faca8f4e361fe19c30d2162", default-features = false, features = [ "tcp-tokio", "yamux", "mplex", "dns-tokio", "noise", "request-response", "websocket", "ping", "rendezvous" ] }
```
thomaseizinger, line 32: Would prefer if we let the lockfile handle the actual revision.

thomaseizinger (commented):
Some thoughts.

swap/src/asb/event_loop.rs
```diff
@@ -148,10 +165,34 @@ where
...
+                    } else {
+                        match Swarm::dial_addr(&mut self.swarm, self.rendezvous_node_addr.clone()) {
+                            Ok(()) => {}
+                            Err(error) => {
+                                tracing::error!(
+                                    "Failed to redial rendezvous node for re-registration: {:#}",
+                                    error
+                                );
+                            }
+                        }
```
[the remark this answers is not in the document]
da-kami: There is no `dial_peer` - I reckong you mean `dial` (using only the peer ID) - if we do that we first have to associate the address of the peer with the peer_id. I could not figure out how that is to be done here.

For e.g. request-response behaviour we can do an `add_address` but I could not figure out how to achieve that on the swarm level without using `dial_addr`.

swap/src/cli/command.rs
thomaseizinger: Actually, by appending `/p2p` to it, only a single address will be dialed. See what I did for the ASB.

swap/src/bin/swap.rs
```diff
@@ -270,6 +270,41 @@ async fn main() -> Result<()> {
...
+                rendezvous_node_addr,
+                namespace,
+                tor_socks5_port,
+                identity,
+            )
+            .await?;
+
+            for seller in sellers {
+                if json {
+                    println!("{}", serde_json::to_string(&seller)?);
```
rishflab, line 295: should we change this to use tracing?

swap/src/asb/config.rs
```diff
@@ -285,12 +293,25 @@ pub fn query_user_for_initial_config(testnet: bool) -> Result<Config> {
     }
     let ask_spread = Decimal::from_f64(ask_spread).context("Unable to parse spread")?;
 
+    let rendezvous_address = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you want to advertise your ASB instance with a rendezvous node? Enter an empty string if not.")
+        .default(DEFAULT_RENDEZVOUS_ADDRESS.to_string())
+        .interact_text()?;
+
+    let rendezvous_point = if rendezvous_address.is_empty() {
```
da-kami, line 301: Does that make sense? Can it actually be empty?
If it is empty (i.e. the user just hits enter), wouldn't it automatically apply the default value?

swap/src/cli/command.rs
```diff
@@ -324,6 +347,24 @@ pub enum RawCommand {
         #[structopt(flatten)]
         bitcoin: Bitcoin,
     },
+    ListSellers {
+        // TODO: sane default value
+        #[structopt(
+            long,
+            help = "The peer-id of a rendezvous node that sellers register with"
+        )]
+        rendezvous_node_peer_id: PeerId,
```
rishflab, line 356: I think we should make this optional and default to asb default. CLI users shouldn't have to deal with this IMO. Its a small change.

swap/src/asb/network.rs
```diff
@@ -138,4 +162,280 @@ pub mod behaviour {
...
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+        use crate::network::test::{new_swarm, SwarmExt};
+        use futures::StreamExt;
+        use libp2p::swarm::SwarmEvent;
+
+        #[tokio::test]
+        async fn given_no_initial_connection_when_constructed_asb_connects_and_registers_with_rendezvous_node(
```
da-kami, line 353: Can we somehow test that we register only once and not bombard the rendezvous node? 
That was one problem of the previous impl using a separate behaviour and I don't see it covered with this test. Not sure how easily it can be added.

swap/src/bin/swap.rs
thomaseizinger: No, the idea is that this output is on stdout so it can easily be used in scripts through piping.

swap/src/asb/config.rs
thomaseizinger: I never actually tried :joy:

I was hoping that entering `""` would return an empty string and not the default.

swap/src/asb/network.rs
thomaseizinger: I am pretty confident the current implementation doesn't do that but I can also see to add a test yes.

thomaseizinger:
ACK https://github.com/comit-network/xmr-btc-swap/pull/593/commits/021288850e5bc3bac4cb87c7add6372ebbc6c6eb

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Configuration setting for the websocket URL that the ASB connects to in order to receive price ticker updates.
   Can be configured manually by editing the config.toml file directly.
   It is expected that the server behind the url follows the same protocol as the [Kraken websocket api](https://docs.kraken.com/websockets/).
+- Registration and discovery of ASBs using the [libp2p rendezvous protocol](https://github.com/libp2p/specs/blob/master/rendezvous/README.md).
+  ASBs can register with a rendezvous node upon startup and, once registered, can be automatically discovered by the CLI using the `list-sellers` command.
+  The rendezvous node address (`rendezvous_point`), as well as the ASB's external addresses (`external_addresses`) to be registered, is to be configured in the `network` section of the ASB config file.
```
rishflab, line 19: nitpick: "is ~~to be~~ configured in the `network` section of the ASB config file."

da-kami:
OK, all fixups are in, rebasing on https://github.com/comit-network/xmr-btc-swap/pulls now and re-use the `extract_peer_id` function.

thomaseizinger:
ACK https://github.com/comit-network/xmr-btc-swap/pull/593/commits/b2efbbdb52b93dfbba87efca896dc4cc4641b96e and https://github.com/comit-network/xmr-btc-swap/pull/593/commits/5ca9947d565e38121c383963ec7b7a4f152594d4. Good to be fixup'ed.

thomaseizinger:
https://github.com/comit-network/xmr-btc-swap/pull/593/commits/43bff7ee338a8f6b0b2c4886498144fdc8313cfb Could also be a fixup into https://github.com/comit-network/xmr-btc-swap/pull/593/commits/af76c4b50be45dd9f4461bd20fa05f4f59857a89 maybe? Changes code that was introduced in this PR.

Should be easy to move https://github.com/comit-network/xmr-btc-swap/pull/593/commits/984b8c72f3dad7e63b1c622a8f02c7bb6265667c to the front of the PR.

PR #597, opened 2021-07-02
Rendezvous asb refactor

da-kami:
Refactor of the rendezvous integration into the ASB that introduces a separate behaviour. 

At the moment registrations don't work - I think the problem is how we connect to the rendezvous node. 
Establishing a connection works, but when we send the registration from ASB to rendezvous node it never arrives. 
We might want to change the way we dial and establish the connection.

Furthermore I propose to change the refresh model to be handled internally, similar to what we do here:

https://github.com/comit-network/rust-libp2p/blob/4cacaf215b334d447faca8f4e361fe19c30d2162/protocols/rendezvous/src/behaviour.rs#L580-L582


thomaseizinger:
Has been largely integrated into https://github.com/comit-network/xmr-btc-swap/pull/593 already.

PR #609, opened 2021-07-06
Replace xmr.to stagenet node with melo.tools node

binarybaron:
XMR.to has abandoned their community services.


rishflab:
seems legit https://www.reddit.com/r/Monero/comments/oersld/melotools_hub_of_free_monero_tools/

thanks for the patch!

binarybaron:
👍

PR #614, opened 2021-07-08
Fix quick start README.md list-sellers command

binarybaron:
Change list-sellers from flag to command


thomaseizinger (approved):
Thanks!

bors r+

PR #626, opened 2021-07-09
Changelog fix: Rendezvous point is not defaulted

thomaseizinger:


da-kami:
Careful, I think Murphy is around the corner!

thomaseizinger:
> Careful, I think Murphy is around the corner!

He seems to be running in circles by the looks of it: https://github.com/comit-network/xmr-btc-swap/pull/626/checks?check_run_id=3025652540

da-kami (approved):

PR #643, opened 2021-08-12
Upgrade to latest bdk to fix occasional `InsufficientFunds` error

### LINKED ISSUE #546: We sometimes hit `InsufficientFunds` when funding Bitcoin LockTx
I think this happens because we use random keys to predict the weight of `TxLock`. Instead, we should probably hardocde a meaningful weight of `TxLock` and always use that when calculating [`max_giveable`](https://github.com/comit-network/xmr-btc-swap/blob/d5d0dda6e73c76911edcae513a0333ca11a35330/swap/src/bin/swap.rs#L102).

thomaseizinger:
Fixes #546.
Please review patch-by-patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- An occasional error where users couldn't start a swap because of `InsufficientFunds` that were off by exactly 1 satoshi.
+
 ## [0.8.0] - 2021-07-09
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,10 +236,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
-name = "bdk"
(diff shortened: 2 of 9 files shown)


thomaseizinger:
bors retry

da-kami (commented):
Thanks for adding the proptests to reproduce the error - and then fixing it by upgrading. 

The test that fails upon bors trying to merge is `can_override_change_address` - I don't think that the specific changes to the test should cause any problem - I think the test failure is related to the changes added in the last commit. (Cannot pinpoint the exact problem without debugging though.)

Do I understand this correctly, that the last commit is just a refactoring on top of the fix? If so I would suggest to drop the last commit and potentially introduce these changes in a follow up.

thomaseizinger:
> Thanks for adding the proptests to reproduce the error - and then fixing it by upgrading.
> 
> The test that fails upon bors trying to merge is `can_override_change_address` - I don't think that the specific changes to the test should cause any problem - I think the test failure is related to the changes added in the last commit. (Cannot pinpoint the exact problem without debugging though.)
> 
> Do I understand this correctly, that the last commit is just a refactoring on top of the fix? If so I would suggest to drop the last commit and potentially introduce these changes in a follow up.

Yeah that is what I am planning on doing :)
Thought it would be nice to clean this up a bit but seems like I've made some mistake so I am not gonna bother with that and just drop it!

da-kami (approved):

PR #678, opened 2021-08-24
Default to quad9 DNS configuration instead of system

thomaseizinger:
Relying on the system DNS configuration causes problems for some
users. Let's try defaulting to quad9's config.


rishflab (approved):
lgtm

thomaseizinger:
Hold on with this one please, I was only putting this up for someone to test if DNS is a problem or not. I think we shouldn't just default to certain DNS settings but leave it to the user to configure their system properly.

Issue #662, opened 2021-08-21
ASB withdraws one output at a time instead of emptying wallet


sethsimmons:
When attempting to empty the ASB Bitcoin wallet via withdraw-BFF, the documented behavior is that if no amount is provided it will empty the wallet.

ATM the behavior is instead withdrawing one output at a time from the wallet.

thomaseizinger:
That is certainly a bug although I don't quite know why that would happen.

thomaseizinger:
[This](https://github.com/comit-network/xmr-btc-swap/blob/42125049413ea2d1fbf389ad9ff0642ea02d6aba/swap/src/bin/asb.rs#L216-L234) is where the withdraw is initiated. To investigate, I would recommend starting with adding a test [here](https://github.com/comit-network/xmr-btc-swap/blob/master/swap/src/bitcoin/wallet.rs#L894) where we initialize the test wallet with > 1 UTXO.

rishflab:
resolved in PR

PR #702, opened 2021-08-29
Fetching all UTXOs when doing max_giveable

devbordecraft:
This is fixing #662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An issue where the ASB was unable to use the Monero wallet in case `monero-wallet-rpc` has been restarted.
   In case no wallet is loaded when we try to interact with the `monero-wallet-rpc` daemon, we now load the correct wallet on-demand.
   See issue https://github.com/comit-network/xmr-btc-swap/issues/652.
+- An issue where the ASB withdraws one bitcoin UTXO at a time instead of the whole balance.
+  See issue https://github.com/comit-network/xmr-btc-swap/issues/662
 
 ## [0.8.1] - 2021-08-16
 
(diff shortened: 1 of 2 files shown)


rishflab (approved):

rishflab (changes requested):
Thanks again. I think this is good to go apart from one thing:

It might be worth  adding a note to the changelog as this change will mean that if you start a swap and you have a leftover UTXO from a previous failed swap, your new swap will include both UTXO's in bitcoin TxLock.

thomaseizinger:
> It might be worth adding a note to the changelog as this change will mean that if you start a swap and you have a leftover UTXO from a previous failed swap, your new swap will include both UTXO's in bitcoin TxLock.

Does that matter much though? We can include it as a `Fixed` as well because I would actually consider it a bug that we weren't doing that currently :)

rishflab:
> > It might be worth adding a note to the changelog as this change will mean that if you start a swap and you have a leftover UTXO from a previous failed swap, your new swap will include both UTXO's in bitcoin TxLock.
> 
> Does that matter much though? We can include it as a `Fixed` as well because I would actually consider it a bug that we weren't doing that currently :)

My memory of the the user interaction is a bit foggy but as far as I remember Bob is prompted to deposit BTC to swap-cli wallet and is under the impression that the amount he deposits is the amount that is swaped. This PR would mean bob's swap more than what he would expect to swap?

thomaseizinger:
> > > It might be worth adding a note to the changelog as this change will mean that if you start a swap and you have a leftover UTXO from a previous failed swap, your new swap will include both UTXO's in bitcoin TxLock.
> > 
> > Does that matter much though? We can include it as a `Fixed` as well because I would actually consider it a bug that we weren't doing that currently :)
> 
> My memory of the the user interaction is a bit foggy but as far as I remember Bob is prompted to deposit BTC to swap-cli wallet and is under the impression that the amount he deposits is the amount that is swaped. This PR would mean bob's swap more than what he would expect to swap?

If there is money left in the wallet, we don't prompt at all. Even dodgier but that is the current behaviour 😅

rishflab:
@devbordecraft recreated this here #717 preserving your commits. just wanted to get it into release 0.8.2 asap and couldn't wait for further back and forth over merge conflicts/ minor changes.

PR #705, opened 2021-08-30
Handle Results of waiting for timelocks to expire

da-kami:
Fixes an issue where the CLI is transitioning into a `CancelTimelockExpired` state but the timelock is actually not expired yet. 
This happened because watching for confirmations errored and the error was not handled but the same behaviour as reaching the cancel timelock height was applied.

For now this PR will fail the execution in case of the block height watching logic failing. Eventually there should be a more resilient strategy that includes retries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   See issue https://github.com/comit-network/xmr-btc-swap/issues/652.
 - An issue where swap protocol was getting stuck trying to submit the cancel transaction.
   We were not handling the error when TxCancel submission fails.
-  See issues: https://github.com/comit-network/xmr-btc-swap/issues/709, https://github.com/comit-network/xmr-btc-swap/issues/688, https://github.com/comit-network/xmr-btc-swap/issues/701.
+  We also configured the electrum client to retry 5 times in order to help with this problem.
+  See issues: https://github.com/comit-network/xmr-btc-swap/issues/709 https://github.com/comit-network/xmr-btc-swap/issues/688, https://github.com/comit-network/xmr-btc-swap/issues/701.
 
 ## [0.8.1] - 2021-08-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An issue where the ASB was unable to use the Monero wallet in case `monero-wallet-rpc` has been restarted.
   In case no wallet is loaded when we try to interact with the `monero-wallet-rpc` daemon, we now load the correct wallet on-demand.
   See issue https://github.com/comit-network/xmr-btc-swap/issues/652.
+- An issue where swap protocol was getting stuck trying to submit the cancel transaction.
+  We were not handling the error when TxCancel submission fails.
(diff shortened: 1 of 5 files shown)


rishflab (approved):

swap/src/protocol/bob/swap.rs
```diff
@@ -175,10 +177,14 @@ async fn next_state(
                 // Bob sends Alice his key
 
                 select! {
-                    _ = event_loop_handle.send_encrypted_signature(state.tx_redeem_encsig()) => {
+                    result = event_loop_handle.send_encrypted_signature(state.tx_redeem_encsig()) => {
+                        // TODO: Evaluate if that should bail?
```
thomaseizinger, line 181: What is the state of this TODO?

thomaseizinger (approved):
Nice. Can we either remove this TODO or do something else about it?

thomaseizinger (changes requested):
Needs a changelog entry.

da-kami:
@rishflab please take this one over

swap/src/protocol/bob/swap.rs
da-kami: I am not sure how we should react to this one - in the long run we should definitely have some retry mechanism here. For now we can `bail` - I marked this one because it is different than the other cases where we waited for timelocks to expire.

CHANGELOG.md
```diff
@@ -9,11 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  The CLI expects to be connected to the ASB throughout the entire swap and hence reconnects as soon as the connection is closed.
-  This resulted in a loop of connections being established but instantly closed again because the ASB deemed the connection to not be necessary.
+- An issue where the connection between ASB and CLI would get closed prematurely. The CLI expects to be connected to the
+  ASB throughout the entire swap and hence reconnects as soon as the connection is closed. This resulted in a loop of
+  connections being established but instantly closed again because the ASB deemed the connection to not be necessary.
   See issue https://github.com/comit-network/xmr-btc-swap/issues/648.
 
+- An issue where swap protocol was getting stuck trying to submit the cancel transaction. We were not handling the error
+  when TxCancel submission fails. See issues: https://github.com/comit-network/xmr-btc-swap/issues/709
+  , https://github.com/comit-network/xmr-btc-swap/issues/688, https://github.com/comit-network/xmr-btc-swap/issues/701
```
thomaseizinger: Not critical but these formatting changes seem to be unnecessary :)

rishflab: hmmm wonder how this happened

thomaseizinger: Let's fix it in a follow-up. Would be good to get a release out with this fix!

thomaseizinger:
@rishflab one of the tests in bors failed. Could take this opportunity to fixup the commits now that we have to retrigger anyway.

swap/src/protocol/bob/swap.rs
thomaseizinger: @rishflab Can we remove this hunk from this commit and just fix the handling of `send_encrypted_signature` entirely in the 3rd patch?

rishflab:
> @rishflab one of the tests in bors failed. Could take this opportunity to fixup the commits now that we have to retrigger anyway.

im on it. It was Monero again

thomaseizinger:
Thanks for improving the commits!

PR #721, opened 2021-09-02
Make --force the default behaviour for manual recovery

rishflab:
Also removed the --force flag


thomaseizinger:
@rishflab Please re-request once ready to review :)

swap/src/asb/recovery/refund.rs
```diff
@@ -105,7 +67,9 @@ pub async fn refund(
         state3.extract_monero_private_key(published_refund_tx)?
     } else {
         let bob_peer_id = db.get_peer_id(swap_id)?;
-        return Ok(Err(Error::RefundTransactionNotPublishedYet(bob_peer_id)));
+        return Err(anyhow::Error::from(
```
thomaseizinger, line 70: `bail!` will do the return + error conversion for you :)

swap/src/cli/command.rs
```diff
@@ -376,25 +370,21 @@ enum RawCommand {
...
         swap_id: SwapId,
 
-        #[structopt(short, long)]
-        force: bool,
-
         #[structopt(flatten)]
         bitcoin: Bitcoin,
     },
-    /// Try to cancel a swap and refund the BTC (expert users only)
+    /// Force submission of the cancel transaction overriding the protocol state
```
thomaseizinger, line 382:
```suggestion
    /// Force submission of the refund transaction overriding the protocol state
```

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command.rs
```diff
@@ -61,40 +60,20 @@ async fn given_alice_and_bob_manually_refund_after_funds_locked_both_refund() {
...
-            alice_swap.bitcoin_wallet,
-            alice_swap.db,
-            false,
-        )
-        .await??;
-
         // manually refund ALice's swap
         ctx.restart_alice().await;
         let alice_swap = ctx.alice_next_swap().await;
-        assert!(matches!(alice_swap.state, AliceState::BtcCancelled { .. }));
```
thomaseizinger, line 89: Is just deleting this okay? Shouldn't we still reach the same state?

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -38,13 +38,8 @@ async fn given_alice_and_bob_manually_cancel_when_timelock_not_expired_errors()
...
         // Bob tries but fails to manually cancel
-        let result = cli::cancel(bob_swap.id, bob_swap.bitcoin_wallet, bob_swap.db, false)
-            .await?
-            .unwrap_err();
-        assert!(matches!(
-            result,
-            cli::cancel::Error::CancelTimelockNotExpiredYet
-        ));
+        let result = cli::cancel(bob_swap.id, bob_swap.bitcoin_wallet, bob_swap.db).await;
+        assert!(result.is_err());
```
thomaseizinger, line 42: Can we assert against the particular string? You should be able to downcast this to an Electrum protocol error.

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired.rs
```diff
@@ -54,29 +49,18 @@ async fn given_alice_and_bob_manually_cancel_when_timelock_not_expired_errors()
...
-        )
-        .await?
-        .unwrap_err();
-        assert!(matches!(
-            result,
-            asb::cancel::Error::CancelTimelockNotExpiredYet
-        ));
+        let result =
+            asb::cancel(alice_swap.swap_id, alice_swap.bitcoin_wallet, alice_swap.db).await;
+        assert!(result.is_err());
```
thomaseizinger, line 54: Same here, lets assert against the actual error from electrum please.

thomaseizinger (changes requested):
Thanks! Some minor comments!

swap/tests/alice_and_bob_refund_using_cancel_and_refund_command.rs
rishflab: I removed `asb::cancel()` which was ensuring that Alice is in the `BtcCancelled` state. Since it has been removed Alice does not proceed past `XmrLockTransactionSent`.

thomaseizinger: Are you saying there is no point in doing `asb::cancel` because it has already been performed by the CLI and hence it would fail whereas previously it would just gracefully do nothing?

rishflab: pretty much. It would advance the CLI state machine to the BtcCancelledState

thomaseizinger: What is the resulting user behaviour now? If a user runs cancel but it has already been cancelled, shouldn't we transition to `BtcCancelled` as part of that?

In other words, is our approach now that these manual commands never save state to the DB unless they are successful?

rishflab: I think the saving to the db behaviour is unchanged in these manual commands?  I think the main change is that `asb::cancel()` is erroring if cancel has already been published which is wrong. `cli::cancel()` is performing this check. I should be able to reintroduce this assertions after I fix that.

thomaseizinger: Well, it may be okay to assume that the manual commands (which effectively forward to `asb::{cancel, refund, ...}`) can fail with all sorts of errors if they don't work out. i.e. we could say:

`asb cancel` can fail if cancel is already published and it will not transition the internal state.

I think that is reasonable and may actually be the best choice. It makes those commands very predictable. If `cancel` fails because `cancel` is already published, the user can always run `punish` and it will work and transition the state to `punished`. To actually step through all the states `resume` should be used as you documented in the commit message!

rishflab: yeh ok. Ill had a info log statement so the user isn't confused and tries to resubmit cancel

swap/src/asb/recovery/cancel.rs
```diff
@@ -42,8 +41,15 @@ pub async fn cancel(
         | AliceState::SafelyAborted => bail!("Swap is is in state {} which is not cancelable", state),
     };
 
-    // TODO: handle these error variants?
-    let txid = state3.submit_tx_cancel(bitcoin_wallet.as_ref()).await?;
+    let txid = match state3.submit_tx_cancel(bitcoin_wallet.as_ref()).await {
+        Ok(txid) => txid,
+        Err(err) => {
+            if let Some(bdk::Error::TransactionConfirmed) = err.downcast_ref::<bdk::Error>() {
```
thomaseizinger, line 47: Are you sure we are actually hitting this variant? Aren't all the electrum errors crammed into `bdk::Error::Electrum(bdk::electrum_client::Error::Protocol(..))`?

rishflab: not sure. I think having the check was actually better. Downcasting and parsing the error seems fragile.  First I will have to figure out the string that corresponds and then check  `bdk::electrum_client::Error::Protocol(String)` if it matches. And also hope that it doesn't change across versions/implementations.

rishflab: Maybe I could try and submit tx_cancel. If it fails, I check if it has already been submitted and then log that "cancel has been published". This minimizes the amount of things that can go wrong before submission of tx cancel but is also robust when determining if the submission failed because tx cancel has already been submitted.

thomaseizinger: To be perfectly honest, I think checking the error is better because it is an atomic operation. There is no guarantee that the state of the external system is the same between your two RPC calls. Assuming Electrum is implemented in a sane way, parsing the error will let you know why things didn't work. It is a reasonable assumption that the API will not undergo breaking changes!

swap/src/bitcoin.rs
```diff
@@ -249,6 +249,61 @@ pub fn current_epoch(
...
+    };
+
+    let error_code_number = match error_code_value {
+        serde_json::Value::Number(num) => num,
+        _ => bail!("Error code is not a number"),
+    };
+
+    if *error_code_number != serde_json::Number::from(i64::from(expected_error_code)) {
+        bail!("Unexpected error code: {}", error_code_number)
+    }
```
thomaseizinger, line 302: Minor nit: I think it would be slightly cleaner to just return the parsed error code from this function and allow the caller to decide what they want to do with it. Gives you a bit more flexibility in how it can be used.

Instead of:

```rust
if ok_if_rpc_error_code(&err, RpcErrorCode::RpcVerifyAlreadyInChain).is_ok() {

}
```

You would say:

```rust
if parse_rpc_error_code(&err)? == RpcErrorCode::RpcVerifyAlreadyInChain {

}
```

thomaseizinger (approved):
Thanks! Concept ACK

Commit history needs a bit of cleanup plus you need to remove the deleted test from the `bors.toml` configuration file and from the CI workflow.

.github/workflows/ci.yml
```diff
@@ -99,7 +99,6 @@ jobs:
             happy_path_restart_alice_after_xmr_locked,
             alice_and_bob_refund_using_cancel_and_refund_command,
             alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired,
-            alice_and_bob_refund_using_cancel_and_refund_command_timelock_not_expired_force,
```
thomaseizinger, line 102: You need to delete it from the `bors.toml` as well: https://github.com/comit-network/xmr-btc-swap/blob/7be6bcd4f5747bcce5ddb9d628eebe4d5da045ab/bors.toml#L15

thomaseizinger: Otherwise bors is going to timeout because it will wait for the status forever.

PR #780, opened 2021-09-16
Use domain types in database API

rishflab:


swap/src/protocol/bob/swap.rs
```diff
@@ -41,9 +40,8 @@ pub async fn run_until(
         )
         .await?;
 
-        let db_state = current_state.clone().into();
         swap.db
-            .insert_latest_state(swap.id, Swap::Bob(db_state))
+            .insert_latest_state(swap.id, &current_state)
```
thomaseizinger, line 44: If we are saving `current_state` directly now, doesn't this mean we would be able to delete `database::Swap`?

thomaseizinger (commented):
We can do this but bear in mind:

1. You are changing the serialization format here, meaning this code will not be able to read old database entries.
2. You are removing an abstraction layer between the core domain and the storage layer. This makes almost every change on the domain layer a breaking change on the persistence layer. Even though it might be slightly annoying at times, mapping into a dedicated DB model allows you to catch these differences and deal with them in a backwards-compatible way. For example, you can rename a field on the domain layer but still use the old name in the DB.
3. In Rust, it is tempting to assume that compiling means that the code works. Renaming a field will not cause a compile error when you rename it everywhere but it will break at runtime if you try to read old data. Thus, having a separate model on all system boundaries (APIs, DB, etc) is good if you care about backwards-compatibility as it makes it much easier to audit, whether or not a particular change is breaking. Instead of having to think about this on every change to the model, only renames / type changes to the specific wire/DB model will be breaking and a conversion layer will catch the rest.

swap/src/database.rs
```diff
@@ -43,36 +47,56 @@ impl Display for Swap {
...
+    }
+}
+
+impl From<AliceState> for Swap {
+    fn from(state: AliceState) -> Self {
+        Self::Alice(Alice::from(state))
+    }
+}
+
+pub struct Database<T> where T: Into<Swap> + TryFrom<Swap>  {
```
thomaseizinger: Try to avoid trait-bounds on structs but instead move them to the `impl` block there they are needed! If they are on the struct, you will have to repeat them everywhere where you use the struct which may ripple through your codebase.

swap/src/database.rs
```diff
@@ -210,34 +236,49 @@ impl Database {
...
             let swap_id = deserialize::<Uuid>(&key)?;
-            let swap = deserialize::<T>(&value).context("Failed to deserialize swap")?;
+            let swap = deserialize::<Swap>(&value).context("Failed to deserialize swap")?;
+
+            let t = T::try_from(swap);
+
+            let t = match t {
+                Ok(a) => a,
+                Err(_) => bail!("a")
+            };
```
thomaseizinger: That is equivalent to calling `.context("a")` :)

swap/src/asb/recovery/punish.rs
```diff
@@ -46,7 +47,7 @@ pub async fn punish(
     let txid = state3.punish_btc(&bitcoin_wallet).await?;
 
     let state = AliceState::BtcPunished;
-    db.insert_latest_state(swap_id, state.clone())
+    db.insert_latest_state(swap_id, state.clone().into())
```
thomaseizinger, line 50: This `.into` should be not be necessary?

swap/src/database/bob.rs
```diff
@@ -56,23 +56,23 @@ impl From<BobState> for Bob {
                 btc_amount,
                 change_address,
             },
-            BobState::SwapSetupCompleted(state2) => Bob::ExecutionSetupDone { state2 },
-            BobState::BtcLocked(state3) => Bob::BtcLocked { state3 },
+            BobState::SwapSetupCompleted(state2) => Bob::ExecutionSetupDone { state2: *state2 },
```
thomaseizinger, line 59: If the `Box`s here are annoying, you can just get rid of them and allow the respective clippy lint that comes up.

swap/src/protocol.rs
```diff
@@ -65,3 +74,80 @@ pub struct Message4 {
...
+    async fn get_peer_id(&self, swap_id: Uuid) -> Result<PeerId>;
+    async fn insert_monero_address(&self, swap_id: Uuid, address: monero::Address) -> Result<()>;
+    async fn get_monero_address(&self, swap_id: Uuid) -> Result<monero::Address>;
+    async fn insert_address(&self, peer_id: PeerId, address: Multiaddr) -> Result<()>;
+    async fn get_addresses(&self, peer_id: PeerId) -> Result<Vec<Multiaddr>>;
+    async fn insert_latest_state(&self, swap_id: Uuid, state: State) -> Result<()>;
+    async fn get_state(&self, swap_id: Uuid) -> Result<State>;
+    async fn all(&self) -> Result<Vec<(Uuid, State)>>;
+    async fn unfinished(&self, unfinished: fn(State) -> bool) -> Result<Vec<(Uuid, State)>>;
+}
```
thomaseizinger: Any reason why this type is now no longer generic over `T`?

swap/src/protocol/bob.rs
```diff
@@ -3,30 +3,31 @@ use std::sync::Arc;
...
 
 pub use self::state::*;
 pub use self::swap::{run, run_until};
+use std::convert::TryInto;
 
 pub mod state;
 pub mod swap;
 
-pub struct Swap {
+pub struct Swap<D: Database> {
```
thomaseizinger, line 16: Same here, unless you are accessing associated types of a type parameter, try to avoid putting bounds on them in the struct declaration.

swap/src/protocol/bob/state.rs
```diff
@@ -31,19 +31,19 @@ pub enum BobState {
...
-    BtcRedeemed(State5),
-    CancelTimelockExpired(State6),
-    BtcCancelled(State6),
-    BtcRefunded(State6),
+    XmrLocked(Box<State4>),
+    EncSigSent(Box<State4>),
+    BtcRedeemed(Box<State5>),
+    CancelTimelockExpired(Box<State6>),
+    BtcCancelled(Box<State6>),
+    BtcRefunded(Box<State6>),
```
thomaseizinger, line 46: Did clippy tell you to do this? :)

I wouldn't bother, this part is not super performance critical and we certainly should not optimize this without having measured an improvement.

swap/migrations/20210903050345_create_swaps_table.sql
```diff
@@ -0,0 +1,33 @@
...
+    peer_id     TEXT                NOT NULL,
+);
+
+CREATE TABLE if NOT EXISTS monero_addresses
+(
+    swap_id     TEXT    PRIMARY KEY NOT NULL,
+    address     TEXT                NOT NULL,
+);
+
+CREATE TABLE if NOT EXISTS addreses
```
thomaseizinger, line 29: Typo :)

```suggestion
CREATE TABLE if NOT EXISTS addresses
```

swap/src/protocol.rs
```diff
@@ -75,7 +75,7 @@ pub struct Message4 {
     tx_cancel_sig: bitcoin::Signature,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
thomaseizinger, line 78: Why do these suddenly have to be serializable?

swap/src/sqlite.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    async fn all(&self) -> Result<Vec<(Uuid, State)>> {
+        let mut conn = self.pool.acquire().await?;
+        all_latest_swaps_iter(&mut conn).await
+    }
+
+    async fn unfinished(&self, unfinished: fn(State) -> bool) -> Result<Vec<(Uuid, State)>> {
+        Ok(self.all()
+            .await?
+            // todo: why is into iter returning an &State
```
thomaseizinger, line 117: `into_iter` is not but `filter` only gives you a reference otherwise you would have to return the owned instance again in a positive case as you would be taking ownership.

swap/src/protocol.rs
```diff
@@ -139,7 +139,7 @@ impl TryInto<AliceState> for State {
 
 #[async_trait]
 pub trait Database {
-    async fn open(path: &Path) -> Result<Self> where Self: std::marker::Sized;
+    async fn open(path: PathBuf) -> Result<Self> where Self: std::marker::Sized;
```
thomaseizinger, line 142: Does `open` have to be part of the trait? We are doing at the very top of main where we know which DB we want to construct :)

thomaseizinger (commented):
Couple of comments!

Bit of back and forth in the commits but it looks like we are getting there :)

swap/src/protocol.rs
thomaseizinger: Ah I see, you are not using the same structs as you did for the sled DB. We already have them, I think we can just use them across both implementations.

Making these types serializable brings back the issue of easily making breaking changes across two versions because you cannot deserialize the old state from the DB.

swap/src/protocol.rs
rishflab: yep you are right, I think we can remove that

swap/src/protocol.rs
rishflab: The main reason I reverted the the type parameters was because I was running into lifetime/generics/async related compile errors. I am also not sure if it is a good idea. It gives the impression the DB can only store Alice or Bob data but in reality it can do either. I think after we migrate we could decide between having a `AliceDatabase` and `BobDatabase` trait vs having a type param?

swap/src/database.rs
rishflab: If i remember correctly, I tried this but there was a really weird error and I had to this workaround

swap/src/database.rs
rishflab: cool didnt know that

thomaseizinger:
@rishflab Ping me when you want a review on this!

swap/src/database.rs
```diff
@@ -68,3 +75,56 @@ impl From<AliceState> for Swap {
...
+        (false, _, true) => {
+            bail!("Sled database does not exist at specified location")
+        }
+        (true, _, true) => {
+            let sled= SledDatabase::open(data_dir.join("sled")).await?;
+            Ok(Arc::new(sled))
+        }
+    }
+
+}
```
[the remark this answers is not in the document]
rishflab: havent run dprint yet

rishflab:
> There is one issue in regards the migration that is a show-stopper. I suspect that should come up once you do testing with the latest release binary.

Is it something fixable or is the migration unfeasible?

> I have an idea of how we could tidy up the history a bit more. Would be more of an educational exercise rather than a necessity though, just let me know :)

I think I will start a new branch. Some of the commits are dependant on each other but in the wrong order.

rishflab:
> > Is it something fixable or is the migration unfeasible?
> 
> It is fixable!
> 
> I was referring to this:
> 
> [#780 (comment)](https://github.com/comit-network/xmr-btc-swap/pull/780#discussion_r717179934)

I think i found another issue.  

https://github.com/comit-network/xmr-btc-swap/blob/0e4891fcc0f15eded190caa84c46150fd3e9b765/swap/src/bin/asb.rs#L93

It seems like the database was being stored in a ridiculous location for the ASB

thomaseizinger (commented):
Thank you for the nice commits, it made it really easy to review :)

Just a couple of minor comments, nothing big. Let me know once you've successfully tested this!

swap/src/asb/event_loop.rs
```diff
@@ -108,7 +108,7 @@ where
         self.inflight_encrypted_signatures
             .push(future::pending().boxed());
 
-        let unfinished_swaps = match self.db.unfinished_alice() {
+        let unfinished_swaps = match self.db.unfinished(|state| !state.swap_finished()).await {
```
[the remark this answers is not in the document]
rishflab: removed it from the db trait and inlined the filtering code as it was only used once.

swap/src/bin/swap.rs
```diff
@@ -152,7 +152,11 @@ async fn main() -> Result<()> {
 
             println!("{}", table);
         }
-
+        Command::Config => {
+            println!("Data Directory: {}", data_dir.display());
+            let config_json = serde_json::to_string_pretty(&env_config)?;
+            println!("Config: {}", config_json);
```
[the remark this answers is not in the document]
rishflab: nope no data dir in the swap config

swap/src/database.rs
```diff
@@ -210,55 +201,64 @@ impl Database {
...
 
-        let state = deserialize(&encoded).context("Could not deserialize state")?;
+        let swap = deserialize::<Swap>(&encoded).context("Could not deserialize state")?;
+
+        let state = State::from(swap);
+
         Ok(state)
     }
 
-    pub fn all_alice(&self) -> Result<Vec<(Uuid, Alice)>> {
```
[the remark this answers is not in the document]
rishflab: I originally wanted to do this using a type parameter. I will try reintroducing that since everything is working now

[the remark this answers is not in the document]
thomaseizinger: Okay, I see now, don't worry about it then :)

swap/src/bin/swap.rs
```diff
@@ -152,7 +152,11 @@ async fn main() -> Result<()> {
 
             println!("{}", table);
         }
-
+        Command::Config => {
+            println!("Data Directory: {}", data_dir.display());
+            let config_json = serde_json::to_string_pretty(&env_config)?;
```
thomaseizinger, line 157: Hang on why are we printing `env_config`?

thomaseizinger: Is this command even any useful on the CLI side?

swap/src/database.rs
```diff
@@ -101,38 +99,39 @@ pub async fn open_db(
     ) {
         (true, false, false) => {
             let sled_db = SledDatabase::open(sled_path.as_ref()).await?;
-            let sqlite = SqliteDatabase::open(sqlite_path).await?;
 
-            let swaps = sled_db.all().await?;
+            tokio::fs::File::create(&sqlite_path).await?;
+            let mut sqlite = SqliteDatabase::open(sqlite_path).await?;
+            sqlite.run_migrations().await?;
```
thomaseizinger, line 106: Shouldn't we just always run this as part of `open`?

swap/src/database.rs
```diff
@@ -126,9 +127,12 @@ pub async fn open_db(
                 sqlite.insert_peer_id(swap_id, peer_id).await?;
             }
 
+            tracing::info!("Sucessfully migrated data to sqlite! Using sqlite.");
```
thomaseizinger, line 130: Nice! Should we add a line that prints the path to the sled DB so users know what to delete?

swap/src/database.rs
```diff
@@ -126,9 +127,12 @@ pub async fn open_db(
...
+
             Ok(Arc::new(sqlite))
         }
         (_, _, false) => {
+            tracing::debug!("Sled database not found. Using sqlite.");
```
thomaseizinger: I know it is explicit from the previous pattern but having `(false, _, false)` here would be more readable I think.

swap/src/database.rs
```diff
@@ -126,9 +127,12 @@ pub async fn open_db(
...
             tokio::fs::File::create(&sqlite_path).await?;
```
thomaseizinger, line 136: Wouldn't this fail if the sqlite DB already exists? I think we may need another pattern that checks for `sqlite` true and false and either doesn't or does create the file.

thomaseizinger (approved):
Very nice! Thank you!

I'd say we squash the last couple of fixes into the original migration commit then we merge this :)

swap/src/database.rs
rishflab: yep thats a good idea

swap/src/bin/swap.rs
rishflab: How will the users know where the database is?

We set the env config. I think this command should print it otherwise it is just a command that prints the data dir

thomaseizinger: But the env config doesn't contain the database path! The CLI doesn't have a config file, we need to compute the database path from the passed in data-dir and print it.

da-kami (commented):
thanks for the great work , I had a quick glance over, looks good. Not sure I have the time for a fullblown review, but I think @thomaseizinger covered it. 

were you able to test with the testnet asb db @rishflab ?

binarybaron:
One small issue I ran into is that when the user hasn't started any swaps yet, running `./swap config` throws an error.

On macOS:
```
./swap config
Error: No such file or directory (os error 2)
```

rishflab:
> One minor issue I've run into is that running `./swap config` throws an error if the user has never started a swap before.
> 
> On macOS:
> 
> ```
> ./swap config
> Error: No such file or directory (os error 2)
> ```

yep, fix has been pushed, thanks for reporting

thomaseizinger:
> Build failed:
> 
>     * [docker_tests (concurrent_bobs_after_xmr_lock_proof_sent)](https://github.com/comit-network/xmr-btc-swap/runs/3808249586)

I don't understand this. Are the tests using the new sqlite database or the sled database?

rishflab:
> > Build failed:
> > ```
> > * [docker_tests (concurrent_bobs_after_xmr_lock_proof_sent)](https://github.com/comit-network/xmr-btc-swap/runs/3808249586)
> > ```
> 
> I don't understand this. Are the tests using the new sqlite database or the sled database?

the sled database. This is the same issue from last week, where the tests works sometimes and sometimes it doesn't. The test fails on the first commit which only reorders modules.

thomaseizinger:
> > > Build failed:
> > > ```
> > > * [docker_tests (concurrent_bobs_after_xmr_lock_proof_sent)](https://github.com/comit-network/xmr-btc-swap/runs/3808249586)
> > > ```
> > 
> > 
> > I don't understand this. Are the tests using the new sqlite database or the sled database?
> 
> the sled database. This is the same issue from last week, where the tests works sometimes and sometimes it doesn't. The test fails on the first commit which only reorders modules.

Can you just move this one to the sqlite database (if that is easier) or maybe all of them at once. We need to do that anyway if we want to delete the sled DB eventually.

Issue #796, opened 2021-09-29
Lower default bitcoin_finality_confirmations


binarybaron:
I think we should consider lowering the default confirmations required for Alice to lock her XMR from 2 to 1. This will significantly reduce the total time it takes to complete a swap (10 minutes) and will not significantly decrease security.
Is there a requirement for 2 confirmations at the protocol level that I'm not aware of?

devbordecraft:
Don't see any drawback, I made this pull (#800) request to reduce the number of block for bitcoin confirmation.

rishflab:
I will check with the team, I think the 2 block requirement was introduced as a workaround for a bug.

rishflab:
I checked with the team. The 2 block requirement on the `cli` only affects the refund tx. 

On the ASB, the bitcoin finality confirmations can be set to 1 using the config file. This will get you the swap reduction time you desire.

binarybaron:
Done

PR #800, opened 2021-10-04
Change the bitcoin finality confirmation requirement from 2 to 1 block

devbordecraft:


devbordecraft:
For reference : #796

binarybaron:
If we do this, we should also reduce it to 1 for the `testnet`.

https://github.com/comit-network/xmr-btc-swap/blob/1b959243c0dc0b200fd265c5410d217fdfede699/swap/src/env.rs#L66

rishflab:
The 2 block requirement on the cli only affects the refund tx.

On the ASB, the bitcoin finality confirmations can be set to 1 using the config file. This will get you the swap reduction time you desire.

PR #804, opened 2021-10-06
Minor style improvements in CLI

proudmuslim-dev:
Noticed these while reading through, figured I'd open a PR

### DIFF
--- a/swap/src/cli/command.rs
+++ b/swap/src/cli/command.rs
@@ -64,10 +64,7 @@ where
         Err(e) => anyhow::bail!(e),
     };
 
-    let debug = args.debug;
-    let json = args.json;
-    let is_testnet = args.testnet;
-    let data = args.data;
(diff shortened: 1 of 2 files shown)


swap/src/cli/command.rs
```diff
@@ -64,10 +64,7 @@ where
         Err(e) => anyhow::bail!(e),
     };
 
-    let debug = args.debug;
-    let json = args.json;
-    let is_testnet = args.testnet;
-    let data = args.data;
+    let (debug, json, is_testnet, data) = (args.debug, args.json, args.testnet, args.data);
```
rishflab, line 67: Not convinced this is really a style improvement. I think it is actually harder to read.

proudmuslim-dev: You're right, but my main reason for doing that was I felt that it was just a waste to assign each of them individually

PR #806, opened 2021-10-07
Upgrade bitcoin and bdk to latest version

thomaseizinger:
The latest version allows us to access `miniscript` via `bdk` which
removes the need for declaring it as an extra dependency.


rishflab (approved):
Thanks for this. A lot smaller that i expected

PR #820, opened 2021-10-12
Use upstream libp2p

rishflab:
todo: remove ping

### DIFF
--- a/swap/src/cli.rs
+++ b/swap/src/cli.rs
@@ -59,10 +59,7 @@ mod tests {
     }
 
     async fn setup_rendezvous_point() -> (Multiaddr, PeerId) {
-        let mut rendezvous_node = new_swarm(|_, _| RendezvousPointBehaviour {
-            rendezvous: rendezvous::server::Behaviour::new(rendezvous::server::Config::default()),
-            ping: Default::default(),
-        });
+        let mut rendezvous_node = new_swarm(|_, _| RendezvousPointBehaviour::default());
         let rendezvous_address = rendezvous_node.listen_on_tcp_localhost().await;
         let rendezvous_peer_id = *rendezvous_node.local_peer_id();
 
@@ -174,5 +171,16 @@ mod tests {
     }
     impl NetworkBehaviourEventProcess<libp2p::ping::PingEvent> for RendezvousPointBehaviour {
         fn inject_event(&mut self, _: libp2p::ping::PingEvent) {}
+    }
+
+    impl Default for RendezvousPointBehaviour {
+        fn default() -> Self {
+            RendezvousPointBehaviour {
+                rendezvous: rendezvous::server::Behaviour::new(
+                    rendezvous::server::Config::default(),
+                ),
+                ping: Default::default(),
+            }
+        }
     }
 }


rishflab:
the rendezvous `unit` test failed locally. There were no failures in CI and im pretty sure it would of run. Investigating... may just be a one off failure

thomaseizinger:
It is not passing CI either but hanging forever, I will have a look!

swap/src/network/cbor_request_response.rs
```diff
@@ -59,7 +55,7 @@ where
     where
         T: AsyncRead + Unpin + Send,
     {
-        let message = upgrade::read_one(io, BUF_SIZE)
+        let message = upgrade::read_length_prefixed(io, BUF_SIZE)
             .await
             .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
```
thomaseizinger, line 60: Also get rid of this.

swap/src/network/json_pull_codec.rs
```diff
@@ -55,7 +55,7 @@ where
     where
         T: AsyncRead + Unpin + Send,
     {
-        let message = upgrade::read_one(io, BUF_SIZE)
+        let message = upgrade::read_length_prefixed(io, BUF_SIZE)
             .await
             .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
```
thomaseizinger, line 60: And here.

swap/src/cli.rs
```diff
@@ -138,14 +138,14 @@ mod tests {
         #[behaviour(ignore)]
         registered: bool,
     }
-
-    impl NetworkBehaviourEventProcess<libp2p::rendezvous::Event> for StaticQuoteAsbBehaviour {
-        fn inject_event(&mut self, event: libp2p::rendezvous::Event) {
-            if let libp2p::rendezvous::Event::Registered { .. } = event {
+    impl NetworkBehaviourEventProcess<libp2p::rendezvous::client::Event> for StaticQuoteAsbBehaviour {
```
thomaseizinger, line 141: Doing `use libp2p::rendezvous;` would go  along way in this file as well.

swap/src/network/cbor_request_response.rs
thomaseizinger: Can we not use `?` here?

thomaseizinger (changes requested):
Thanks! Couple of comments :)

swap/src/cli.rs
```diff
@@ -59,10 +59,9 @@ mod tests {
...
-        let mut rendezvous_node = new_swarm(|_, identity| RendezvousPointBehaviour {
-            rendezvous: libp2p::rendezvous::Rendezvous::new(
-                identity,
-                libp2p::rendezvous::Config::default(),
+        let mut rendezvous_node = new_swarm(|_, _| RendezvousPointBehaviour {
+            rendezvous: libp2p::rendezvous::server::Behaviour::new(
+                libp2p::rendezvous::server::Config::default(),
             ),
             ping: Default::default(),
         });
```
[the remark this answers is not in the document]
rishflab: its only constructed once and in a test. not sure if worth it

PR #821, opened 2021-10-12
Remove the support for the sled database

rishflab:


thomaseizinger (approved):
Thanks!

Can you attempt to transition to sqlite for the wallet as well? Would be nice to get rid of the sled dependency in our dep-tree altogether!

thomaseizinger (approved):

PR #843, opened 2021-11-05
Add section that discourages bumping fees

da-kami:
Someone ran into trouble cancelling after bumping fees on lock.
I did not think it through completely but would more clearly discourage from doing so in the readme.


thomaseizinger (approved):

da-kami:
bors ping

Issue #797, opened 2021-09-30
CloseNotify alert received


devbordecraft:
**Describe the bug**
Sometimes the swap can be stopped by a "CloseNotify alert". I saw this error on both the swap-cli and the asb.

**Lost/trapped Funds**
No

**Debug logs**
 WARN Failed to get status of script: Failed to subscribe to header notifications: Made one or multiple attempts, all errored:
        - CloseNotify alert received
 txid=[REDACTED]

**Platform (please complete the following information):**

- CLI and ASB V0.8.3 Ubuntu

devbordecraft:
![image](https://user-images.githubusercontent.com/36452299/136685263-6bbfd0c9-682b-4577-8d7e-95f5eb41b6f6.png)
This is on swap-cli. Terminal stopped after that error message.

And the same thing on the ASB side. It stop following the bitcoin transaction and I have to restart the ASB to resume the pending swap.
![image](https://user-images.githubusercontent.com/36452299/136685331-e6bee043-7e1d-4f39-bada-23d56aa5ce96.png)

And yes I use the default electrum node.

icy-ux:
Switching to a locally hosted `electrs` instance, instead of the default electrum node, seemed to help.

PR #866, opened 2021-12-11
Fix electrum query failing (CloseNotify bug)

devbordecraft:
Retry query the electrum server if it fail. This address #797

### DIFF
--- a/swap/src/asb/config.rs
+++ b/swap/src/asb/config.rs
@@ -58,7 +58,7 @@ impl GetDefaults for Mainnet {
             data_dir: default_asb_data_dir()?.join("mainnet"),
             listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
-            electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:50002")?,
+            electrum_rpc_url: Url::parse("ssl://blockstream.info:700")?,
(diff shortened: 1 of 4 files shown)


binarybaron:
Great, thanks for taking the initiative to fix this. I think it might be useful to have a list of Electrum nodes and fall back to another server if the primary one fails.

swap/src/asb/config.rs
```diff
@@ -58,7 +58,7 @@ impl GetDefaults for Mainnet {
             data_dir: default_asb_data_dir()?.join("mainnet"),
             listen_address_tcp: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9939")?,
             listen_address_ws: Multiaddr::from_str("/ip4/0.0.0.0/tcp/9940/ws")?,
-            electrum_rpc_url: Url::parse("ssl://electrum.blockstream.info:50002")?,
+            electrum_rpc_url: Url::parse("ssl://blockstream.info:700")?,
```
rishflab, line 61: Does the old url no longer work?

rishflab (approved):

swap/src/asb/config.rs
devbordecraft: It work but it is very unstable. That's why I propose to change to this electrum node referenced in the official blog of Blockstream : https://blog.blockstream.com/en-esplora-and-other-alternatives-to-electrumx/

devbordecraft:
> Great, thanks for taking the initiative to fix this. I think it might be useful to have a list of Electrum nodes and fall back to another server if the primary one fails.

No problem ❤️, good idea, will make an issue for that

da-kami (approved):

Issue #537, opened 2021-05-25
CLI cannot spend from generated XMR wallet after resume


da-kami:
## What happened

* Start swap between CLI on laptop and ASB on raspberry, branch `no-transfer-proof-failure`, commit `facab99b8af6b2bebd15096c2f324c8460e9089d`
* Swap got stale waiting for transfer proof on the CLI (ASB errored with `ERROR Communication error. Error /comit/xmr/btc/transfer_proof/1.0.0 failed because we did not receive a response within the configured timeout peer=12D3KooWRqVnNWuahKawSSmxQQptGcEApNBFXNodzRv2q6tBvhUW` - monero tx `09e361acb3e6e71d627a945a30672776a6f8fec7c97f4cae5e09b0780b75c158`
* Restarted the CLI, the CLI kept waiting (this is expected, because the ASB does not go on with swap execution once it fails with an unrecoverable error)
* Let the CLI run and wait
* Restart the ASB
* The ASB re-sends the transfer proof
* The CLI received the transfer proof
* The Monero tx already had more than 10 confirmations
* The CLI sent the encsig
* The ASB received the encsig and published Bitcoin redeem
* The CLI extraced key from Bitcoin redeem tx and generated Monero wallet
* When draining the Monero wallet to spend the funds the CLI reports that there is no unlocked balance in the Monero wallet

## Expected

CLI should be able to detect and spend the Monero locked by the ASB.

## Logs

Logs are cluttered with some libp2p events that are logged on that branch for investigation reasons, sorry. 
It might be very hard to reproduce this problem so I thought I better record it.

CLI logs upon resume:

```
/Users/dakami/.cargo/bin/cargo run --color=always --bin swap --manifest-path /Users/dakami/CoBloX/repos/xmr-btc-swap_comit-network/swap/Cargo.toml -- --testnet --debug --data-dir /Users/dakami/TEMP/swap/bob/01 resume --swap-id 41f849f8-f4f0-4af8-990e-e0910c939a57 --receive-address 58hKkN5JrirdNehmTXaHhTEg3N5zRYZ6Wb5g5jwDk3wRC4rtNCJvx7hENsbLmfPakC3spGhciosagdVbSqq9vfXsV3zusCn --seller-addr /ip4/141.168.172.35/tcp/8765

    Finished dev [unoptimized + debuginfo] target(s) in 1.09s
[...]

PR #869, opened 2021-12-13
Print cli logs as json if flag is set

rishflab:

PR #880, opened 2021-12-20
Add `monero-recovery` command to print address + view/spend keys

da-kami:
This allows manual recovery of problems like https://github.com/comit-network/xmr-btc-swap/issues/537
Since we could not figure out what causes this issue, and it is most likely an upstream problem this is the best we can do so far to allow the user to manually interact with `monero-wallet-cli` or `monero-wallet-rpc`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `monero-recovery` command that can be used to print the monero address, private spend and view key so one can manually recover instances where the `monero-wallet-rpc` does not pick up the Monero funds locked up by the ASB.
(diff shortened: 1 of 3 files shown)


da-kami:
Just saw https://github.com/comit-network/xmr-btc-swap/pull/860 - shall we always output on tracing? I don't remember why there are still `println` in the binary code - I don't see a reason not to use tracing.

rishflab (approved):

rishflab:
> Just saw #860 - shall we always output on tracing? I don't remember why there are still `println` in the binary code - I don't see a reason not to use tracing besides the fact that all the output will be json then if run with the json flag.

the json behaviour seems to be desirable. merging this.

da-kami:
> > Just saw #860 - shall we always output on tracing? I don't remember why there are still `println` in the binary code - I don't see a reason not to use tracing besides the fact that all the output will be json then if run with the json flag.
> 
> the json behaviour seems to be desirable. merging this.

then we should change what was merged to also use `trace` and update #869

PR #881, opened 2021-12-20
Record the monero-wallet-restore blockcheight before locking BTC

da-kami:
This solves issues where the CLI went offline after sending the BTC transaction, and the monero wallet restore blockheight being recorded after Alice locked the Monero, resulting in the generated XMR redeem wallet not detecting the transaction and reporting `No unlocked balance in the specified account`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The swap has to be in a `BtcRedeemed` state.
   Use `--help` for more details.
 
+### Changed
+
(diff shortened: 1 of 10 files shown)


binarybaron:
Do you know if this change is backward compatible?

da-kami:
> 

This change does not affect the network layer, so old versions of the CLI will still work with an upgraded ASB.

This change is not backwards compatible in the sense that you can resume old swaps because it touches the database and I cannot migrate old swaps (because we would not know to what height...). So, new swaps will not have the problem anymore, but you won't be able to resume old ones. For the old ones you might have to go for manual recovery using #880

thomaseizinger (approved):

da-kami:
bors retry

Issue #849, opened 2021-11-28
Save logs for swap started with `--json` flag and save all logs in JSON format


binarybaron:
**Describe the bug**
The logs are not saved when the swap is started with the `--json` flag. This makes it much more difficult to reconstruct the previous swap state when resuming a swap, which is necessary for the GUI I am working on. In addition, it would be very helpful if all saved logs were saved in JSON format, as this would allow for the GUI to resume and display proper information about swaps that were not initiated using the GUI.

**Lost/trapped Funds**
No

**Platform (please complete the following information):**

- CLI
- Software Version All
- Operating System: All

rishflab:
Hmm could you pipe the json output to a file? What about directly accessing the recently introduced sqlite database?

binarybaron:
> Hi sorry for the delayed resposen.

No worries, I know you guys are also working on other projects.

> Could you pipe the json output to a file?

This is basically what we are doing at the moment, but it seems like something that should be handled by the cli itself. It would also help to be fully compatible with swaps initiated with the cli and not with our GUI, since we can't store logs manually for those and it would give us some important information about those swaps that can't be extracted from the database (or at least not easily)

https://github.com/UnstoppableSwap/unstoppableswap-gui/blob/d80b8082a1fb342e2b04cb26e5db22f634270a1e/src/main/cli/commands/buyXmrCommand.ts#L49-L60

> What about directly accessing the recently introduced sqlite database?

The GUI does use the sqlite database to retrieve information about the states of swaps but primarily for displaying information about swaps that have occurred in the past (swap history page). For running swaps we rely on the logs because some data is not easily extractable from the database (e.g transaction ids)

PR #893, opened 2021-12-23
Always write logs as JSON to file

binarybaron:
Related issue: https://github.com/comit-network/xmr-btc-swap/issues/849

### DIFF
--- a/swap/src/cli/tracing.rs
+++ b/swap/src/cli/tracing.rs
@@ -20,27 +20,22 @@ pub fn init(debug: bool, json: bool, dir: impl AsRef<Path>, swap_id: Option<Uuid
 
         std::mem::forget(guard);
 
-        let file_logger = fmt::layer()
-            .with_ansi(false)
-            .with_target(false)
-            .with_writer(appender);
+        let file_logger = registry.with(
+            fmt::layer()
+                .with_ansi(false)
+                .with_target(false)
+                .json()
+                .with_writer(appender),
+        );
 
         if json && debug {
-            set_global_default(
-                registry
-                    .with(file_logger.json())
-                    .with(debug_json_terminal_printer()),
-            )?;
+            set_global_default(file_logger.with(debug_json_terminal_printer()))?;
         } else if json && !debug {
-            set_global_default(
-                registry
-                    .with(file_logger.json())
-                    .with(info_json_terminal_printer()),
-            )?;
+            set_global_default(file_logger.with(info_json_terminal_printer()))?;
         } else if !json && debug {
-            set_global_default(registry.with(file_logger).with(debug_terminal_printer()))?;
+            set_global_default(file_logger.with(debug_terminal_printer()))?;
         } else {
-            set_global_default(registry.with(file_logger).with(info_terminal_printer()))?;
+            set_global_default(file_logger.with(info_terminal_printer()))?;
         }
 
         Ok(())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Revert logs to use rfc3339 local time formatting.
+- Always write logs as JSON to files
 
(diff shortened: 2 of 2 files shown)


da-kami (approved):
Couldn't that cause confusion for users looking at the logs? 
I know it helps with the UI. I don't mind merging this. Any thoughts @rishflab @thomaseizinger ?

thomaseizinger (changes requested):
What is the intent of this PR? Can you provide more context on whether or not this is a behavioural change or a refactoring?

binarybaron:
This is a behavior change. It changes the way the logs are written to disk. Right now, the format that the logs are written to depends on whether or not the swap was started with the --json flag. To provide more consistency and allow other projects built on the cli to properly parse and access information about swaps that were started manually, it would be handy if these logs that are written to a log file were always in JSON format. That's what this pull requests intends change. Should I add a changelog entry for this?

swap/src/cli/tracing.rs
```diff
@@ -23,18 +23,19 @@ pub fn init(debug: bool, json: bool, dir: impl AsRef<Path>, swap_id: Option<Uuid
         let file_logger = fmt::layer()
             .with_ansi(false)
             .with_target(false)
+            .json()
             .with_writer(appender);
 
         if json && debug {
```
thomaseizinger, line 29: Testing for the `json` flag is now pointless here, isn't it?

swap/src/cli/tracing.rs
```diff
@@ -23,18 +23,19 @@ pub fn init(debug: bool, json: bool, dir: impl AsRef<Path>, swap_id: Option<Uuid
...
             set_global_default(
                 registry
-                    .with(file_logger.json())
+                    .with(file_logger)
                     .with(debug_json_terminal_printer()),
             )?;
         } else if json && !debug {
```
thomaseizinger, line 35: Same here.

swap/src/cli/tracing.rs
```diff
@@ -23,18 +23,19 @@ pub fn init(debug: bool, json: bool, dir: impl AsRef<Path>, swap_id: Option<Uuid
...
             set_global_default(
                 registry
-                    .with(file_logger.json())
+                    .with(file_logger)
                     .with(info_json_terminal_printer()),
             )?;
         } else if !json && debug {
```
thomaseizinger, line 41: And here!

thomaseizinger (changes requested):
> This is a behavioural change. It changes the way the logs are written to disk. Right now, the format that the logs are written to depends on whether or not the swap was started with the --json flag. To provide more consistency and allow other projects built on the cli to properly parse and access information about swaps that were started manually, it would be handy if these logs that are written to a log file were always in JSON format. That's what this pull requests intends change. Should I add a changelog entry for this?

Thanks, that clarifies things!

This needs a changelog entry yes.
I think there is also quite some unnecessary testing for the `json` flag now. See the inline comments. Please clean that up :)

Overall, I am okay with this change. The log file was only ever intended for post-run debugging and not something that the normal user would look at.

concept ACK but needs changelog entry and cleanup

binarybaron:
Thanks for the feedback! I will make these changes accordingly later today.

swap/src/cli/tracing.rs
binarybaron: I don't think so. We just want the file appender to always use JSON. The default terminal printer should still adhere to whether the JSON flag was passed or not. I might be misunderstanding something here.

thomaseizinger: We are always passing the same `file_logger` though so I think that can be extracted?

binarybaron: Like this? Can this be simplified further?

thomaseizinger (approved):
Yep that makes more sense now, thanks!

bors r+

thomaseizinger:
bors retry

binarybaron:
bors retry

thomaseizinger:
bors d=binarybaron

binarybaron:
Thanks @thomaseizinger 👍

Issue #901, opened 2021-12-27
Move from ChronoLocal to LocalTime completely breaks terminal output in some cases


binarybaron:
**Describe the bug**
This commit https://github.com/comit-network/xmr-btc-swap/commit/9e85637419594f69d05e0d03767d66b07001a0ae completely breaks the terminal output for me on (tested on macOS 11.0.1 and Ubuntu 18.04) when the `--debug` flag is used. The terminal stays completely blank aside from the QRCode that is printed. I've tried reverting that commit and recompiled the binary and it fixed the issue.

MacOS (compiled from 3ce4dffb84d0c2e01b011864cd36dabc64faf25c):
```
> sw_vers
ProductName:	macOS
ProductVersion:	11.0.1
BuildVersion:	20B50
> ./swap --debug --testnet buy-xmr --change-address tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx --receive-address 59McWTPGc745SRWrSMoh8oTjoXoQq6sPUgKZ66dQWXuKFQ2q19h9gvhJNZcFTizcnT12r63NFgHiGd6gBCjabzmzHAMoyD6 --seller /dns4/xmr-btc-asb.coblox.tech/tcp/9939/p2p/12D3KooWCdMKjesXMJz1SiZ7HgotrxuqhQJbP5sgBm2BwP1cqThi
█████████████████████████████████████
█████████████████████████████████████
████ ▄▄▄▄▄ █ ███▄▀██▄█▀▄▀█ ▄▄▄▄▄ ████
████ █   █ █▀█▀█▀ ▀▄ ▀▄ ██ █   █ ████
████ █▄▄▄█ ██▀█▀▄▄ ▄  █ ██ █▄▄▄█ ████
████▄▄▄▄▄▄▄█ █ ▀ █ █ ▀ ▀▄█▄▄▄▄▄▄▄████
████ █ ▄█ ▄ ▀████▀▀▄  ▄ ▀▀ ▀▀▄█▄ ████
████▄█▄█ █▄▄▀▄▄▀▀▀ █ ███▀▄█▀▄▄▄ ▀████
████   ▄ ▄▄█ ▀ █ ▄▄▀▄ █▀ █▀██▄  ▀████
████ ▀▀█▀▀▄▄▀█▀ █▄ ▀▄▄█ ▀ ▄▄█ ▄▀ ████
[...]

PR #902, opened 2021-12-29
Revert move from `chrono` to `local-time`

binarybaron:
This reverts commit 9e85637419594f69d05e0d03767d66b07001a0ae

Closes https://github.com/comit-network/xmr-btc-swap/issues/901


rishflab:
I am guessing this happened because we updated the tracing dependencies

Happy to merge this now but I think we will have to upgrade these dependencies at some point.

rishflab (approved):

PR #953, opened 2022-02-24
Adjust quote based on Bitcoin balance

### LINKED ISSUE #939: Adjust quote based on Bitcoin balance
The maximum swap amount should be calculated based on the Balance of the internal bitcoin wallet. A swap failing after depositing Bitcoin leads to poor UX.

lescuer97:
Fixes #939 #963 
Please comment in the new method for getting a Monero value, I had to allow clippy::cast_precision_loss to convert to a f64, for our use case I don't really thing we will lose much precision. 

Please comment on the implementation of the check.

### DIFF
--- a/swap/src/asb/event_loop.rs
+++ b/swap/src/asb/event_loop.rs
@@ -319,13 +319,46 @@ where
         min_buy: bitcoin::Amount,
         max_buy: bitcoin::Amount,
     ) -> Result<BidQuote> {
-        let rate = self
+        let ask_price = self
             .latest_rate
             .latest_rate()
-            .context("Failed to get latest rate")?;
+            .context("Failed to get latest rate")?
+            .ask()
+            .context("Failed to compute asking price")?;
+
+        let max_bitcoin_for_monero = self
+            .monero_wallet
+            .get_balance()
+            .await?
+            .max_bitcoin_for_price(ask_price);
+
+        if min_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                        "Your Monero balance is too low to initiate a swap, as your minimum swap amount is {}. You could at most swap {}",
+                        min_buy, max_bitcoin_for_monero
+                    );
+
+            return Ok(BidQuote {
+                price: ask_price,
+                min_quantity: bitcoin::Amount::ZERO,
+                max_quantity: bitcoin::Amount::ZERO,
+            });
+        }
+
+        if max_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                    "Your Monero balance is too low to initiate a swap with the maximum swap amount {} that you have specified in your config. You can at most swap {}",
+                    max_buy, max_bitcoin_for_monero
+                );
+            return Ok(BidQuote {
+                price: ask_price,
+                min_quantity: min_buy,
+                max_quantity: max_bitcoin_for_monero,
+            });
+        }
 
         Ok(BidQuote {
-            price: rate.ask().context("Failed to compute asking price")?,
+            price: ask_price,
             min_quantity: min_buy,
             max_quantity: max_buy,
         })
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -29,7 +29,7 @@ use swap::cli::{list_sellers, EventLoop, SellerStatus};
 use swap::database::open_db;
 use swap::env::Config;
 use swap::libp2p_ext::MultiAddrExt;
-use swap::network::quote::BidQuote;
+use swap::network::quote::{BidQuote, ZeroQuoteReceived};
 use swap::network::swarm;
 use swap::protocol::bob;
 use swap::protocol::bob::{BobState, Swap};
@@ -99,7 +99,7 @@ async fn main() -> Result<()> {
             let event_loop = tokio::spawn(event_loop.run());
 
             let max_givable = || bitcoin_wallet.max_giveable(TxLock::script_size());
-            let (amount, fees) = determine_btc_to_swap(
+            let (amount, fees) = match determine_btc_to_swap(
                 json,
                 event_loop_handle.request_quote(),
                 bitcoin_wallet.new_address(),
@@ -107,7 +107,16 @@ async fn main() -> Result<()> {
                 max_givable,
                 || bitcoin_wallet.sync(),
             )
-            .await?;
+            .await
+            {
+                Ok(val) => val,
+                Err(error) => match error.downcast::<ZeroQuoteReceived>() {
+                    Ok(_) => {
+                        bail!("Seller's XMR balance is currently too low to initiate a swap, please try again later")
+                    }
+                    Err(other) => bail!(other),
+                },
+            };
 
             tracing::info!(%amount, %fees,  "Determined swap amount");
 
@@ -556,6 +565,11 @@ where
 {
     tracing::debug!("Requesting quote");
     let bid_quote = bid_quote.await?;
+
+    if bid_quote.max_quantity == bitcoin::Amount::ZERO {
+        bail!(ZeroQuoteReceived)
+    }
+
     tracing::info!(
         price = %bid_quote.price,
         minimum_amount = %bid_quote.min_quantity,
@@ -913,6 +927,32 @@ mod tests {
  INFO swap: Received Bitcoin new_balance=0.21000000 BTC max_giveable=0.20000000 BTC
 "
         );
+    }
+
+    #[tokio::test]
+    async fn given_bid_quote_max_amount_0_return_errorq() {
+        let givable = Arc::new(Mutex::new(MaxGiveable::new(vec![
+            Amount::from_btc(0.0001).unwrap(),
+            Amount::from_btc(0.01).unwrap(),
+        ])));
+
+        let determination_error = determine_btc_to_swap(
+            true,
+            async { Ok(quote_with_max(0.00)) },
+            get_dummy_address(),
+            || async { Ok(Amount::from_btc(0.0101)?) },
+            || async {
+                let mut result = givable.lock().unwrap();
+                result.give()
+            },
+            || async { Ok(()) },
+        )
+        .await
+        .err()
+        .unwrap()
+        .to_string();
+
+        assert_eq!("Received quote of 0", determination_error);
     }
 
     struct MaxGiveable {
--- a/swap/src/monero.rs
+++ b/swap/src/monero.rs
@@ -99,6 +99,20 @@ impl Amount {
         self.0
     }
 
+    pub fn max_bitcoin_for_price(&self, ask_price: bitcoin::Amount) -> bitcoin::Amount {
+        let piconero_minus_fee = self.as_piconero().saturating_sub(MONERO_FEE.as_piconero());
+
+        if piconero_minus_fee == 0 {
+            return bitcoin::Amount::ZERO;
+        }
+
+        // There needs to be an offset for difference in zeroes beetween Piconeros and
+        // Satoshis
+        let piconero_calc = (piconero_minus_fee * ask_price.as_sat()) / PICONERO_OFFSET;
+
+        bitcoin::Amount::from_sat(piconero_calc)
+    }
+
     pub fn from_monero(amount: f64) -> Result<Self> {
         let decimal = Decimal::try_from(amount)?;
         Self::from_decimal(decimal)
@@ -358,6 +372,30 @@ mod tests {
             error.downcast_ref::<OverflowError>().unwrap(),
             &OverflowError(overflow_pics.to_owned())
         );
+    }
+
+    #[test]
+    fn geting_max_bitcoin_to_trade() {
+        let amount = Amount::parse_monero("10").unwrap();
+        let bitcoin_price_sats = bitcoin::Amount::from_sat(382_900);
+
+        let monero_max_from_bitcoin = amount.max_bitcoin_for_price(bitcoin_price_sats);
+
+        assert_eq!(
+            bitcoin::Amount::from_sat(3_828_988),
+            monero_max_from_bitcoin
+        );
+    }
+
+    #[test]
+    fn geting_max_bitcoin_to_trade_with_balance_smaller_than_locking_fee() {
+        let monero = "0.00001";
+        let amount = Amount::parse_monero(monero).unwrap();
+        let bitcoin_price_sats = bitcoin::Amount::from_sat(382_900);
+
+        let monero_max_from_bitcoin = amount.max_bitcoin_for_price(bitcoin_price_sats);
+
+        assert_eq!(bitcoin::Amount::ZERO, monero_max_from_bitcoin);
     }
 
     use rand::rngs::OsRng;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Revert logs to use rfc3339 local time formatting.
 - Always write logs as JSON to files
 
+### Added
+
+- Adjust quote based on Bitcoin balance.
+  If the max_buy_btc in the ASB config is higher than the available balance to trade it will return the max available balance discounting the locking fees for monero, in the case the balance is lower than the min_buy_btc config it will return 0 to the CLI. If the ASB returns a quote of 0 the CLI will not allow you continue with a trade.
+
 ## [0.10.2] - 2021-12-25
 
 ### Changed
@@ -305,7 +310,7 @@ It is possible to migrate critical data from the old db to the sqlite but there 
 - Fixed an issue where Alice would not verify if Bob's Bitcoin lock transaction is semantically correct, i.e. pays the agreed upon amount to an output owned by both of them.
   Fixing this required a **breaking change** on the network layer and hence old versions are not compatible with this version.
 
-[Unreleased]: https://github.com/comit-network/xmr-btc-swap/compare/0.10.2...HEAD
+[unreleased]: https://github.com/comit-network/xmr-btc-swap/compare/0.10.2...HEAD
 [0.10.2]: https://github.com/comit-network/xmr-btc-swap/compare/0.10.1...0.10.2
 [0.10.1]: https://github.com/comit-network/xmr-btc-swap/compare/0.10.0...0.10.1
 [0.10.0]: https://github.com/comit-network/xmr-btc-swap/compare/0.9.0...0.10.0
--- a/swap/src/network/quote.rs
+++ b/swap/src/network/quote.rs
@@ -37,6 +37,10 @@ pub struct BidQuote {
     pub max_quantity: bitcoin::Amount,
 }
 
+#[derive(Clone, Copy, Debug, thiserror::Error)]
+#[error("Received quote of 0")]
+pub struct ZeroQuoteReceived;
+
 /// Constructs a new instance of the `quote` behaviour to be used by the ASB.
 ///
 /// The ASB is always listening and only supports inbound connections, i.e.


binarybaron:
Thanks a lot for implementing this! I'll look over it and give some feedback where I can

swap/src/monero.rs
```diff
@@ -98,6 +98,10 @@ impl Amount {
     pub fn as_piconero(&self) -> u64 {
         self.0
     }
+    #[allow(clippy::cast_precision_loss)]
```
thomaseizinger, line 101: Instead of allowing the clippy lint, please use `Decimal` for the calculations.

thomaseizinger: A test would be good for this function!

swap/src/asb/event_loop.rs
```diff
@@ -318,16 +318,30 @@ where
...
             .context("Failed to get latest rate")?;
 
+        max_bitcoin_for_monero = bitcoin::Amount::from_btc(monero_balance * rate.ask()?.as_btc())?;
+        if max_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                    "Your Max bitcoin to sell: {} is smaller than your current balance of monero available to sell {}",
+                    max_buy, max_bitcoin_for_monero
+                );
+            max_bitcoin_buy = max_bitcoin_for_monero
+        }
```
thomaseizinger, line 339: Please implement this without mutable variables, this is too hard to follow in what is being modified under which conditions.

swap/src/asb/event_loop.rs
```diff
@@ -318,16 +318,30 @@ where
...
+
         Ok(BidQuote {
             price: rate.ask().context("Failed to compute asking price")?,
```
thomaseizinger, line 342: Getting the `ask` price is duplicated here with line 332.

swap/src/asb/event_loop.rs
```diff
@@ -318,16 +318,30 @@ where
...
+        let max_bitcoin_for_monero: bitcoin::Amount;
+        let mut max_bitcoin_buy = max_buy;
+        let monero_balance = monero_wallet.get_balance().await?.as_monero();
+
         let rate = self
             .latest_rate
             .latest_rate()
             .context("Failed to get latest rate")?;
 
+        max_bitcoin_for_monero = bitcoin::Amount::from_btc(monero_balance * rate.ask()?.as_btc())?;
```
thomaseizinger, line 332: No floating point math please. Either operate on sats and piconeros (which are both u64 I think) or use `Decimal` if you can't use natural numbers.

swap/src/asb/event_loop.rs
```diff
@@ -318,16 +318,30 @@ where
         &mut self,
         min_buy: bitcoin::Amount,
         max_buy: bitcoin::Amount,
+        monero_wallet: Arc<monero::Wallet>,
```
thomaseizinger, line 321: No need to pass this as a parameter, you have it on `self`. (Also visible by the fact that you pass `self.monero_wallet` into this function!)

swap/src/asb/event_loop.rs
thomaseizinger: Also, it is a bit of an anti-pattern to extract the raw value from a newtype (the `as_monero` function call) because the type system no longer helps you in correctly interpreting the result of the computation.

You should be able to express this functionality by implementing `Mul` for `Rate` and `Monero`.

swap/src/monero.rs
thomaseizinger: Also: Missing whitespace above :)

thomaseizinger (changes requested):
Thanks! Left some comments :)

swap/src/asb/event_loop.rs
```diff
@@ -318,16 +318,30 @@ where
...
+            tracing::warn!(
+                    "Your Max bitcoin to sell: {} is smaller than your current balance of monero available to sell {}",
+                    max_buy, max_bitcoin_for_monero
+                );
+            max_bitcoin_buy = max_bitcoin_for_monero
+        }
+
         Ok(BidQuote {
             price: rate.ask().context("Failed to compute asking price")?,
             min_quantity: min_buy,
```
binarybaron, line 343: If `max_bitcoin_buy` is less than `min_buy`, the user cannot initiate a swap at all. We need to handle this case somehow. I suggest that we set both `min_quantity` and `max_quantity` to zero to communicate to the user that the ASB is not accepting swaps at the moment.

binarybaron (changes requested):

swap/src/asb/event_loop.rs
lescuer97: I think we could make another issue for the project so we can solve it in parts

swap/src/asb/event_loop.rs
lescuer97: Thank you for the input! I ended up using decimals at the end. I think the UX for the BTC and XMR nomenclature is better than Piconero and Satoshi one

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +320,40 @@ where
...
+                tracing::warn!("Using Max purchase from config: {}", max_buy);
+                max_buy
+            }
+        };
+
+        if max_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                    "Your Max bitcoin to sell config: {} is bigger than your current balance of Monero available to sell {}",
+                    max_buy, max_bitcoin_for_monero
+                );
```
thomaseizinger, line 347: Can you please extract this calculation into a dedicated function or even better, move it _onto_ the actual types itself?

There is a lot of low-level logic happening here and we don't have any tests for it.

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +320,40 @@ where
...
             .latest_rate()
-            .context("Failed to get latest rate")?;
+            .context("Failed to get latest rate")?
+            .ask()
+            .context("Failed to compute asking price")?;
+
+        let bitcoin_decimal: Decimal = Decimal::from(rate.as_sat()) / Decimal::from(100_000_000);
+
+        let max_bitcoin_for_monero = match (monero_decimal * bitcoin_decimal).to_f64() {
+            Some(val) => bitcoin::Amount::from_btc(val)?,
```
thomaseizinger, line 335: Use `to_u64` instead and `from_sat` please. `from_btc` involves floating point arithmetic!

swap/src/monero.rs
```diff
@@ -360,6 +370,22 @@ mod tests {
...
+        assert_eq!(Decimal::new(10, 0), monero_decimal);
+    }
+
+    #[test]
+    fn create_piconero_offset_decimal() {
+        let monero = "10";
+        let amount = Amount::parse_monero(monero).unwrap();
+        let monero_decimal = amount.as_monero_decimal();
+        assert_eq!(Decimal::new(10, 0), monero_decimal);
+    }
```
thomaseizinger, line 387: The same test twice with a different name?

swap/src/monero.rs
```diff
@@ -99,6 +99,16 @@ impl Amount {
...
 
+    pub fn as_monero_decimal(&self) -> Decimal {
+        let piconero_decimal = Decimal::from(self.0);
+        let pic_offcet_decimal = Self::piconero_offset_decimal();
+        piconero_decimal / pic_offcet_decimal
+    }
+
+    pub fn piconero_offset_decimal() -> Decimal {
+        Decimal::from(PICONERO_OFFSET)
+    }
```
thomaseizinger, line 110: Does not need to be public AFAIK.

swap/src/asb/event_loop.rs
thomaseizinger: If you look at what the function did before, then you can see that it only gathered some data and combined it into a `BidQuote` but the actual calculation of what an _ask_ price is, does not happen in this function.

It would like to it stay that way. In other words: The responsibility of `make_quote` should be to construct a `BidQuote` by gathering data from several dependencies (like wallets, rates etc). The actual calculation of the price is better extracted into a pure function that does the computation based on the inputs. That function we can then test for all its edge-cases.

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +320,40 @@ where
         min_buy: bitcoin::Amount,
         max_buy: bitcoin::Amount,
     ) -> Result<BidQuote> {
+        let monero_decimal = self.monero_wallet.get_balance().await?.as_monero_decimal();
```
thomaseizinger, line 323: Nit: The life-time of this variable is unnecessarily long. It is only used from line 334 so there isn't really any reason why this is at the very top!

thomaseizinger (changes requested):
Thanks for incorporating the feedback!

A few more comments :)

swap/src/asb/event_loop.rs
binarybaron: I don't think we need another PR for this. Afaik this is already done for when the actual swap is started. You just need to subtract the fee from the Monero balance when calculating the `max_quantity`

lescuer97: You are right, I have implemented a the calculation to subtract the fee

swap/src/asb/event_loop.rs
lescuer97: I managed a way to use u64 but had to watch for an underflow for subtracting the Monero fee

swap/src/monero.rs
lescuer97: yeah sorry I actually made an erroneous commit

swap/src/asb/event_loop.rs
binarybaron: Thanks a lot for implementing this fee subtraction thing, looks good to me!

The scenario I originally mentioned where the min_quantity is bigger than the max_quantity is still possible afaik (mistake on my end). I'd suggest than we just check if min_quantity > max_quantity and then return a Quote where both are zero.

If either of you has a better suggestion, I'm definitely open to it.

swap/src/monero.rs
```diff
@@ -99,6 +99,31 @@ impl Amount {
...
+
+    pub fn max_bitcoin_from_balance(&self, satoshis: u64) -> Result<bitcoin::Amount> {
+        let piconero_minus_fee = self.as_piconero().saturating_sub(MONERO_FEE.as_piconero());
+
+        if piconero_minus_fee == 0 {
+            tracing::error!("The current Monero balance is smaller than the basic locking fee");
+            bail!("The current Monero balance is smaller than the basic locking fee");
+        }
```
binarybaron, line 118: Wouldn't it be more logical to just return zero? When calling this function we are essential asking "How much Bitcoin can we convert this into at this exchange rate while accounting for the fee?". If the answer to that questions happens to be zero, that's not really an error imo.

Also, will the ASB return a quote at all when the balance is zero?

lescuer97: the 0 is actually not the real number if I had to use a normal subtraction I would get a negative number but because its a `u64` integer I would get an integer underflow panic, so I returned this as an elegant error to the asb provider. 

For the swap app if you get request a quote from a rendezvous server you wont get one from the the particular ASB with the lower then minimum balance and if you request it directly the connection will drop.

swap/src/asb/event_loop.rs
```diff
@@ -322,10 +322,30 @@ where
         let rate = self
```
thomaseizinger, line 322: Nit: Perhaps this variable should now say `ask_price`?

swap/src/asb/event_loop.rs
```diff
@@ -322,10 +322,30 @@ where
...
-            .context("Failed to get latest rate")?;
+            .context("Failed to get latest rate")?
+            .ask()
+            .context("Failed to compute asking price")?;
+
+        let max_bitcoin_for_monero = self
+            .monero_wallet
+            .get_balance()
+            .await?
+            .max_bitcoin_from_balance(rate.as_sat())?;
```
thomaseizinger, line 333: Why not pass the `bitcoin::Amount` directly? Gives you more type-safety! :)

swap/src/monero.rs
```diff
@@ -99,6 +99,31 @@ impl Amount {
         self.0
     }
 
+    pub fn as_monero_decimal(&self) -> Decimal {
```
thomaseizinger, line 102: Do we still need this now?

swap/src/monero.rs
```diff
@@ -99,6 +99,31 @@ impl Amount {
...
+        let piconero_decimal = Decimal::from(self.0);
+        let pic_offcet_decimal = Self::piconero_offset_decimal();
+        piconero_decimal / pic_offcet_decimal
+    }
+
+    fn piconero_offset_decimal() -> Decimal {
+        Decimal::from(PICONERO_OFFSET)
+    }
+
+    pub fn max_bitcoin_from_balance(&self, satoshis: u64) -> Result<bitcoin::Amount> {
```
thomaseizinger, line 112: I think a more descriptive name would be:


```suggestion
    pub fn max_bitcoin_for_price(&self, ask_price: u64) -> Result<bitcoin::Amount> {
```

swap/src/monero.rs
```diff
@@ -99,6 +99,31 @@ impl Amount {
...
+        let piconero_minus_fee = self.as_piconero().saturating_sub(MONERO_FEE.as_piconero());
+
+        if piconero_minus_fee == 0 {
+            tracing::error!("The current Monero balance is smaller than the basic locking fee");
+            bail!("The current Monero balance is smaller than the basic locking fee");
```
thomaseizinger, line 117: Either log or return error please but not both! That will lead to duplicate information in the logs :)

swap/src/monero.rs
```diff
@@ -99,6 +99,31 @@ impl Amount {
...
+        }
+
+        // There needs to be an offset for difference in zeroes beetween Piconeros and
+        // Satoshis
+        let piconero_calc = (piconero_minus_fee * satoshis) / 1_000_000_000_000;
```
thomaseizinger, line 122: Where does this constant come from? Can this be made more descriptive?

thomaseizinger (changes requested):
Thank you! I think we are getting close to a good solution here. Some more comments :)

swap/src/monero.rs
lescuer97: not really, do I remove it or leave it be?

swap/src/monero.rs
lescuer97: Done!

swap/src/asb/event_loop.rs
lescuer97: Done!

swap/src/monero.rs
thomaseizinger: Remove it then please!

In general, I'd prefer if we strive towards having only code around that is actively used and relevant :)

thomaseizinger: For PRs in particular, it is easier to review if the diff is minimal.

lescuer97: Done!

swap/src/monero.rs
```diff
@@ -99,6 +99,20 @@ impl Amount {
         self.0
     }
 
+    pub fn max_bitcoin_for_price(&self, ask_price: bitcoin::Amount) -> Result<bitcoin::Amount> {
+        let piconero_minus_fee = self.as_piconero().saturating_sub(MONERO_FEE.as_piconero());
+
+        if piconero_minus_fee == 0 {
+            bail!("The current Monero balance is smaller than the basic locking fee");
```
binarybaron, line 106: I don't know if it's really the best idea to drop the connection when the balance is too low. It could lead to unexpected behavior and confuse the user of swap-cli, since he won't get any real feedback, but only some cryptic timeout error.

I still think returning a quote where the `max_quantity` is zero is the best workaround.

What do you think? @thomaseizinger

binarybaron (commented):
Great work, thanks for your effort!

swap/src/monero.rs
thomaseizinger: Thanks for bringing this to my attention!

Yes, I think for a start, this function knows too much about the context it is called in. We should just return 0 here if that is the result of the computation.

If we still want to fail the quote protocol then we should do that one layer up!

Can we have a test run with both implementations and compare the user output?

swap/src/monero.rs
```diff
@@ -360,6 +374,27 @@ mod tests {
...
+    #[test]
+    fn geting_max_bitcoin_to_trade() {
+        let monero = "10";
+        let amount = Amount::parse_monero(monero).unwrap();
+        let bitcoin_price_sats = bitcoin::Amount::from_sat(382_900);
+        let monero_max_from_bitcoin = amount.max_bitcoin_for_price(bitcoin_price_sats).unwrap();
+        assert_eq!(
+            bitcoin::Amount::from_sat(3_828_988),
+            monero_max_from_bitcoin
+        );
```
thomaseizinger, line 386: A code style nit:

Have you heard of "arrange - act - assert" when it comes to testing? It makes tests heaps more readable when we strive to:

- Have a single statement per block (arrange/act/assert) or at least minimise their size
- Use whitespace to separate those blocks

For example:


```suggestion
        let amount = Amount::parse_monero("10").unwrap();
        let bitcoin_price_sats = bitcoin::Amount::from_sat(382_900);

        let monero_max_from_bitcoin = amount.max_bitcoin_for_price(bitcoin_price_sats).unwrap();

        assert_eq!(
            bitcoin::Amount::from_sat(3_828_988),
            monero_max_from_bitcoin
        );
```

lescuer97: No! actually I didn't I thought that the normal thing was to call the function inside the Assertion!. Thank you for bringing that up!

thomaseizinger: I wouldn't call the function inside the assertion, that would merge the "act" and "assert" stage :)

Simply adding whitespace between them makes it visually easy to identify the relevant parts of the test.

lescuer97: just changed the tests

swap/src/monero.rs
lescuer97, line 106: do you want me to return something like this?
```suggestion
             if piconero_minus_fee == 0 {
            return bitcoin::Amount::ZERO;
```

thomaseizinger: Yep, that is better!

lescuer97: I just uploaded the suggestion from @binarybaron . But I think this should be a temporary solution because we are communicating to the user there is min_ buy amount and that could bring confusion. 

If there is not enough Monero to trade maybe we could give the SWAP user a graceful disconnect with a message or something like that

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +319,33 @@ where
...
+            tracing::warn!(
+                    "Your Max bitcoin to sell config: {} is bigger than your current balance of Monero available to sell {}",
+                    max_buy, max_bitcoin_for_monero
+                );
+            return Ok(BidQuote {
+                price: ask_price,
+                min_quantity: min_buy,
+                max_quantity: max_bitcoin_for_monero,
+            });
+        }
```
binarybaron, line 346: If max_bitcoin_buy is less than min_buy, the user cannot initiate a swap at all. We need to handle this case somehow. I suggest that we set both min_quantity and max_quantity to zero to communicate to the user that the ASB is not accepting swaps at the moment. Otherwise there might be a weird case where for example `min_quantity = 1 BTC` and `max_quantity = 0.5 BTC` which doesn't really make sense.

Something like this:
```rust
        if min_buy > max_bitcoin_for_monero {
             tracing::warn!(
                     "Your Monero balance is too low to initiate a swap, as your minimum swap amount is {}. You could at most swap {}",
                     min_buy, max_bitcoin_for_monero
                 );
             return Ok(BidQuote {
                 price: ask_price,
                 min_quantity: bitcoin::Amount::ZERO,
                 max_quantity: bitcoin::Amount::ZERO,
             });
         }
```

binarybaron (changes requested):

swap/src/asb/event_loop.rs
lescuer97: I will wait for @thomaseizinger  response on the solution before upload, but it seems fine to me

thomaseizinger: Yep, seems reasonable to me too. How does the swap tool handle both amounts being 0? Did someone do a test run for this?

lescuer97: I didn't attempt a trade but this where the outputs that  it gave me. 
Screenshot from 2022-03-01 09-05-42-cropped

thomaseizinger: Can you attempt a trade? :)

lescuer97: Okeyy, this are the results: The first time I tried to make a swap I't actually gave me a qr code to deposit and I tried it, as soon as the bitcoin arrived it failed and I received this message: 
Screenshot from 2022-03-01 11-05-47-cropped

Then if I tried to do a swap again it wouldn't even let me swap it would just send the error again.
Screenshot from 2022-03-01 11-07-23

The Bitcoin from the first transaction redeemed correctly

thomaseizinger: Okay, so the behaviour here is that we try to do a swap with the bitcoin already stored in the wallet and don't have you deposit new coins.

I think together with this change, we should add a check to the client whether or not `maximum_amount` is `0` and if so, abort directly.

lescuer97: Done!

swap/src/bin/swap.rs
```diff
@@ -556,6 +556,11 @@ where
 {
     tracing::debug!("Requesting quote");
     let bid_quote = bid_quote.await?;
+
+    if bid_quote.max_quantity == bitcoin::Amount::ZERO {
+        bail!("The Monero seller you contacted can't trade monero at the moment")
```
thomaseizinger, line 561: I know you wrote this from a user's perspective but in the context of this function, I'd like a different error message here. Ideally, we would create a custom error type like `struct ZeroQuote;` and return that and react to that outside of this function.

If you don't want to go that far, I'd suggest just bailing with something like:

```suggestion
        bail!("Received quote with `max_quantity` of 0");
```

And adding a `.context` to the call of `determine_btc_to_swap` here: https://github.com/comit-network/xmr-btc-swap/blob/8f3f4d5895409775506eb256f35e945da1e52de6/swap/src/bin/swap.rs#L110

This will add a "layer" to the error message that should give enough context to the user of what is going wrong. (Read about [anyhow](https://docs.rs/anyhow/latest/anyhow/index.html) if you are not familiar with it.)

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +319,45 @@ where
...
+            .monero_wallet
+            .get_balance()
+            .await?
+            .max_bitcoin_for_price(ask_price);
+
+        if min_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                        "Your Monero balance is too low to initiate a swap, as your minimum swap amount is {}. You could at most swap {}",
+                        min_buy, max_bitcoin_for_monero
+                    );
```
thomaseizinger, line 339: Formatting is a bit off here. Have you run the formatter on this code? :)

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +319,45 @@ where
...
+            return Ok(BidQuote {
```
thomaseizinger, line 340: Nit: Would prefer a newline before `return` statements.

thomaseizinger (changes requested):
Thanks!

Some more comments and a formatting issue but overall, this is looking pretty good! Excited to merge this :)

swap/src/asb/event_loop.rs
lescuer97: No, I forgot to do the command

swap/src/bin/swap.rs
lescuer97: The structure of the Custom error how would you do it? with the same structure as BidQuote?

thomaseizinger: In a nutshell, it works like this:

First, we need to make the high-level decision of the function signature:

- Do we want to use `anyhow::Result` or `std::Result`? `anyhow` provides a boxed error type which can hold anything that implements `std::error::Error`. That is convenient but reduces type-safety a bit because we no longer know exactly, which errors are returned from the function.

At the moment, we use `anyhow::Result` and I'd suggest to keep it just to keep the code-churn of this PR low. If you research about `anyhow` and Rust error handling in general, you should find a few good resources and opinions.

With `anyhow::Error` (the error type of `anyhow::Result`), it is not quite as obvious how to do type-safe error handling. What we can do is:

1. Define a new error struct: `struct ZeroQuoteReceived;` that `derive`s `thiserror::Error` or implements `std::error::Error` manually.
2. Use `bail!(ZeroQuoteReceived)` at the above place.
3. On the call-site of `determine_btc_to_swap`, we can `match` on the returned `Result` (instead of `?` it) and then use the [`anyhow::Error::downcast` API](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast) to check whether the returned `anyhow::Error` "contains" a the `ZeroQuoteReceived` error type. If the `downcast` returns `Ok`, you can then handle that specific case by printing a nice message to the user. In the `Err` case, we can just `return` the error which is equivalent to what the current `?` does.

lescuer97: nice!, didn't knew about downcast

swap/src/bin/swap.rs
```diff
@@ -99,15 +99,24 @@ async fn main() -> Result<()> {
...
                 max_givable,
                 || bitcoin_wallet.sync(),
             )
-            .await?;
+            .await
+            {
+                Ok(val) => val,
+                Err(error) => match error.downcast::<ZeroQuoteReceived>() {
+                    Ok(_) => {
+                        bail!("The Monero seller you contacted can't trade monero at the moment")
```
binarybaron, line 115: Small thing but maybe replace this with something like `Seller's XMR balance is currently too low to initiate a swap, please try again later`. We know what caused the error so we can be a little more specific.

swap/src/asb/event_loop.rs
```diff
@@ -319,13 +319,46 @@ where
...
+                price: ask_price,
+                min_quantity: bitcoin::Amount::ZERO,
+                max_quantity: bitcoin::Amount::ZERO,
+            });
+        }
+
+        if max_buy > max_bitcoin_for_monero {
+            tracing::warn!(
+                    "Your Max bitcoin to sell config: {} is bigger than your current balance of Monero available to sell {}",
```
binarybaron, line 350: Small thing but maybe replace this with something like this so it's a little more consistent with the other log message

`Your Monero balance is too low to initiate a swap with the maximum swap amount {} that you have specified in your config. You can at most swap {}`

swap/src/bin/swap.rs
lescuer97: Much better message hahaha

swap/src/asb/event_loop.rs
lescuer97: Done!

lescuer97:
Just added a small test to check for checking the error message

thomaseizinger (approved):
Nice! Thank you for your patience and incorporating all the feedback :)

To finalise this, can you please:

- Add a changelog entry
- Squash all commits into one

lescuer97:
Done! @thomaseizinger. No patience needed hahaha I learned a lot in the process

thomaseizinger:
> this also fixes #963

If you add this to thr PR description, it will close it automatically!

thomaseizinger:
Looks good! Kicked off CI. If it passes, feel free to merge.

bors d=lescuer97

thomaseizinger:
> > this also fixes #963
> 
> If you add this to thr PR description, it will close it automatically!

Note: Only if you give it its own line like this:

Fixes https://github.com/comit-network/xmr-btc-swap/issues/939.
Fixes https://github.com/comit-network/xmr-btc-swap/issues/963.

PR #989, opened 2022-03-20
Check if binary is latest version

### LINKED ISSUE #988: Check for the latest version
Would be useful for the end user to be notified if there is a new version, so they get the latest features and fixes

lescuer97:
Needs to be decided if this is going to be used for the ASB too and where in the code base the function should be put. 

I'll wait for comments ;)

Fixes #988

### DIFF
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -446,7 +446,7 @@ async fn main() -> Result<()> {
             match swap_state {
                 BobState::Started { .. }
                 | BobState::SwapSetupCompleted(_)
-                | BobState::BtcLocked(_)
+                | BobState::BtcLocked { .. }
                 | BobState::XmrLockProofReceived { .. }
                 | BobState::XmrLocked(_)
                 | BobState::EncSigSent(_)
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -26,6 +26,7 @@ use std::time::Duration;
 use swap::bitcoin::TxLock;
 use swap::cli::command::{parse_args_and_apply_defaults, Arguments, Command, ParseResult};
 use swap::cli::{list_sellers, EventLoop, SellerStatus};
+use swap::common::check_latest_version;
 use swap::database::open_db;
 use swap::env::Config;
 use swap::libp2p_ext::MultiAddrExt;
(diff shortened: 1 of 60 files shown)


swap/src/bin/swap.rs
```diff
@@ -629,10 +692,43 @@ where
     Ok((btc_swap_amount, fees))
 }
 
+pub async fn check_latest_version() -> Result<()> {
+    const GITHUB_LATEST_VERSION_URL: &str =
+        "https://github.com/comit-network/xmr-btc-swap/releases/latest";
+
+    let response = match reqwest::get(GITHUB_LATEST_VERSION_URL).await {
+        Ok(res) => res,
+        Err(_) => bail!(
```
delta1, line 701: I don't think it should ever bail if it can't check for an update.

lescuer97: should I catch and print the exception inside the function instead that calling?

delta1: oh sorry! you are actually matching on the error and logging at the callsite so I think that's fine 👍

binarybaron:
Thanks for submitting this. I think it's a reasonable feature to have. Can you rebase on master? the tests will fail otherwise

swap/src/bin/swap.rs
```diff
@@ -67,6 +67,12 @@ async fn main() -> Result<()> {
             let swap_id = Uuid::new_v4();
 
             cli::tracing::init(debug, json, data_dir.join("logs"), Some(swap_id))?;
+
+            match check_latest_version().await {
```
binarybaron, line 71: Couldn't we match and print this error in the `check_latest_version` function? We're not using the err for anything else and that would avoid some of this duplication.

binarybaron (commented):
LGTM but left one comment

swap/src/bin/swap.rs
lescuer97: I'll make a note to change this and I will see how it looks. Thank you!

binarybaron:
bors try

lescuer97:
bors try

delta1:
Hey @lescuer97 i think this could be DRYed up a bit by doing the check in one place per binary before matching on the command? 

If you have time to try that please give it a go otherwise I’ll add a commit later today

lescuer97:
> Hey @lescuer97 i think this could be DRYed up a bit by doing the check in one place per binary before matching on the command?
> 
> If you have time to try that please give it a go otherwise I’ll add a commit later today

The reason this action gets repeated a lot, is because the loggin tracing gets activated inside the commands if I excecute it before: 
`cli::tracing::init`

it will not print anything, we need to have just one initialization, that could be posible but that is just another issue

delta1:
Pushed an update @lescuer97 thanks for your work on this!

delta1:
bors merge

PR #995, opened 2022-04-09
Update dprint plugins to fix CI error

binarybaron:
Closes https://github.com/comit-network/xmr-btc-swap/issues/994

PR #996, opened 2022-04-10
Implement `Identify` protocol to make network and version of ASB available to peers

binarybaron:
As suggested by @thomaseizinger, this PRs implement the [Identify](https://github.com/libp2p/specs/blob/master/identify/README.md) protocol for the ASB. It makes the rendezvous namespace (network) and the version available to other peers. This is not a breaking change yet because Bob does not rely on it yet. This can for example be used by application building on top of this project (e.g [unstoppableswap-gui](http://github.com/UnstoppableSwap/unstoppableswap-gui/)) or might serve for some kind of compatibility insurance in the future.
I'd love some feedback on this from the more experienced developers. I might very well have made some obvious mistake.

> In line with [HTTP user agent strings](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) (which is what the `agentVersion` field in identify is based on), we could do something like:
> 
> `asb/0.8.0 (Mainnet)`
> 
> A PR adding `/identify` with these details is likely going to be accepted.
(https://github.com/comit-network/xmr-btc-swap/discussions/492#discussioncomment-970294)

### DIFF
--- a/swap/src/asb/network.rs
+++ b/swap/src/asb/network.rs
@@ -13,6 +13,7 @@ use libp2p::core::connection::ConnectionId;
 use libp2p::core::muxing::StreamMuxerBox;
 use libp2p::core::transport::Boxed;
 use libp2p::dns::TokioDnsConfig;
+use libp2p::identify::{Identify, IdentifyConfig, IdentifyEvent};
 use libp2p::ping::{Ping, PingConfig, PingEvent};
 use libp2p::request_response::{RequestId, ResponseChannel};
 use libp2p::swarm::{
@@ -111,6 +112,7 @@ pub mod behaviour {
         pub swap_setup: alice::Behaviour<LR>,
         pub transfer_proof: transfer_proof::Behaviour,
         pub encrypted_signature: encrypted_signature::Behaviour,
+        pub identify: Identify,
 
         /// Ping behaviour that ensures that the underlying network connection
         /// is still alive. If the ping fails a connection close event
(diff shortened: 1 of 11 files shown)


binarybaron:
The CI is failing because of an issue with dprint. I'll rebase this PRs once [995](https://github.com/comit-network/xmr-btc-swap/pull/995) has been merged.

swap/src/bin/asb.rs
```diff
@@ -136,23 +136,23 @@ async fn main() -> Result<()> {
             };
 
             let kraken_rate = KrakenRate::new(config.maker.ask_spread, kraken_price_updates);
+            let namespace = {
+                if testnet {
+                    XmrBtcNamespace::Testnet
+                } else {
+                    XmrBtcNamespace::Mainnet
+                }
+            };
```
delta1, line 145: Pretty sure you can just do this 

```suggestion
            let namespace = if testnet {
                XmrBtcNamespace::Testnet
            } else {
                XmrBtcNamespace::Mainnet
            };
```

delta1:
Hmm maybe just a spurious test failure? Looks like a timeout, suggest running it again to see if the same happens

binarybaron:
Already ran it twice. I'm pretty sure it's related to some change I made.

swap/src/bin/asb.rs
binarybaron: Thanks!

swap/src/asb/network.rs
```diff
@@ -163,6 +172,12 @@ pub mod behaviour {
         }
     }
 
+    impl From<IdentifyEvent> for OutEvent {
+        fn from(_: IdentifyEvent) -> Self {
+            OutEvent::Other
```
binarybaron, line 177: Maybe this needs its own `OutEvent`?

delta1: I saw that but then wasn't sure. Will have a closer look when I can.

binarybaron:
The test is also failing on this PR https://github.com/comit-network/xmr-btc-swap/runs/4912948900?check_suite_focus=true so might not be related to any of my changed after all.

binarybaron:
bors try

delta1:
bors try

delta1:
bors retry

binarybaron:
bors try

delta1:
bors try

binarybaron:
@delta1 @lescuer97 All tests are passing now✅ I fixed them by implementing the protocol for the CLI as well. I'd merge this after both of you take a look again and give your OK. The changes touch the network layer so I think we should be careful not to break anything here (although it doesn't seem like anything could cause issues)

binarybaron:
bors try

delta1:
Great well done will review today

delta1 (approved):
LGTM. All tests passed locally.

delta1:
bors merge

PR #1008, opened 2022-04-14
Upgrade libp2p and stabilized the version

lescuer97:
This is an upgrade for libp2p, I did a previous PR but got mixed up with rebases

### DIFF
--- a/swap/src/cli/event_loop.rs
+++ b/swap/src/cli/event_loop.rs
@@ -9,6 +9,7 @@ use anyhow::{Context, Result};
 use futures::future::{BoxFuture, OptionFuture};
 use futures::{FutureExt, StreamExt};
 use libp2p::request_response::{RequestId, ResponseChannel};
+use libp2p::swarm::dial_opts::DialOpts;
 use libp2p::swarm::SwarmEvent;
 use libp2p::{PeerId, Swarm};
 use std::collections::HashMap;
@@ -81,7 +82,7 @@ impl EventLoop {
     }
 
     pub async fn run(mut self) {
-        match self.swarm.dial(&self.alice_peer_id) {
+        match self.swarm.dial(DialOpts::from(self.alice_peer_id)) {
             Ok(()) => {}
             Err(e) => {
                 tracing::error!("Failed to initiate dial to Alice: {}", e);
@@ -167,12 +168,13 @@ impl EventLoop {
                             tracing::info!("Successfully closed connection to Alice");
                             return;
                         }
-                        SwarmEvent::UnreachableAddr { peer_id, address, attempts_remaining, error } if peer_id == self.alice_peer_id && attempts_remaining == 0 => {
-                            tracing::warn!(%address, "Failed to dial Alice: {}", error);
+                        SwarmEvent::OutgoingConnectionError { peer_id,  error } if matches!(peer_id, Some(alice_peer_id) if alice_peer_id == self.alice_peer_id) => {
+                            tracing::warn!( "Failed to dial Alice: {}", error);
 
                             if let Some(duration) = self.swarm.behaviour_mut().redial.until_next_redial() {
                                 tracing::info!("Next redial attempt in {}s", duration.as_secs());
                             }
+
                         }
                         _ => {}
                     }
--- a/swap/src/cli/list_sellers.rs
+++ b/swap/src/cli/list_sellers.rs
@@ -43,6 +43,7 @@ pub async fn list_sellers(
         .behaviour_mut()
         .quote
         .add_address(&rendezvous_node_peer_id, rendezvous_node_addr.clone());
+
     swarm
         .dial(DialOpts::from(rendezvous_node_peer_id))
         .context("Failed to dial rendezvous node")?;
@@ -162,15 +163,17 @@ impl EventLoop {
                                 );
                             } else {
                                 let address = endpoint.get_remote_address();
+                                tracing::debug!(%peer_id, %address, "Connection established to peer");
                                 self.reachable_asb_address.insert(peer_id, address.clone());
                             }
                         }
                         SwarmEvent::OutgoingConnectionError { peer_id, error } => {
-                            if let Some(peer_id_from_error) = peer_id {
-                                if peer_id_from_error == self.rendezvous_peer_id {
+                            if let Some(peer_id) = peer_id {
+                                if peer_id == self.rendezvous_peer_id {
                                     tracing::error!(
-                                        "Failed to connect to rendezvous point at {}: {}",
-                                        &self.rendezvous_addr,
+                                        %peer_id,
+                                        %self.rendezvous_addr,
+                                        "Failed to connect to rendezvous point: {}",
                                         error
                                     );
 
@@ -178,30 +181,23 @@ impl EventLoop {
                                     return Vec::new();
                                 } else {
                                     tracing::error!(
-                                        "You connected to the wrong Peer: {} Error: {}",
-                                        &peer_id_from_error,
+                                        %peer_id,
+                                        "Failed to connect to peer: {}",
                                         error
                                     );
-                                    // if for some reason the peer is not the same it will return empty too.
-                                    // this is just for the CLI to not get stuck
-                                    return Vec::new();
-                                }
-                            } else {
-                                tracing::debug!(
-                                    "Failed to connect to peer at {}: {}",
-                                    &self.rendezvous_addr,
-                                    error
-                                );
-                                self.unreachable_asb_address.insert(self.rendezvous_peer_id, self.rendezvous_addr.clone());
+                                    self.unreachable_asb_address.insert(peer_id, Multiaddr::empty());
 
-                                match self.asb_quote_status.entry(self.rendezvous_peer_id) {
-                                    Entry::Occupied(mut entry) => {
-                                        entry.insert(QuoteStatus::Received(Status::Unreachable));
-                                    },
-                                    _ => {
-                                        tracing::debug!(%self.rendezvous_peer_id, %error, "Connection error with unexpected peer")
+                                    match self.asb_quote_status.entry(peer_id) {
+                                        Entry::Occupied(mut entry) => {
+                                            entry.insert(QuoteStatus::Received(Status::Unreachable));
+                                        },
+                                        _ => {
+                                            tracing::debug!(%peer_id, %error, "Connection error with unexpected peer");
+                                        }
                                     }
                                 }
+                            } else {
+                                tracing::debug!("Failed to connect (no peer id): {}", error);
                             }
                         }
                         SwarmEvent::Behaviour(OutEvent::Rendezvous(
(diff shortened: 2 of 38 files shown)


lescuer97:
bors try

swap/src/cli/event_loop.rs
```diff
@@ -167,12 +168,9 @@ impl EventLoop {
                             tracing::info!("Successfully closed connection to Alice");
                             return;
                         }
-                        SwarmEvent::UnreachableAddr { peer_id, address, attempts_remaining, error } if peer_id == self.alice_peer_id && attempts_remaining == 0 => {
-                            tracing::warn!(%address, "Failed to dial Alice: {}", error);
+                        SwarmEvent::OutgoingConnectionError { peer_id,  error } if matches!(peer_id, Some(alice_peer_id) if alice_peer_id == self.alice_peer_id) => {
+                            tracing::warn!( "Failed to dial Alice: {}", error);
 
-                            if let Some(duration) = self.swarm.behaviour_mut().redial.until_next_redial() {
```
binarybaron, line 173: Why was this removed?

binarybaron (commented):
just left one small comment.

swap/src/cli/event_loop.rs
lescuer97: this was a mistake thank you for catching it

binarybaron: Ah great

swap/src/cli/list_sellers.rs
```diff
@@ -164,32 +165,32 @@ impl EventLoop {
                                 self.reachable_asb_address.insert(peer_id, address.clone());
                             }
                         }
-                        SwarmEvent::UnreachableAddr { peer_id, error, address, .. } => {
-                            if address == self.rendezvous_addr {
+                        SwarmEvent::OutgoingConnectionError { peer_id, error } => {
+                            if let Some(peer_id_rendevous) = peer_id {
```
binarybaron, line 169: I think you are making some big changes to the logic here, not only refactoring. Not sure that was intended by you.

**Old:**
```rust
if address == self.rendezvous_addr {
    tracing::error!(
        "Failed to connect to rendezvous point at {}: {}",
        address,
        error
    );

    // if the rendezvous node is unreachable we just stop
    return Vec::new();
} else {
    tracing::debug!(
        "Failed to connect to peer at {}: {}",
        address,
        error
    );
    self.unreachable_asb_address.insert(peer_id, address.clone());

    match self.asb_quote_status.entry(peer_id) {
        Entry::Occupied(mut entry) => {
            entry.insert(QuoteStatus::Received(Status::Unreachable));
        },
        _ => {
            tracing::debug!(%peer_id, %error, "Connection error with unexpected peer")
        }
    }
}
```

New:
````rust
if let Some(peer_id_rendevous) = peer_id {
  tracing::error!(
      "Failed to connect to rendezvous point at {}: {}",
      self.rendezvous_addr,
      error
  );

  self.unreachable_asb_address.insert(peer_id_rendevous, self.rendezvous_addr.clone());

  match self.asb_quote_status.entry(peer_id_rendevous) { // <-- You moved this from the else block into the if block
      Entry::Occupied(mut entry) => {
          entry.insert(QuoteStatus::Received(Status::Unreachable));
      },
      _ => {
          tracing::debug!(%peer_id_rendevous, %error, "Connection error with unexpected peer")
      }
  }
} else {
  tracing::error!(
      "Failed to connect to rendezvous point at {}: {}",
      self.rendezvous_addr,
      error
  );
  // if the rendezvous node is unreachable we just stop
  return Vec::new(); // <-- You moved this from the if block to the else block
}
```

swap/src/cli/list_sellers.rs
```diff
@@ -164,32 +165,32 @@ impl EventLoop {
...
                                         entry.insert(QuoteStatus::Received(Status::Unreachable));
                                     },
                                     _ => {
-                                        tracing::debug!(%peer_id, %error, "Connection error with unexpected peer")
+                                        tracing::debug!(%peer_id_rendevous, %error, "Connection error with unexpected peer")
                                     }
                                 }
+                            } else {
+                                tracing::error!(
+                                    "Failed to connect to rendezvous point at {}: {}",
```
binarybaron, line 188: This should be "Failed to connect to peer at {}: {}". The peer is not the rendezvous point

binarybaron (changes requested):
Left two more comments

lescuer97:
bors try

lescuer97:
bors try

lescuer97:
there are conflicts becuase I merged with upstream after the [Identify protocol feature](https://github.com/comit-network/xmr-btc-swap/pull/996) was added, trying to solve them right now.

swap/src/cli/list_sellers.rs
binarybaron: Did I make a mistake here? Aren't you checking if its the rendezvous point in line 169 and if its not you will print this message?

binarybaron (changes requested):
Did we make a swap on this branch already?

lescuer97:
> Did we make a swap on this branch already?
I built and did one in my local machine against your ASB, but of course your ASB is not stabilized

lescuer97:
@binarybaron because I rebased instead of doing a direct commit for a review I lost your comment (sorry). I'm now checking for the rendezvous using the peerId instead of the address(no longer available), you will see an additional else statement is just an in case for the CLI to not get stuck

swap/src/cli/list_sellers.rs
```diff
@@ -164,30 +165,42 @@ impl EventLoop {
...
-                                    address,
+                                    &self.rendezvous_addr,
                                     error
                                 );
 
                                 // if the rendezvous node is unreachable we just stop
                                 return Vec::new();
+                            } else {
+                                tracing::error!(
+                                    "You connected to the wrong Peer: {} Error: {}",
```
binarybaron, line 182: I was so confused because of the indentation :D

Do you know if these events are also called for connections with asbs?
If they are we didn't really fail to connect to a wrong peer but we failed to connect to an asb.

swap/src/cli/list_sellers.rs
```diff
@@ -164,30 +165,42 @@ impl EventLoop {
...
+                                );
+                                     // if for some reason the peer is not the same it will return empty too.
+                                     // this is just for the CLI to not get stuck
+                                     return Vec::new();
+                            }
                             } else {
                                 tracing::debug!(
                                     "Failed to connect to peer at {}: {}",
-                                    address,
+                                    &self.rendezvous_addr,
```
binarybaron, line 193: If these events are also called for connections with asbs, then we don't know if the address is `self.rendezvous_addr` right?

swap/src/cli/list_sellers.rs
```diff
@@ -164,30 +165,42 @@ impl EventLoop {
...
                                     error
                                 );
-                                self.unreachable_asb_address.insert(peer_id, address.clone());
+                                self.unreachable_asb_address.insert(self.rendezvous_peer_id, self.rendezvous_addr.clone());
```
binarybaron, line 196: Same thing. I think the peer might be an asb. We are adding a rendezvous point here to a list called `self.unreachable_asb_address ` which doesn't really make sense.

binarybaron (changes requested):
Left some comments on the event handling issue. I'm still not really sure what's wrong there.  Maybe @delta1 can provide some insight when he reviews the PR

binarybaron:
> @binarybaron because I rebased instead of doing a direct commit for a review I lost your comment (sorry). I'm now checking for the rendezvous using the peerId instead of the address(no longer available), you will see an additional else statement is just an in case for the CLI to not get stuck

Thank you! No worries

delta1:
Looking good to me, please address the requested changes by @binarybaron and we can get this in!

delta1:
bors try

swap/src/cli/list_sellers.rs
delta1: 100% right, I've updated the logic and added some more logic here: 7838442947c471ce6deb91f19e989bd636f6f076

swap/src/cli/list_sellers.rs
delta1: Agreed, fixed in same commit: 7838442947c471ce6deb91f19e989bd636f6f076

swap/src/cli/list_sellers.rs
delta1: Updated this to reflect the following cases: 
- failed to connect to the rendezvous, this is the case where we bail out
- failed to connect to a peer and have the peer_id in the event, record this peer_id in the unreachable list 
- failed to connect to a peer and peer_id is None, just log the error 

7838442947c471ce6deb91f19e989bd636f6f076

swap/src/cli/list_sellers.rs
delta1: Agreed with you and fixed in 7838442947c471ce6deb91f19e989bd636f6f076

swap/src/cli/list_sellers.rs
delta1: Yeah it's awful that rustfmt fails in the select! macro. Updated this indentation correctly here: dc68344b62db029809a4999c11a645238e7c6ad7 

In general, we want to keep as little logic as possible in the select macro and rather call a function with the event. I'll do that update sometime after this goes in so as not to complicate this PR too much.

delta1:
I did test this successfully on testnet yesterday, I'm keen to get it in. Pushed some fixes regarding @binarybaron's comments which were spot on. 

Once the CI passes then I'm going to run a manual test again with multiple ASBs and a rendezvous to make sure that logic works right. 

In the meanwhile, libp2p is on version 0.46 but I'll do that upgrade down the line since it's a bit more complicated.

lescuer97:
@delta1 did you test with both asb and cli?

swap/src/cli/list_sellers.rs
binarybaron: Looks good to me. I wasn't sure if I missed something here, but you're spot on!

swap/src/cli/list_sellers.rs
binarybaron: Thanks!

swap/src/cli/list_sellers.rs
binarybaron: Awesome, thanks!

delta1:
> @delta1 did you test with both asb and cli?

Yes I ran a testnet swap locally with asb and swap built from this branch.

binarybaron:
LGTM! I'm very keen to finally get this merged as well. I'll also do some manual testing but feel free to merge once you think you've tested enough.

delta1:
> This should be backwards compatible, correct?

It should be - I'll try test that too

lescuer97:
The RPC error that I reported on this [issue](https://github.com/comit-network/xmr-btc-swap/issues/1061) happened in a build of this pull request

delta1:
> The RPC error that I reported on this [issue](https://github.com/comit-network/xmr-btc-swap/issues/1061) happened in a build of this pull request

I don't think it's related, the issue has occurred on 0.10.2 as well

delta1:
I'll do some more testing this weekend but I think this is ready to go in

delta1:
I'm not 100% that this is backwards compatible, and trying to test with asb 0.10.2 isn't great because of the lack of log output. 

So I plan to include this with the upgrade to Monero in #1081 instead

delta1:
bors merge

delta1 (approved):

PR #1062, opened 2022-07-26
fix: change logs to UTC due to bug in time crate

delta1:
On Linux and macOS, no program output was being observed. This is referenced in the [LocalTime] docs for `tracing-subscriber`,
which links to this [unsoundness issue] in the time crate.

Rather than introducing a possible vector for undefined behaviour and segfaults, I have just changed the logging to use UTC time instead.

When running the ASB as a systemd service, one would generally use the `--disable-timestamps` flag anyway as systemd adds its own timestamps which can be local to the server.

If the situation with `tracing-subscriber` and the time crate is fixed then this can be updated.

This commit also updates the `tracing-subscriber` and `tracing-appender` dependencies, closing #987.

[LocalTime]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.LocalTime.html
[unsoundness issue]: https://github.com/time-rs/time/issues/293#issuecomment-748151025

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,6 @@ flycheck_*.el
 .swap-db/
 
 # End of https://www.toptal.com/developers/gitignore/api/rust,clion+all,emacs
+
+*.log
+.vscode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Change Monero nodes to [Rino tool nodes](https://community.rino.io/nodes.html)
-- Revert logs to use rfc3339 local time formatting.
 - Update from monero v17.2.0 to monero v17.3.0
 - Always write logs as JSON to files
+- Change to UTC time for log messages, due to a bug causing no logging at all to be printed (linux/macos), and an [unsoundness issue](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.LocalTime.html) with local time in [the time crate](https://github.com/time-rs/time/issues/293#issuecomment-748151025)
 
 ### Added
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -523,7 +523,6 @@ dependencies = [
  "libc",
  "num-integer",
  "num-traits",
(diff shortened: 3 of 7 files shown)


binarybaron (approved):
LGTM, thanks for fixing this!
Have you had the chance to test this on Windows, Linux and Mac OS?
We've already had the same issue some time ago when we updated `tracing-appender` (Bug was introduced in https://github.com/comit-network/xmr-btc-swap/commit/9e85637419594f69d05e0d03767d66b07001a0ae and fixed in https://github.com/john-parker1/xmr-btc-swap/commit/3feaf32dfaf686b2a27c815590e0085f36c7d9e3).

delta1:
> Have you had the chance to test this on Windows, Linux and Mac OS? 

Actually only tested on Linux, although the linked issue in `time` indicates it happens on macOS too. I do have a windows box I can test on so will give that a spin. 

Yeah I saw that it had previously been reverted, but in this case the lack of any output at all was driving me mad until I found this issue. 

I wonder if it's worth trying to remove the tracing timers to see what tracing-subscriber does by default? It looks like it uses [SystemTime](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.SystemTime.html)

binarybaron:
Seems to work on Mac OS:
```bash
$ ./swap_time --testnet balance
2022-07-26T13:05:02.04807Z  INFO Checked Bitcoin balance balance=0.00000000 BTC
$ ./swap_time --testnet --json balance
{"timestamp":"2022-07-26T13:05:09.955257Z","level":"INFO","fields":{"message":"Checked Bitcoin balance","balance":"0.00000000 BTC"}}
```

delta1:
> I wonder if it's worth trying to remove the tracing timers to see what tracing-subscriber does by default? It looks like it uses [SystemTime](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.SystemTime.html) 

It looks like this is UTC as well, at least on my Linux machine. 

> Seems to work on Mac OS 

Thanks for testing! As soon as I confirm Windows then I'll let you know.

delta1:
bors merge

delta1:
bors retry

PR #1068, opened 2022-07-28
fix potential overflow in `max_bitcoin_for_price`

delta1:
In testing, ASB panicked in `max_bitcoin_for_price` when the Monero balance x Bitcoin price was enough to overflow `u64`.

This PR changes the function to do the piconero offset division first, and then to use `checked_mul` to return None if the calculation would overflow. This required changing the function return signature to an `Option`. Additional tests for the function were also added.

MONERO_FEE was also changed from 0.000030 to 0.000016, which is still double the current median transaction fee listed at
https://www.monero.how/monero-transaction-fees as of 2022-07-28.


delta1:
bors merge

PR #1082, opened 2022-08-01
ci: fix windows binary uploads in ci.yml

delta1:
- CI jobs weren't actually uploading the windows artifacts as they didn't specify the `.exe` extension. 
- Updates the upload-artifact action to v3, and adds an error for files not found


delta1:
bors merge

PR #1188, opened 2022-10-28
[0.11.x] fix: immediately fetch transaction status upon subscription

binarybaron:
Immediately fetch transaction status upon subscription instead of waiting 1 minute before making call to electrum server


binarybaron:
Can we still get this in? @delta1

delta1:
Yeah absolutely - just fixing the CI workflow for the draft release then I'll merge this

delta1 (approved):

Issue #1192, opened 2022-11-01
Use unlocked Monero balance to calculate `max_bitcoin_for_monero`


binarybaron:
**Describe the bug**
We need to use the unlocked Monero balance instead of normal Monero balance to calculate `max_bitcoin_for_monero`.

**Lost/trapped Funds**
Yes or No

**Debug logs**
Please post your debug logs. You find them in:
Linux: /home/<user>/.local/share/xmr-btc-swap/logs
OSX: /Users/<user>/Library/ApplicationSupport/xmr-btc-swap/logs

**Platform (please complete the following information):**

- CLI or ASB [eg. CLI]
- Software Version [e.g. 0.8.3]
- Operating System: [e.g. Ubuntu]

delta1:
Nice, I was wondering about this

delta1:
https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#get_balance

PR #1195, opened 2022-11-07
fix(asb): use unlocked monero balance for quotes

delta1:
Related issue #1192 

- changes the ASB to use the unlocked (spendable) Monero balance when providing a quote
- upgrades monero docker images and downloads to v0.18.1.2 
- includes "immediately fetch transaction status upon subscription" from 0.11.x branch #1188  

This will also be backported to [0.11.x](https://github.com/comit-network/xmr-btc-swap/tree/0.11.x) branch for a new release

### DIFF
--- a/monero-harness/src/image.rs
+++ b/monero-harness/src/image.rs
@@ -25,7 +25,7 @@ impl Image for Monerod {
     type EntryPoint = str;
 
     fn descriptor(&self) -> String {
-        "rinocommunity/monero:v0.18.0.0".to_owned()
+        "rinocommunity/monero:v0.18.1.2".to_owned()
     }
 
     fn wait_until_ready<D: Docker>(&self, container: &Container<'_, D, Self>) {
@@ -70,7 +70,7 @@ impl Image for MoneroWalletRpc {
     type EntryPoint = str;
 
     fn descriptor(&self) -> String {
-        "rinocommunity/monero:v0.18.0.0".to_owned()
+        "rinocommunity/monero:v0.18.1.2".to_owned()
     }
 
     fn wait_until_ready<D: Docker>(&self, container: &Container<'_, D, Self>) {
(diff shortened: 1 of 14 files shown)


monero-harness/src/image.rs
```diff
@@ -229,6 +229,7 @@ impl IntoIterator for MoneroWalletRpcArgs {
             format!("--daemon-address={}", self.daemon_address),
             format!("--rpc-bind-port={}", RPC_PORT),
             "--log-level=4".to_string(),
+            "--allow-mismatched-daemon-version".to_string(), /* https://github.com/monero-project/monero/issues/8600 */
```
binarybaron, line 232: Did you experience the problem described in the issue yourself?

swap/src/bin/asb.rs
```diff
@@ -236,16 +246,13 @@ async fn main() -> Result<()> {
...
             let monero_balance = monero_wallet.get_balance().await?;
+            tracing::info!(%monero_balance);
 
-            tracing::info!(
-                %bitcoin_balance,
-                %monero_balance,
-                "Current balance");
+            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_balance = bitcoin_wallet.balance().await?;
+            tracing::info!(%bitcoin_balance);
```
binarybaron, line 255: Could you merge this log message into one and add the message "Current balance" again? Makes it easier for the log to be parsed by scripts/the GUI etc.

binarybaron (changes requested):
Left two small comments. Rest LGTM!

monero-harness/src/image.rs
delta1: Yeah exactly

swap/src/bin/asb.rs
delta1: Sure will do

delta1: Added as an additional log message

swap/src/bin/asb.rs
```diff
@@ -236,16 +246,14 @@ async fn main() -> Result<()> {
...
```
binarybaron, line 255: We don't need to print this twice

binarybaron (changes requested):
@delta1 I left one small comment then we can get this merged

binarybaron:
bors retry

PR #1207, opened 2022-11-12
Add monero-wallet-rpc `DOWNLOAD_URL` for M1, M2 Mac

binarybaron:


swap/src/monero/wallet_rpc.rs
```diff
@@ -23,6 +23,9 @@ const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-x64-v
 #[cfg(all(target_os = "macos", target_arch = "arm"))]
 const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-armv8-v0.18.0.0.tar.bz2";
 
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-armv8-v0.18.0.0.tar.bz2";
+
```
delta1, line 28: Oops, this is what I was trying to do above but got the target arch wrong 😅 

So I think lines 23 - 25 can be removed.

binarybaron: ah you're right

delta1 (approved):
bors r+

delta1:
Thanks @binarybaron - do you want to backport this to the 0.11.x branch too?

binarybaron:
> Thanks @binarybaron - do you want to backport this to the 0.11.x branch too?

Sure. I'll open a PR to merge into `0.11.x`.

About the incompatibility between `0.11.x` and the `main` branch: How do we want to proceed? Do you think we can at least  provide some backwards compatibility by somehow converting the old wallet format into the new one?
I think that's the biggest problem with the new release. It's pretty inconvenient if everyone has to *delete* their old wallet to be able to use the new version of the `CLI`/`ASB`.

delta1:
Yeah I’m still trying to work on a migration for that issue :/ 

until then releases can stay on the 0.11.x branch 

bors retry

Issue #1183, opened 2022-10-25
create a test in CI that tests opening wallet/s created with previous release version 


delta1:
Since #1116, trying to open a Testnet bitcoin wallet created with version 0.11.0 fails with `ChecksumMismatch`

The BDK dependency was bumped from 0.16 -> 0.22 and had a number of breaking changes.

I will try see what changed in BDK to cause that, but besides that I think this issue shows it's worth having a test in CI to test opening wallets created with the previous release version.

_Originally posted by @delta1 in https://github.com/comit-network/xmr-btc-swap/issues/1182#issuecomment-1290263366_

Issue #1182, opened 2022-10-24
Bitcoin wallet broken "ChecksumMismatch"


binarybaron:
**Describe the bug**
The bitcoin wallet is disfunctional and can't be initialized.

**Lost/trapped Funds**
Yes or No

**Debug logs**

```
./swap --testnet export-bitcoin-wallet
2022-10-24T16:46:03.72861Z  INFO Logging initialized to /Users/***/Library/Application Support/xmr-btc-swap/cli/testnet/logs
Error: Failed to initialize Bitcoin wallet

Caused by:
    ChecksumMismatch
```

**Platform (please complete the following information):**

- CLI
- Software Version: HEAD (preview)
- Operating System: M2 Mac OS

delta1:
I think I hit this before - is it possible you tried to open a mainnet wallet with the `--testnet` flag or vice versa?

delta1:
This comes from the call into [`bdk::Wallet::new`](https://docs.rs/bdk/latest/bdk/wallet/struct.Wallet.html#method.new) that says 

`The only way this can fail is if the descriptors passed in do not match the checksums in database` 

I suspect the wallet you're trying to open is for the other network.

delta1:
> M2 Mac OS 

NOICE

binarybaron:
> I hit this before - is it possible you tried to open a mainnet wallet with the `--testnet` flag or vice versa?

I don't think so. Why would that be the case?

binarybaron:
Can you reprodue this issue? It doesn't occur on the latest stable release (https://github.com/comit-network/xmr-btc-swap/releases/tag/0.11.0). Only on preview.

Maybe it was caused by the upgrade to bdk 0.23 https://github.com/comit-network/xmr-btc-swap/commit/ef7922f898191cfcf1e843af70c6a010a9d733d7 ?

delta1:
> > I hit this before - is it possible you tried to open a mainnet wallet with the `--testnet` flag or vice versa?
> 
> I don't think so. Why would that be the case?

I’ve had it happen before - the testnet descriptor from the seed would be different to mainnet

delta1:
> Can you reprodue this issue? It doesn't occur on the latest stable release (https://github.com/comit-network/xmr-btc-swap/releases/tag/0.11.0). Only on preview.
> 
[...]

PR #1219, opened 2022-11-22
feat: add wallet database migration from older bdk

delta1:
- upgrades to bdk 0.24 #1198
- adds a regression test for opening older wallets #1183
- adds a migration for older wallets that encounter the ChecksumMismatch error #1182


binarybaron:
Wow! Amazing work!!!
I'll try to review the code a bit more in detail later today but it does look really good on first impression.

binarybaron (approved):
Lgtm!

PR #1230, opened 2022-11-30
docs: update sqlx script and add ci

delta1:
- upgrades sqlx to 0.6
- fixes `sqlite_dev_setup.sh` script
- adds ci to test the script

### DIFF
--- a/swap/sqlite_dev_setup.sh
+++ b/swap/sqlite_dev_setup.sh
@@ -1,6 +1,14 @@
-# crated temporary DB
-# run the migration scripts to create the tables
-# prepare the sqlx-data.json rust mappings
+#!/bin/bash
+
+# run this script from the swap dir
+# make sure you have sqlx-cli installed: cargo install sqlx-cli
+
+# this script creates a temporary sqlite database
+# then runs the migration scripts to create the tables (migrations folder)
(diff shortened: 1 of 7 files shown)


swap/sqlite_dev_setup.sh
```diff
@@ -1,6 +1,14 @@
...
+# this script creates a temporary sqlite database
+# then runs the migration scripts to create the tables (migrations folder)
+# then it prepares the offline sqlx-data.json rust mappings
 DATABASE_URL=sqlite:tempdb cargo sqlx database create
 DATABASE_URL=sqlite:tempdb cargo sqlx migrate run
-DATABASE_URL=sqlite:./swap/tempdb cargo sqlx prepare -- --bin swap
\ No newline at end of file
+# needs the absolute path here
+# https://github.com/launchbadge/sqlx/issues/1399
+DB_PATH=$(realpath tempdb)
```
binarybaron, line 13: Can we replace `realpath` with something else? This does not exist on Mac OS

binarybaron: I *think* this should work:
`DB_PATH=$(readlink -f -- "$@" tempdb)
`

binarybaron: Yea works for me!

PR #1259, opened 2022-12-30
Allow asb to set a bitcoin address that is controlled by the asb itself to redeem/punish Bitcoin to

binarybaron:


binarybaron:
@delta1 @lescuer97 I'm just going to merge this because I want to get it included in the new release. I hope that's okay. If you have any issues with the changes made we can revert the commit.

PR #1263, opened 2023-01-04
feat(swap): merge cancel/refund commands into one command

### LINKED ISSUE #1020: [0.2 XMR] Consider merging refund and cancel command into one
There is not really a reason for the end user to disguise between the two. They must always be executed together anyway.

delta1:
Merges the `cancel`/`refund` commands in the swap cli into one `cancel-and-refund` command

closes #1020


swap/src/cli/cancel_and_refund.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::sync::Arc;
+use uuid::Uuid;
+
+pub async fn cancel_and_refund(
+    swap_id: Uuid,
+    bitcoin_wallet: Arc<Wallet>,
+    db: Arc<dyn Database>,
+) -> Result<BobState> {
+    match cancel(swap_id, bitcoin_wallet.clone(), db.clone()).await {
+        Ok((_, subscription, _)) => subscription.wait_until_final().await?,
```
binarybaron, line 16: Do we need to wait for the cancel transaction to be confirmed? I don't really see any reason to do so.

delta1: Yes the refund and punish transactions spend the output of the cancel transaction, so I believe it would at least have to be confirmed. 

“ Alice constructs and shares with Both an adaptor signature on Tx_refund spending from the output of Tx_cancel to address_b.” 

https://comit.network/blog/2020/10/06/monero-bitcoin/

binarybaron: But if I'm not mistaken, Bitcoin allows you to publish transactions that spend unconfirmed outputs. They (Refund tx) just won't be confirmed until their parent (Cancel Tx) has been included in a block.

Not waiting for the Cancel Tx to be confirmed allows both TxCancel and TxRefund to be included in the same block.

delta1: Done

swap/src/cli/cancel_and_refund.rs
```diff
@@ -0,0 +1,115 @@
...
+        ),
+    };
+
+    tracing::info!(%swap_id, "Manually cancelling swap");
+
+    let (txid, subscription) = match state6.submit_tx_cancel(bitcoin_wallet.as_ref()).await {
+        Ok(txid) => txid,
+        Err(err) => {
+            if let Ok(error_code) = parse_rpc_error_code(&err) {
+                tracing::debug!(%error_code);
```
binarybaron, line 62: Can we add a message here or remove it entirely?

swap/src/cli/cancel_and_refund.rs
```diff
@@ -0,0 +1,115 @@
...
+use uuid::Uuid;
+
+pub async fn cancel_and_refund(
+    swap_id: Uuid,
+    bitcoin_wallet: Arc<Wallet>,
+    db: Arc<dyn Database>,
+) -> Result<BobState> {
+    if let Err(err) = cancel(swap_id, bitcoin_wallet.clone(), db.clone()).await {
+        tracing::info!(%err, "Could not submit cancel transaction");
+    };
```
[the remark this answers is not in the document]
delta1: It will log a message and then submit the refund tx. It doesn’t bail out.

swap/src/cli/cancel_and_refund.rs
delta1: It's for debugging so I'll add a message.

binarybaron (approved):

Issue #1273, opened 2023-01-10
`asb` wallet inaccessible after refreshing it with `monero-wallet-cli`


icy-ux:
**Describe the bug**
After accessing the `asb` wallet (`asb-wallet`) using `monero-wallet-cli`, `asb` is no longer able to access the wallet. This persists even after `monero-wallet-rpc` has been restarted, and after the entire machine has been restarted.

As part of troubleshooting #1061 I refreshed the wallet with `monero-wallet-cli`. This worked, but now I can't run `asb` anymore.

**Lost/trapped Funds**
No

**Debug logs**
```
2023-01-09T13:59:57.426024255Z DEBUG Opening Monero wallet
Error: Unable to create Monero wallet, please ensure that the monero-wallet-rpc is available

Caused by:
    0: JSON-RPC request failed with code -21: Cannot create wallet. Already exists.
    1: JSON-RPC request failed with code -21: Cannot create wallet. Already exists.
```

**Platform:**

- ASB
- Software Version `asb 0.12.0`
- Operating System: Debian 11.6

delta1:
Thanks for the report. When you open the wallet with the monero-wallet-cli did it make you set a password?

icy-ux:
On Tue, Jan 10, 2023 at 02:50:19AM -0800, Byron Hambly wrote:
>Thanks for the report. When you open the wallet with the monero-wallet-cli did it make you set a password?

No it did not, at least not as far as I can recall. Since I restarted the server I don't have terminal logs anymore. I do know it asked for a password and I successfully used a blank password ("just press enter").

delta1:
Thanks yeah confirmed that it's an empty password. Busy investigating further, trying to reproduce. Any logs from `monero-wallet-rpc`?

icy-ux:
Nothing in `journalctl -u monero-wallet-rpc`. Should I look somewhere else?

icy-ux:
Found a `monero-wallet-rpc.log`, which contained the following relevant if unhelpful entry:

```
2023-01-10 11:08:03.427 [RPC0]  ERROR   wallet.wallet2  src/wallet/wallet2.cpp:4744     boost::filesystem::exists(m_wallet_file, ignored_ec). THROW EXCEPTION: error::file_exists
[...]

PR #1275, opened 2023-01-10
ci: update msrv and github actions toolchains

delta1:
The change in github actions to use dtolnay/rust-toolchain action was defaulting to the latest stable and was not overridden by the rust-toolchain.toml file. 

This updates the MSRV and relevant CI.


delta1:
#1273

binarybaron:
bors retry

PR #1278, opened 2023-01-11
upgrade base64, vergen, bdk

delta1:
Upgrades dependencies, base64 had a breaking change but is only used in a test 

- #1271
- #1266 
- #1264


delta1:
bors retry

delta1:
bors retry

PR #1231, opened 2022-12-03
 feat(asb): allow config overrides from env vars

delta1:
- upgrades config crate to 0.13.2 #1087
- adds environment source for config overrides 

This change allows the ASB config to be overridden by env vars starting with `ASB__`. 
Adds a test to check this, and updates the de/serialization of the network `listen` and `external_addresses` fields to be a comma separated string from the environment. The `separator` is a pair of underscores, which is necessary since some of the fields already have underscores in their names. 

Example env var: `ASB__NETWORK__EXTERNAL_ADDRESSES="/dns4/example.org/tcp/9939,/ip4/1.2.3.4/tcp/9940"` 

This particular change will help when running the ASB as a docker container (wip), allowing the config to be set in the env for docker-compose etc.

Issue #633, opened 2021-07-14
[1 XMR] Allow ASB to register with multiple rendezvous points


thomaseizinger:
At the moment, an ASB can only register with a single rendezvous point.

It would be advantageous if a single ASB instance could register with multiple rendezvous points simultaneously. There are three main areas that need to be changed:

## 1. The ASB's rendezvous `NetworkBehaviour`

The trickiest part about support multiple rendezvous points is to keep track of the registrations and renew them as they expire. [This](https://github.com/comit-network/xmr-btc-swap/blob/52d5f2d83bdd584d734d671198878b4670325d3a/swap/src/asb/network.rs#L173-L343) is where the current magic happens.

The module is reasonably-well supported by tests so I would suggest to start with a test that asserts that we can register with multiple rendezvous points and renew each registration as it expires.

The API changes of the module should be minimal and the rest of the application should still be able to run.

## 2. The configuration

Once the `NetworkBehaviour` supports multiple rendezvous points, extending the [configuration](https://github.com/comit-network/xmr-btc-swap/blob/master/swap/src/asb/config.rs) shouldn't be too hard. We already support a list of multiaddresses for the `listen` key of `network`. Just need to do the same thing for `rendezvous_point`.

## 3. The initial setup wizard

There are multiple ways of how to deal with the initial setup wizard:

1. Either only ask the user once for a rendezvous point and insert that into the list
2. Ask the user in a loop for rendezvous points until they enter an empty address

---

As with all changes, this should definitely come with a changelog entry.

protomens:
This would be advantageous for providers and swappers alike. We started our own Rendezvous Endpoint last night and tried to add a second one with similar array syntax for multiaddress as ["", ""]
but obviously was a no go.

thomaseizinger:
We currently don't have to bandwidth to implement this but if you need any more instructions on how to do get started, let me know!

protomens:
> We currently don't have to bandwidth to implement this but if you would like to tackle it and need any more instructions on how to do get started, let me know!

I manage a couple servers that I would be able try this on. First things first, I have to familiarize myself with Rust. I'm versed in several other languages and sat down with the Rust book this afternoon to get started.

I took a look at the code first and it appears to me that the registration to a rendezvous endpoint happens in:

```
swap/src/asb/network.rs
```

Correct me if I'm wrong.

[...]

PR #1417, opened 2023-08-01
feat(asb): allow asb to register with mulitple rendezvous nodes

delta1:
implements #633  

This PR updates the ASB to be able to register with multiple rendezvous nodes. A unit test is added that checks this behaviour. I also tested it manually overnight and it repeatedly reregistered with the specified nodes.  

The config file option is left as `rendezvous_point` for backwards compatibility, but now uses the same deserialization as introduced in #1231 so that multiple addresses can be specified in a comma separated string, or as a toml array of strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Minimum Supported Rust Version (MSRV) bumped to 1.63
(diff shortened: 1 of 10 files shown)


delta1:
@binarybaron wanna give this a spin?

binarybaron:
Will do 👍

delta1:
awesome, good catch I’ll update the docs and squash these commits

binarybaron:
thanks for the work! @delta1

bors r+

PR #1420, opened 2023-08-04
Add Tor SOCKS option to ASB for registering to Tor Rendezvous servers and for Electrum client

ghost:
Allows user to specify the Tor SOCKS5 port exactly like the cli equivalent command so that registration over Tor is possible for clearnet or onion Rendezvous servers.

Example Tor Rendezvous server: `/onion3/yn5wziq6coomwvk2regzvwmjc2hvbw3z2w35bdfersr3kdqat6jwh6ad:8888/p2p/12D3KooWDgGZKJHbHfaUFMoQHj4zEA8xcXW7tdw7rtQDMU7zoXVE`


pokkst:
I just realized the Tor client stuff is right above where the swarm is created, which I suppose could be used instead, though the port option during launch is much easier than configuring the Tor control port and permissions and whatnot.

delta1:
hey thanks for the PR, i’ll only get a chance to review during the week

pokkst:
Sounds good. I've cleaned it up a bit so that it uses the port option from the config file, and also made it so when Tor is running the Electrum client will also use the Tor proxy, and made it so the ElectrumBlockchain client takes into account the config.

I have my own fork that also removes the Kraken price websocket in favor of the Feather price websocket, with Tor support for that as well, but I don't think it's as stable. I'm not sure whether it's on my end or Feather's end, but it keeps disconnecting and having to back off and try again.

swap/src/bin/asb.rs
```diff
@@ -100,6 +100,25 @@ async fn main() -> Result<()> {
...
+            Some(ac)
+        }
+        Err(_) => {
+            tracing::warn!("Tor not found. Running on clear net");
+            None
+        }
+    };
+    let tor_port = if _ac.is_some() { config.tor.socks5_port } else { 0u16 };
+    let proxy_string = if tor_port != 0u16 { format!("127.0.0.1:{}", tor_port) } else { "".to_string() };
+    tracing::info!(%proxy_string, "SOCKS5");
```
binarybaron, line 120: Please add a proper message here. SOCKS5 is ambiguous

pokkst:
Any chance for a re-review?

swap/src/bin/asb.rs
```diff
@@ -140,29 +170,13 @@ async fn main() -> Result<()> {
                 }
             }
 
-            let bitcoin_wallet = init_bitcoin_wallet(&config, &seed, env_config).await?;
+            let bitcoin_wallet =
+                init_bitcoin_wallet(&config, &seed, env_config, proxy_string).await?;
```
delta1, line 174: if the env_config has the tor port then this additional argument is not necessary

swap/src/bin/asb.rs
```diff
@@ -100,6 +100,36 @@ async fn main() -> Result<()> {
...
+        format!("127.0.0.1:{}", tor_port)
+    } else {
+        "".to_string()
+    };
+    if proxy_string.is_empty() {
+        tracing::info!(%proxy_string, "Not using SOCKS5 proxy");
+    } else {
+        tracing::info!(%proxy_string, "Using SOCKS5 proxy at");
+    }
+
```
delta1, line 132: this can be cleaned up into 1 match statement that returns an Option for the tor port, and do the logging there too

delta1 (changes requested):
looks good, see if you agree with my feedback and take a stab at it

delta1:
pushed a commit to fix the fmt. there is a test compilation error

pokkst:
Sounds good. Will work on it this weekend!

PR #1422, opened 2023-08-08
ci: add rustfmt to draft-new-release

delta1:
fixing the draft new release action, churn in dprint means that now rustfmt is run separately so it needs to be installed at this step 

previous failure: https://github.com/comit-network/xmr-btc-swap/actions/runs/5787093337/job/15683238703


delta1:
bors merge

PR #1426, opened 2023-08-13
Fix syncing issues with ASB XMR wallet during startup

ghost:
I have been encountering an issue where the monero wallet for the ASB will properly sync upon startup, and can perform its usual tasks, but it does not save this state. This means that when restarting the ASB program, the balance will be that of from BEFORE the last startup and sync, sometimes going back until it was first created, essentially erasing all progress.

This adds https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#store as a method for the RPC client, allowing the ASB program to save the wallet's state periodically.


binarybaron:
Thanks for the PR. Seems sensible to me. Please fix the clippy errors (see the failed test) before we can get this merged.

swap/src/protocol/alice/swap.rs
```diff
@@ -45,6 +45,8 @@ where
             .await?;
     }
 
+    swap.monero_wallet.store().await?;
```
binarybaron, line 48: Quite arbitrary to store the wallet at this point. What if the ASB does not complete a swap for a whole week? Then the wallet will never be stored

binarybaron (changes requested):

swap/src/protocol/alice/swap.rs
binarybaron: I think we should do this periodically. Scanning blocks, regardless of whether or not they contain any transactions that affect our balance, takes time and storing the wallet skips scanning those empty blocks for the next time. 


I'm quite confused why the monero-wallet-rpc doesn't do this itself. I'd expect the wallet to be stored periodically...

binarybaron: Quote from a Monero core dev

<moneromooo> It saves at (non fatal) exit. Save it yourself when you know it's important. Saving can take some time.

delta1: Yeah periodically sounds good then

binarybaron:
Should we do this for the CLI as well? I'm not sure if the "swap-monitoring-wallet is recreated at each startup

pokkst:
Any chance for a re-review? Thanks

binarybaron:
Ne and delta1 both think we should save the wallet periodically instead of at certain states. (See the previous review)

ikmckenz:
@binarybaron how should periodic saving be scheduled? On a regular cadence (every X minutes)? Or after triggered after events like transactions? Both?

PR #1428, opened 2023-08-13
Fix issue where Windows cannot open ASB wallet to create lock XMR transaction

pokkst:
It seems on Windows that opening an already opened wallet results in an error/crash. This does not seem to be the case for Linux and macOS. Removing this line fixed the issue on Windows, and had no regressions on macOS/Linux from my testing.


delta1 (approved):
Thanks for the PR @pokkst !

binarybaron:
I don't really understand why we can just remote that open_wallet call

delta1:
> I don't really understand why we can just remove that open_wallet call

`open_wallet` has already been called in `Wallet::open_or_create` by the time it gets here

Issue #835, opened 2021-10-21
`monero-wallet-rpc` stays alive after cli is killed


binarybaron:
**Describe the bug**
While working on a GUI for the swap-cli, I ran into a bug. When the cli is running in a non-interactive shell and the process is terminated by sending a SIGINT signal, the monero-wallet-rpc stays alive. This causes an issue where the user cannot initiate new swaps because two instances of monero-wallet-rpc are trying to access the same wallet, resulting in an error. I currently handle this by manually terminating monero-wallet-rpc, but I believe this should be handled by the cli itself.

**How to reproduce**
Start swap and detach from terminal
```
./swap --json --debug --testnet buy-xmr --change-address tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx --receive-address 59McWTPGc745SRWrSMoh8oTjoXoQq6sPUgKZ66dQWXuKFQ2q19h9gvhJNZcFTizcnT12r63NFgHiGd6gBCjabzmzHAMoyD6 --seller /dnsaddr/unstoppableswap.net/p2p/12D3KooWLHbHtkK53WoBA82Yht3PCCx6KXYxoJ6QqgXg1rSNtWhR &
```

```
ps aux | grep buy-xmr
 7112 ttys001    0:00.76 ./swap --json --debug --testnet buy-xmr --change-address tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx --receive-address 59McWTPGc745SRWrSMoh8oTjoXoQq6sPUgKZ66dQWXuKFQ2q19h9gvhJNZcFTizcnT12r63NFgHiGd6gBCjabzmzHAMoyD6 --seller /dnsaddr/unstoppableswap.net/p2p/12D3KooWLHbHtkK53WoBA82Yht3PCCx6KXYxoJ6QqgXg1rSNtWhR

ps aux | grep monero-wallet-rpc
7113   0,0  0,3  4538508  11204 s001  SN    2:31pm   0:04.95 /Users/****/Library/Application Support/xmr-btc-swap/cli/testnet/monero/monero-wallet-rpc --stagenet --daemon-address stagenet.melo.tools:38081 --rpc-bind-port 53238 --disable-rpc-login --wallet-dir /Users/*****/Library/Application Support/xmr-btc-swap/cli/testnet/monero/monero-data
```

Send `SIGINT` to cli
```
kill -SIGINT 7112
[...]

PR #1434, opened 2023-08-30
Kill monero-wallet-rpc child process when receiving SIGINT, SIGTERM o…

binarybaron:
…r SIGHUP

When the CLI is terminated via a signal, the child process monero-wallet-rpc is not automatically terminated. This discrepancy arises from the behavior of terminal-initiated interrupts (e.g., STRG-C), which send the termination signal to all processes in the process group, effectively killing the child process.

To address this, we implement a signal handler that explicitly terminates the monero-wallet-rpc child process upon receiving a termination signal for the parent process. This ensures that the child process does not continue running in the background, detached from the parent. Failing to terminate the child process would lock the wallet, preventing future instances of the CLI from starting.

Fixes https://github.com/comit-network/xmr-btc-swap/issues/835

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,13 +103,13 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: swap-${{ matrix.target }}
-          path: target/${{ matrix.target }}/debug/swap
+          path: target/${{ matrix.target }}/debug/swap*
 
       - name: Upload asb binary
         uses: actions/upload-artifact@v3
         with:
           name: asb-${{ matrix.target }}
-          path: target/${{ matrix.target }}/debug/asb
+          path: target/${{ matrix.target }}/debug/asb*
 
   test:
     strategy:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


binarybaron:
This fixes this https://github.com/comit-network/xmr-btc-swap/issues/835

delta1:
seems to break the windows build, is there a cross platform way to do the same?

binarybaron:
I'll look into it

binarybaron:
I made it platform specific but we still need to test this on Windows. I'll do it in the next few days

binarybaron:
Our current Tokyo version doesn't have support for signals on windows... We'll have to upgrade

binarybaron:
@delta1 do you know why the windows binaries arent getting attached as artifacts?

I only have a windows VM and struggling to compile on my machine.

delta1:
oh right, it’s missing the exe extensions on the upload artifact step. i had a fix for this somewhere, let me try find it or redo it.

binarybaron:
Also doesn't work on tokio 1.21.0

delta1:
> Ok, this does not work on windows due to the tokio version upgrade... Swap terminates with
> 
> ```
> thread 'main' has overflowed its stack
> ```

does this happen immediately or only when it catches the signal?

binarybaron:
Immediately. It's definitely due to the Tokio upgrade. Upgrading Tokio on master without the other changes also caused a crash.

It also crashes on commands like `balance` where the monero wallet is not initialized. Before crashing the tracing initialization log message is printed which means that it doesn't overflow immediately on startup.

binarybaron:
> Immediately. It's definitely due to the Tokio upgrade. Upgrading Tokio on master without the other changes also caused a crash.
> 
> It also crashes on commands like `balance` where the monero wallet is not initialized. Before crashing the tracing initialization log message is printed which means that it doesn't overflow immediately on startup.
> 
> Maybe we really need to increase the stack size but I wonder why.

Update: This is not related to the tokio upgrade. All debug releases on windows currently panic due to a stackoverflow which means that this PR does not break anything.
However, I'm not sure yet if this PR achieves the desired behaviour on windows. I'll have to look into that.

binarybaron:
Closing this for now...

BrandyJSon:
@binarybaron did you ever find if this could be merged without breaking windows builds. Makes spawning the swap tool programmatically quite annoying to clean up.

binarybaron:
No I haven't but feel free to open a PR. I'll happily review and merge it if it works on all platforms (or at least doesn't break any platforms)

PR #1436, opened 2023-08-30
Refresh monero monitoring wallet at startup to ensure daemon is is av…

binarybaron:
…ailable

### DIFF
--- a/swap/src/bin/swap.rs
+++ b/swap/src/bin/swap.rs
@@ -541,6 +541,11 @@ async fn init_monero_wallet(
     )
     .await?;
 
+    monero_wallet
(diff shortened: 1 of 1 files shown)


delta1 (approved):
bors r+

binarybaron:
We have to revert this. I don't know if this takes forever if we have a lots of blocks to scan (we haven't used the CLI in a while) and I think we can make this more efficient as the wallet doesn't contain any transactions (99% sure) and is only used to watch the tx transfer proof

delta1:
bors retry

delta1:
bors retry

Issue #1432, opened 2023-08-24
Issues with redeeming the Monero


binarybaron:
**Describe the bug**
A bunch of users have reported an issue where the CLI fails to redeem the Monero.

**Lost/trapped Funds**
Yes, but recoverable using the monero-recover command.

**Debug logs**
https://github.com/UnstoppableSwap/unstoppableswap-gui/issues/147

https://github.com/UnstoppableSwap/unstoppableswap-gui/issues/155

**Platform (please complete the following information):**

- CLI
- Software Version: 0.12.0

binarybaron:
This seems to be very prevalent. I've gotten more and more users contacting me with this issue. I'll start working on it and try to find the root cause.

binarybaron:
Logs of user `gad35...`


```
{"timestamp":"2023-04-28T02:23:46.368003Z","level":"DEBUG","fields":{"message":"Advancing state","state":"btc is redeemed"}}
{"timestamp":"2023-04-28T02:23:46.370827Z","level":"WARN","fields":{"message":"Failed to generate monero wallet from keys: Failed to generate new wallet from keys: JSON-RPC request failed with code -1: Wallet already exists.: JSON-RPC request failed with code -1: Wallet already exists."}}
{"timestamp":"2023-04-28T02:23:46.370856Z","level":"INFO","fields":{"message":"Falling back to trying to open the the wallet if it already exists","wallet_file_name":"5b352bf0-3de6-4d70-a761-27bc48c97b9d"}}
{"timestamp":"2023-04-28T02:23:46.893715Z","level":"INFO","fields":{"message":"Connected to Alice at /ip4/45.138.50.29/tcp/9939/p2p/12D3KooWMgGjeW7ErQxCQzaeHiXxJn42wegCPFepixEXfBJT1PNS"}}
{"timestamp":"2023-04-28T02:57:07.400481Z","level":"INFO","fields":{"message":"Logging initialized to C:\Users\Brady\AppData\Roaming\xmr-btc-swap\cli\mainnet\logs"}}
{"timestamp":"2023-04-28T02:57:07.400652Z","level":"DEBUG","fields":{"message":"Using existing sqlite database."}}
{"timestamp":"2023-04-28T02:57:07.402302Z","level":"DEBUG","fields":{"message":"Reading in seed from C:\Users\Brady\AppData\Roaming\xmr-btc-swap\cli\mainnet\seed.pem"}}
[...]

PR #1437, opened 2023-08-31
Change xmr node moneroworld

binarybaron:


delta1:
Motivation for the change?

binarybaron:
I'm looking into this issue https://github.com/comit-network/xmr-btc-swap/issues/1432

I have been able to reproduce it myself simply by trying to conduct a mainnet swap. One of the error we encounter (thrown by the monero-wallet-rpc) is `no connection to daemon` which leads me to believe there might be an issue with the monero daemon we are connecting to. Resuming the swap with a different daemon specified did fix the issue for me.

However the node does seem to be online as can be seen: https://community.rino.io/nodes.html

Don't merge this yet please. Maybe this is caused by something else.

binarybaron:
Closed in favour of https://github.com/comit-network/xmr-btc-swap/pull/1441

PR #1441, opened 2023-09-02
Dynamically choose monero daemon to connect to

binarybaron:
The GUI now does this with https://github.com/UnstoppableSwap/unstoppableswap-gui/pull/157. However, ideally, this should be handled by swap itself and not by a third party wrapper which is why this PR implements a similar behaviour for  swap. On startup we loop through list of public Monero daemons and choose one dynamically based on their response to the get_info RPC call.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


binarybaron:
@CodiumAI-Agent /review

delta1:
Nice! Yeah it does look good to me, maybe some minor things that I'd do differently, but certainly not showstoppers.

delta1 (approved):
bors r+

delta1:
oh @binarybaron I forgot, this definitely needs a changelog entry!

binarybaron:
> oh @binarybaron I forgot, this definitely needs a changelog entry! 

yes I just noticed as well. I'd we just do a new PR for changelog

PR #1447, opened 2023-09-29
Update lock timeout to 10 min

pokkst:
This updates the lock timeout from 3 min to 10 min.

Over Tor, it is sometimes very slow to setup a swap, completely missing the 3 minute window. The BTC will still lock and broadcast, but the ASB will have already stopped watching for the lock tx, thus will not lock the XMR once the BTC lock tx confirms. This mitigates that issue.

### DIFF
--- a/swap/src/env.rs
+++ b/swap/src/env.rs
@@ -46,7 +46,7 @@ pub struct Regtest;
 impl GetConfig for Mainnet {
     fn get_config() -> Config {
         Config {
-            bitcoin_lock_mempool_timeout: 3.std_minutes(),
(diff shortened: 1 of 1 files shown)


pokkst:
It's recommended, if merged, to notify the current mainnet ASBs that appear on the UnstoppableSwap Rendezvous server to use this new 10 minute timeout change.

binarybaron:
Lgtm. This can be changed by the ASB itself as they please as it's not dependent on the two peers agreeing on a timeout. It's enforced by the ASB

PR #1450, opened 2023-10-23
build(deps): bump pem to 3.0

delta1:
replaces #1415

also fixes the `seed_from_pem_fails_for_long_seed` unit test, which was "passing" but not actually testing what it meant to.


delta1 (commented):
bors r+

PR #1456, opened 2023-12-05
ci: update minimum rust version to 1.70

delta1:
this is to fix the issue with installing sqlx-cli in ci


delta1:
bors merge

delta1:
looks like bors has gone away @binarybaron

Issue #1459, opened 2023-12-09
Build error[E0282]: type annotations needed


fullmetalScience:
With *master* at commit db400ad4c4bf4c3f5e252966ff4fda8614242c2e:

```
$ cargo build
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
   Compiling libgit2-sys v0.14.2+1.5.1
   Compiling libsqlite3-sys v0.24.2
   Compiling secp256kfun v0.7.1 (https://github.com/LLFourn/secp256kfun#9657d8c1)
   Compiling electrum-client v0.12.0
error[E0282]: type annotations needed for `point::Point<point_type::EvenY, S>`
   --> ~/.cargo/git/checkouts/secp256kfun-fb74af16957b2754/9657d8c/secp256kfun/src/point.rs:496:45
    |
496 |         Point::from_xonly_bytes(bytes).map(|p| p.set_secrecy::<S>())
    |                                             ^
    |
help: consider giving this closure parameter an explicit type, where the type for type parameter `S` is specified
    |
496 |         Point::from_xonly_bytes(bytes).map(|p: point::Point<point_type::EvenY, S>| p.set_secrecy::<S>())
    |                                              ++++++++++++++++++++++++++++++++++++

   Compiling git2 v0.16.1
   Compiling sqlx-core v0.6.3
For more information about this error, try `rustc --explain E0282`.
error: could not compile `secp256kfun` (lib) due to previous error
```

delta1:
rust version?

fullmetalScience:
`rustc 1.74.0`

delta1:
thanks will take a look. in the meanwhile 1.70 should work (it does in the CI)

ikmckenz:
Also broken on 1.72

delta1:
resolved short term by using rust 1.70 and long term by #1466

PR #1466, opened 2023-12-11
upgrade secp256kfun

delta1:
re: #1459  

upgrades the secp256kfun dependency, the previously pinned version fails to compile on rust 1.74 

fixes a few resulting issues, and some minor cleanup refactoring 

also adds a new CI job to run cargo check on rust stable 

also closes: 
- #1410 
- #1412


binarybaron:
I will need some time to review this as this is quite critical code and I'm not too familiar with the library.

ikmckenz:
Out of curiosity, why resolve on 0.24.1 instead of the most recent release (0.28)?

delta1:
> Out of curiosity, why resolve on 0.24.1 instead of the most recent release (0.28)?

looks like you’re referring to secp256k1, their PR is upgrading secp256kfun. as for the versions, from the lockfile it looks like bitset still has a dependency on 0.24 of secp256k1, and secp256kfun is still depending on 0.27 of secp256k1

swap/src/bitcoin.rs
```diff
@@ -206,20 +206,21 @@ pub fn verify_encsig(
...
-    let miniscript = MINISCRIPT_TEMPLATE.replace('A', &A).replace('B', &B);
+    let miniscript = MINISCRIPT_TEMPLATE
+        .replace('A', &A.to_string())
+        .replace('B', &B.to_string());
 
     let miniscript =
         bdk::miniscript::Miniscript::<bitcoin::PublicKey, Segwitv0>::from_str(&miniscript)
             .expect("a valid miniscript");
 
-    Descriptor::Wsh(Wsh::new(miniscript).expect("a valid descriptor"))
```
binarybaron, line 222: why are we removing this expect(...) here?

binarybaron:
LGTM but have you completed a swap using the new CLI version and an old ASB version yet?

delta1:
updated to include #1546 #1547

delta1:
updated to include #1549 #1550

delta1:
update to include #1556 and #1557 

also sets the revision in the Cargo.toml manifest

binarybaron:
Is this 100% backwards compatible in terms of the database and p2p protocols?

delta1:
feel free to test it

delta1:
@ikmckenz if you have a moment, please try test build this branch asb/swap, and gets those binaries on testnet performing a swap against the previous version

delta1:
Sigh I don't have any testnet bitcoin to test with :weary: @binarybaron any donations? 

`tb1qttry3wfcwcf7gyv26a5elt9hcran7nsnfj9zgj`

delta1:
> Is this 100% backwards compatible in terms of the database and p2p protocols?

tested a number of testnet swaps between this version of asb and latest swap release, as well as this version of swap with latest asb release 

happy for this change to be a version change to 0.13

linsui:
Why do you use the git repo instead of 0.10.0 on crates.io?

delta1:
> Why do you use the git repo instead of 0.10.0 on crates.io?

good question, i think that’s a hangover from where we were ahead of their releases to crates.io 

i’m going to rework this PR to use the crates version, thanks for the suggestion

binarybaron (approved):
LGTM!

delta1:
includes a fix for #1645

PR #1487, opened 2023-12-14
Monero wallet refresh fix

binarybaron:
This PR changes the following behaviour in the refresh functionality of the monero wallet
- Allows for multiple retries because in some cases users have experienced an issue where the wallet rpc returns `no connection to daemon` even though the daemon is available. I'm not 100% sure why this happens but retrying often fixes the issue
- Attempt to print the current sync height while the wallet is syncing. This only works to some degree because the `monero-wallet-rpc` stops responding (or takes a long time to respond) while it's refreshing
- The `monero-wallet-rpc` is started with the `--no-initial-sync` flag which ensures that as soon as it's started, it's ready to respond to requests
- The `monero-wallet-rpc` was upgraded to `v0.18.3.1` because this PR https://github.com/monero-project/monero/pull/8941 has improved some of the issues mentioned above

This PR is part of a larger effort to fix this issue https://github.com/comit-network/xmr-btc-swap/issues/1432


binarybaron:
@delta1 This is ready for review

delta1 (approved):
Nice LGTM. Do you want to add a note to the changelog? At least for bumping the Monero version

delta1:
seems like some tests are now running into the 6 hour CI time limit?

delta1:
@binarybaron a few of the ci jobs are not working with this change

binarybaron:
@delta1 Okay it seems like some of the tests will time out. Looking at the logs it seems that the CI is stuck at refreshing the Monero wallet which seems related to this PR. I have no idea why.

When refreshing the CI calls this
https://github.com/comit-network/xmr-btc-swap/blob/d8dacbdee99a8d951977e62db1c888cc9ed56ea8/swap/src/monero/wallet.rs#L265
function with `max_attempts=1 `. 

This line
https://github.com/comit-network/xmr-btc-swap/blob/d8dacbdee99a8d951977e62db1c888cc9ed56ea8/swap/src/monero/wallet.rs#L299
is never reached. Nowhere in the logs is this log message to be found. The other tokio task which is responsible for logging the sync progress is called and prints periodic log messages.

Do you have any idea what the issue could be here? I'm clueless. Here is the entire function for reference:

https://github.com/comit-network/xmr-btc-swap/blob/d8dacbdee99a8d951977e62db1c888cc9ed56ea8/swap/src/monero/wallet.rs#L265-L330

delta1:
revert in #1593 so we can recreate this while debugging

PR #1543, opened 2024-02-03
Upgrade to vergen 8.3, move to git describe --dirty --tags for semver

ikmckenz:
Vergen is used to to put git tag version information into the binary, specifically for when running the binaries with `--help` you can see what version you are running. The current master branch shows `preview-18-g4a758f7`, for example.  

As far as I can tell Vergen version 8 got rid of the VERGEN_GIT_SEMVER_LIGHTWEIGHT environment variable that is currently used in the code (along with changing the format of the build.rs file), and instead we should switch to VERGEN_GIT_DESCRIBE. I didn't find this explicitly documented anywhere, but we can see how in vergen v7 they had [VERGEN_GIT_SEMVER and VERGEN_GIT_SEMVER_LIGHTWEIGHT](https://github.com/rustyhorde/vergen/blob/legacy/v7/src/constants.rs#L26-L27) defined, and in the current code there is only [VERGEN_GIT_DESCRIBE, but the comment for git describe still says VERGEN_GIT_SEMVER](https://github.com/rustyhorde/vergen/blob/master/vergen-lib/src/keys.rs#L86-L88). 

Currently this PR sets both `--dirty` and `--tags` for the git describe command, and specifically `--dirty` is a slight change from existing behavior that is only relevant for developers. 

The current behavior is the equivalent of only setting `--tags` without setting `--dirty`. If there are uncommitted changes to the code, it will still display the previous tag with no modifications. 

The new behavior with the `--dirty` flag set is that if there are uncommitted changes, it will add `-dirty` to the end of the version string. So for example applying the changes in this pull request without committing them shows the version as `preview-18-g4a758f7-dirty`, with the `-dirty` showing there are uncommitted changes. Only developers making changes to the code should ever see this `-dirty` extension. Then committing the changes shows a new commit version as expected: `preview-19-ga74d448`. I think this is a small improvement, in case a developer ever runs into weird issues because of changing a file they didn't intend to, it could help with debugging that.


delta1 (approved):
LGTM pending CI 

Thanks @ikmckenz

delta1:
i think you just need cargo fmt

delta1:
closes #1531

PR #1552, opened 2024-02-11
Upgrade testcontainers from 0.12 to 0.14

ikmckenz:
closes #1038

### DIFF
--- a/monero-harness/src/image.rs
+++ b/monero-harness/src/image.rs
@@ -11,11 +11,8 @@ pub const MONEROD_DEFAULT_NETWORK: &str = "monero_network";
 /// this doesn't matter.
 pub const RPC_PORT: u16 = 18081;
 
-#[derive(Debug, Default)]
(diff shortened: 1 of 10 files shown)


monero-harness/src/image.rs
```diff
@@ -200,21 +168,18 @@ pub struct MoneroWalletRpcArgs {
     pub daemon_address: String,
 }
 
-impl Default for MoneroWalletRpcArgs {
-    fn default() -> Self {
-        unimplemented!("A default instance for `MoneroWalletRpc` doesn't make sense because we always need to connect to a node.")
-    }
-}
-
 impl MoneroWalletRpcArgs {
```
ikmckenz, line 171: Didn't remove the debug statements or otherwise touch this one, I think that with the semi colon it doesn't return temp... Does the new() do anything in this case?

delta1:
Thanks!

PR #1276, opened 2023-01-11
RPC server for API Interface 

yamabiiko:
Opening this PR so it can start getting reviewed and possibly get some help.
As discussed in #1185, this is quite a substantial rework which adds the possibility of running an RPC server and unifies the RPC methods and CLI commands under an internal API.

To be done before we can get this merged:
- [x] Tests: fixing and expanding tests in `swap/tests/rpc.rs`: right now the test runs the RPC server but the functionality cannot be tested as the Bitcoin wallet does not have any balance and the swap database is empty (so right now it's using some placeholder BTC/XMR addresses and swap ids and it will always fail).
- [x] Tests: write more tests (containers?)
- [x] Reevaluate the placing of the new files (request.rs, api.rs, rpc.rs) within the project file strucuture
- [ ] Organize commits by squashing some, changing commit messages to be more descriptive to make reviewing this easier
- [ ] Get the entire code reviewed by one other person (@delta1)

If someone could help with writing proper tests (or pointing out at some examples in the code to how swaps and transactions can be simulated/tested) it would be appreciated.

PR #1555, opened 2024-02-14
ci: free up space on ubuntu test job

delta1:
the ubuntu test job has sometimes been failing with an out of space error. 

this change worked on #1276 so i'm trying it here


delta1:
adds 10 minutes to the job...

Issue #674, opened 2021-08-22
Swap failed, failed to broadcast Bitcoin cancel transaction


TA1002:
Rather than try to explain the issue, I decided to just copy/paste my terminal:

user@host:~/swaptool$ ./swap history
+--------------------------------------+----------------------------+
| SWAP ID                              | STATE                      |
+===================================================================+
| 7e9c4b62-1411-46be-b06e-77511f06969b | Cancel timelock is expired |
+--------------------------------------+----------------------------+
user@host:~/swaptool$ ./swap resume --swap-id 7e9c4b62-1411-46be-b06e-77511f06969b
Connected to Alice at /ip4/51.15.44.142/tcp/9939/p2p/12D3KooWAPci3x4GuWDi9P86efnR1q9a7phKnPwnwQcRNMW2HDUy
Error: Failed to broadcast Bitcoin cancel transaction 0a03a88c6eaf61defc83943b874ecec5bba0884338fb0774f97808bd78f86117

Caused by:
    Electrum(Protocol(String("sendrawtransaction RPC error: {\"code\":-26,\"message\":\"non-BIP68-final\"}")))

thomaseizinger:
Can you manually refund the swap?

TA1002:
How would I accomplish this?

I guess I should explain more in case there's a misunderstanding. I was trying to execute a swap. I had already sent the BTC to the provided address. While I was at the stage "Waiting for Alice to lock Monero", I lost connection. Eventually I figured out how to check the history of the swap and was able to "resume" with the correct swap ID. Again I was at the stage "Waiting for Alice to lock Monero". After several minutes, I received a message indicating this had timed out and that the swap failed. 

I interpreted the messages above in my original post that the BTC would be returned to my refund address, but that had also failed. Is that correct? 

Sorry if there's any misunderstanding on my part.

TA1002:
Update: I noticed the "state" changed from "cancel timelock is expired" to "Bitcoin refundable"

So I attempted to use the refund command. But I still received an error:

user@host:~/swaptool$ ./swap history
+--------------------------------------+--------------------+
| SWAP ID                              | STATE              |
+===========================================================+
| 7e9c4b62-1411-46be-b06e-77511f06969b | Bitcoin refundable |
+--------------------------------------+--------------------+
user@host:~/swaptool$ ./swap refund --swap-id 7e9c4b62-1411-46be-b06e-77511f06969b
[...]

PR #1571, opened 2024-03-06
Improve usability of refund

### LINKED ISSUE #683: [1.0 XMR] Provide better feedback when cancel/refund transaction publish fails by parsing electrum error
As discovered in https://github.com/comit-network/xmr-btc-swap/issues/674, users may try to refund a swap even though that is no longer possible because Alice already punished.

We should catch this case by parsing the error message returned from electrum and display something more informative to the user.

Also, we need to advance the internal state to correctly reflect what happened on the blockchain, otherwise `swap list` returns misleading information.

thenhnn:
User can try to refund even if it is no longer possible. We will show plain text error rather than ugly JSON error, see #674 for more info.

Fixes #683


thenhnn:
I'm not sure about `fn cancel` because it does error handling in proper way (as i understand), but logs it to tracing, rather than showing it to user.

delta1:
Please work on an accompanying test, you could use this one as a starting point 

https://github.com/comit-network/xmr-btc-swap/blob/master/swap/tests/alice_manually_punishes_after_bob_dead.rs

swap/src/cli/cancel_and_refund.rs
```diff
@@ -105,7 +105,23 @@ pub async fn refund(
...
+        Ok(_) => (),
+        Err(refund_error) => {
+            if let Some(refund_error) = refund_error.downcast_ref::<BtcPunishedWhileRefundError>() {
+                let state = BobState::BtcPunished {
+                    tx_lock_id: state6.tx_lock_id(),
+                };
+                db.insert_latest_state(swap_id, state.clone().into())
+                    .await?;
+
+                println!("{}", refund_error);
```
binarybaron, line 119: Please always use tracing for printing logs and add a message for some context.

thenhnn:
sorry, can you unassign me? i'm a bit busy and i'm not sure if i can make this.

binarybaron:
Closing this in favour of https://github.com/comit-network/xmr-btc-swap/pull/1668

PR #1572, opened 2024-03-09
Verify hashes of monero cli on download

### LINKED ISSUE #673: monero wallet-rpc: check binaries sha256sum
After downloading the binaries (in case we don't see them in the filesystem already)

https://github.com/comit-network/xmr-btc-swap/blob/42125049413ea2d1fbf389ad9ff0642ea02d6aba/swap/src/monero/wallet_rpc.rs#L84

there are no checks for integrity being performed, but I believe it'd be best if we did - sure, downloading over HTTPS would indeed guarantee that we're not getting anything that's not coming from the servers we trust (as long as we trust the root CAs), but ideally we should make sure they are also what we expect them to be: the exact contents are reproducibly built by the community (e.g., see https://github.com/monero-project/gitian.sigs/blob/master/v0.17.2.0-linux/selsta/monero-linux-0.17-build.assert), not just what's said to be true by the `getmonero.org` servers.

My recommendation:

1. like we have a constant for the URL to fetch binaries for each OS & arch, have a constant for their checksums
2. pick the sums from https://www.getmonero.org/downloads/hashes.txt after validating that the signed message is valid (+ confirm looking at gitian sigs)
3. after `reqwest` downloads the contents and before the extraction, pipe those bytes through sha256sum and check

Wdyt?

thx!

ikmckenz:
Resolves #673

Hardcodes hashes and checks against them, reviewers should use pgp verification of the hashes I've included here to confirm them.

Also bumps the MacOS arm64 version to match the version string for other platforms.

### DIFF
--- a/swap/src/monero/wallet_rpc.rs
+++ b/swap/src/monero/wallet_rpc.rs
@@ -47,7 +47,7 @@ compile_error!("unsupported operating system");
 const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-x64-v0.18.1.2.tar.bz2";
 
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
-const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-armv8-v0.18.0.0.tar.bz2";
+const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-mac-armv8-v0.18.1.2.tar.bz2";
 
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 const DOWNLOAD_URL: &str = "https://downloads.getmonero.org/cli/monero-linux-x64-v0.18.1.2.tar.bz2";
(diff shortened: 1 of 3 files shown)


swap/src/monero/wallet_rpc.rs
```diff
@@ -256,6 +274,17 @@ impl WalletRpc {
                 file.write_all(&bytes).await?;
             }
 
+            let result = hasher.finalize();
+            let result_hash = HEXLOWER.encode(result.as_ref());
+            if result_hash != DOWNLOAD_HASH {
+                panic!(
+                    "SHA256 of download ({}) does not match expected ({})!",
+                    result_hash, DOWNLOAD_HASH
+                );
```
delta1, line 283: Please return an error here instead of panic

delta1 (commented):
Looks good, thanks for this. Just one change to return an error instead of panic. 

If you have time, it would be great to look into a CI job that can test this functionality across platforms. This might require an additional option to the swap cli like "setup" which would just download the rpc software and then bail.

delta1 (approved):

ikmckenz:
Thanks for the fix and merge! Been busier lately

binarybaron:
did you compute the hashes yourself or are they identical to the ones on https://www.getmonero.org/downloads/hashes.txt?

I need to update the hashes because I'm bumping the rpc version.

ikmckenz:
Taken from the monero hashes page, not computed.

PR #1596, opened 2024-03-26
fix: monero wallet refresh

delta1:
This PR changes the following behaviour in the refresh functionality of the monero wallet
- Allows for multiple retries because in some cases users have experienced an issue where the wallet rpc returns `no connection to daemon` even though the daemon is available. I'm not 100% sure why this happens but retrying often fixes the issue
- Print the current sync height after each failed attempt at syncing to see how far we've come
- The `monero-wallet-rpc` is started with the `--no-initial-sync` flag which ensures that as soon as it's started, it's ready to respond to requests
- The `monero-wallet-rpc` was upgraded to `v0.18.3.1` because this PR https://github.com/monero-project/monero/pull/8941 has improved some of the issues mentioned above

This PR is part of a larger effort to fix this issue https://github.com/comit-network/xmr-btc-swap/issues/1432

### DIFF
--- a/swap/src/monero/wallet.rs
+++ b/swap/src/monero/wallet.rs
@@ -280,7 +280,7 @@ impl Wallet {
 
                     // We would not want to fail here if the height is not available
                     // as it is not critical for the operation of the wallet.
-                    // We can just log the error and continue.
+                    // We can just log a warning and continue.
                     let height = match self.inner.lock().await.get_height().await {
                         Ok(height) => height.to_string(),
                         Err(_) => {
--- a/swap/src/monero/wallet.rs
+++ b/swap/src/monero/wallet.rs
@@ -132,7 +132,10 @@ impl Wallet {
         // it saves its state correctly
         let _ = self.inner.lock().await.close_wallet().await?;
 
-        let _ = self.inner.lock().await
+        let _ = self
+            .inner
+            .lock()
+            .await
             .generate_from_keys(
                 file_name,
                 temp_wallet_address.to_string(),
@@ -167,7 +170,12 @@ impl Wallet {
             }
         }
 
-        let _ = self.inner.lock().await.open_wallet(self.name.clone()).await?;
+        let _ = self
+            .inner
+            .lock()
+            .await
+            .open_wallet(self.name.clone())
+            .await?;
 
         Ok(())
     }
--- a/swap/src/monero/wallet.rs
+++ b/swap/src/monero/wallet.rs
@@ -130,9 +130,9 @@ impl Wallet {
 
         // Close the default wallet before generating the other wallet to ensure that
         // it saves its state correctly
-        let _ = wallet.close_wallet().await?;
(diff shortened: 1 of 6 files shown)


swap/src/monero/wallet.rs
```diff
@@ -261,8 +262,70 @@ impl Wallet {
         self.main_address
     }
 
-    pub async fn refresh(&self) -> Result<Refreshed> {
-        Ok(self.inner.lock().await.refresh().await?)
+    pub async fn refresh(&self, max_attempts: usize) -> Result<Refreshed> {
+        const GET_HEIGHT_INTERVAL: Duration = Duration::from_secs(5);
+        const RETRY_INTERVAL: Duration = Duration::from_secs(2);
+
+        let inner = self.inner.lock().await;
```
ikmckenz, line 269: @binarybaron doing a bit of debugging on this, it appears this line never returns, at least for `given_alice_and_bob_manually_refund_after_funds_locked_both_refund`

ikmckenz: Stack trace looks like 
```
swap::monero::wallet::Wallet::refresh::{{closure}} wallet.rs:269
swap::monero::wallet::Wallet::create_from::{{closure}} wallet.rs:148
swap::protocol::alice::state::State3::refund_xmr::{{closure}} state.rs:527
swap::asb::recovery::refund::refund::{{closure}} refund.rs:82
alice_and_bob_refund_using_cancel_and_refund_command::given_alice_and_bob_manually_refund_after_funds_locked_both_refund::{{closure}}::{{closure}}::{{closure}} alice_and_bob_refund_using_cancel_and_refund_command.rs:76
alice_and_bob_refund_using_cancel_and_refund_command::harness::setup_test::{{closure}} mod.rs:140
alice_and_bob_refund_using_cancel_and_refund_command::given_alice_and_bob_manually_refund_after_funds_locked_both_refund::{{closure}} alice_and_bob_refund_using_cancel_and_refund_command.rs:82
```

Inside `refund_xmr`, immediately before the `create_from` call there is a call to `watch_for_transfer` which itself calls `wait_for_confirmations` which also locks the client mutex. From the little I know about these locks they should unlock when it goes out of scope, but perhaps there is something going on with the borrowing that keeps it in scope? Not sure.

binarybaron: Thanks a lot. @delta1 do you know what exact issue could be? I haven't been able to figure it out.

binarybaron: I don't even understand why the HTTP Clients needs to be a Mutex. That seems so unnecessary and complicates everything...

delta1:
> The tests are passing. 

No it seems the same tests are timing out

binarybaron:
> > The tests are passing.
> 
> No it seems the same tests are timing out

Now they are passing :)

binarybaron:
@delta1 I'm going to merge this. Tests are passing and the code is relatively simple. Please ping me if you want to revert something.

PR #1640, opened 2024-05-24
Lower DEFAULT_BITCOIN_CONFIRMATION_TARGET to 1 to ensure timely confirmation of Bitcoin transactions

binarybaron:
I think we should lower this to 1 for the moment until we figure out a better way to go about this.

If Bitcoin transactions are not confirmed in time that represents a huge problem especially for Bob. He might be punished even though he published all his transactions in time.


binarybaron:
@delta1 I'm gonna merge this. If you have an objection, we can revert.

Issue #1644, opened 2024-05-25
ASB should give CLI a time for how long the offer is valid


binarybaron:
The offer sent from the ASB to the CLI is only valid for a default duration of a few minutes. The ASB should transmit that duration alongside with the quote.

delta1:
@ikmckenz do you feel like picking this up?

ikmckenz:
Does something like #1662 align with the goal?

Issue #1268, opened 2023-01-08
Add version to p2p identify 


delta1:
Knowing what version a peer is running will help with compatibility issues/debugging 

see #1184 and #1267

delta1:
breaking changes should be bumping the `protocolVersion` too, which means we also need some sort of integration test against old versions when releasing a new version

binarybaron:
Fyi I already made the agentVersion available with https://github.com/comit-network/xmr-btc-swap/commit/9e96ef64d3d5c6444dcf8cda92eba232186614e4

delta1:
Thanks! Yes saw that, and just made this issue to track having some logic in the cli/asb regarding that version

PR #1662, opened 2024-05-31
Attach offer duration to quote

ikmckenz:
Eventually I like the idea of this being a "max offer duration", which could be configurable, and then advanced users could have their own variable duration based around things like volatility. But for now I think this accomplishes the goals of #1644 .

### DIFF
--- a/swap/src/asb/event_loop.rs
+++ b/swap/src/asb/event_loop.rs
@@ -18,6 +18,7 @@ use std::collections::HashMap;
 use std::convert::{Infallible, TryInto};
 use std::fmt::Debug;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::mpsc;
 use uuid::Uuid;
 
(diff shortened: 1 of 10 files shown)


binarybaron:
Yes, that's exactly what I imagined! Making the duration variable can be added at a later point.

But unfortunately this is not backwards compatible is it? If it's not possible without introducing a seperate new libp2p protocol I'd suggest we postpone mergin this this until we need to introduce another (more important) breaking network protocol change. What do do you think?

delta1:
It can probably be made optional and remain backwards compatible?

ikmckenz:
Perhaps #1268 should be implemented prior to merging

ikmckenz:
Made the valid_duration (the BidQuote part the CLI receives) Optional. Not sure if the the integration tests cover this kind of change that may or may not be backwards compatible. Perhaps additional integration test can be made where one party uses the current code, and the other party pulls the most recent released code from GitHub and runs that? Not sure.

Speaking of tests, locally the unit test `cli::command::tests::test_cli_arguments` fails for me, both on this branch but also on the master branch.

binarybaron:
Have you tested manually if this works?

ikmckenz:
With this repository no longer being the canonical repo, should I rebase this change on UnstoppableSwap/core?

binarybaron:
> With this repository no longer being the canonical repo, should I rebase this change on UnstoppableSwap/core?

Yes, that'd be awesome

PR #1668, opened 2024-06-04
feat (Cli): Display reason for failed cancel-refund operation to the user

### LINKED ISSUE #683: [1.0 XMR] Provide better feedback when cancel/refund transaction publish fails by parsing electrum error

patrini32:
We now display the reason for a failed cancel-refund operation to the user.

Fixes #683


swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,26 +60,35 @@ pub async fn cancel(
...
+        }
         Err(err) => {
             if let Ok(error_code) = parse_rpc_error_code(&err) {
                 tracing::debug!(%error_code, "parse rpc error");
-                if error_code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain) {
-                    tracing::info!("Cancel transaction has already been confirmed on chain");
-                } else if error_code == i64::from(RpcErrorCode::RpcVerifyError) {
-                    tracing::info!("General error trying to submit cancel transaction");
+                if error_code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain)
+                    || error_code == i64::from(RpcErrorCode::RpcVerifyError)
```
binarybaron, line 76: Does `RpcErrorCode::RpcVerifyError` really always mean that the transaction has already been confirmed? What if the cancel timelock hasn't expired yet and we try to publish the cancel transaction?

We cannot move into the `BtcCancelled` state if that is the case.

swap/tests/alice_manually_punishes_after_bob_dead_and_bob_cancels.rs
```diff
@@ -0,0 +1,83 @@
+pub mod harness;
```
binarybaron, line 1: This is a reasonable test but is it required for this PR? 

Honest question. Are you testing any behaviour here that was changed by your changes?

swap/src/cli/cancel_and_refund.rs
patrini32: Oh, i think you are right.

binarybaron (changes requested):
Thank you for the PR. I left a few comments.

Also: We need to implement this logic for the refund transaction aswell.

swap/src/cli/cancel_and_refund.rs
patrini32: I added a timelock check, if I understand this part of the code correctly it should work fine now.

binarybaron:
Please add comments everywhere in your changes explaining why exactly the logic you chose is correct. This is quite critical code. We have to avoid at all cost transitioning into an incorrect state or anything along thos elines.

patrini32:
> Please add comments everywhere in your changes explaining why exactly the logic you chose is correct. This is quite critical code. We have to avoid at all cost transitioning into an incorrect state or anything along those lines.

I described my assumptions and what code does, but I'm not sure if that's enough.

binarybaron:
This whole thing with checking the response codes is too unreliable. I know it was specified in the issue but I think we should pivot to using a function like this:
https://github.com/comit-network/xmr-btc-swap/blob/49a7f7eed6fbfa0c4b5b09c431e13dd4dc30ecbd/swap/src/protocol/bob/state.rs#L492-L511

We can use it to check directly if the cancel/punish transaction is present on chain (Check when we receive an error, when trying to publish cancel/refund tx). Then we can be sure that we are in the correct state.

patrini32:
> This whole thing with checking the response codes is too unreliable. I know it was specified in the issue but I think we should pivot to using a function like this:
> 
> https://github.com/comit-network/xmr-btc-swap/blob/49a7f7eed6fbfa0c4b5b09c431e13dd4dc30ecbd/swap/src/protocol/bob/state.rs#L492-L511
> 
> We can use it to check directly if the cancel/punish transaction is present on chain (Check when we receive an error, when trying to publish cancel/refund tx). Then we can be sure that we are in the correct state.

I implemented this.

swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,9 +57,31 @@ pub async fn cancel(
...
+            Ok((txid, state))
+        }
         Err(err) => {
+            if state6
+                .check_for_tx_cancel(bitcoin_wallet.as_ref())
+                .await
+                .is_ok()
+            {
+                // Alice already cancelled, so we are out-of-sync with Alice.
+                let txid = state6
```
binarybaron, line 74: No need to manually construct the transaction here again. check_for_tx_cancel returns the transaction

swap/src/protocol/bob/state.rs
```diff
@@ -663,20 +680,44 @@ impl State6 {
 
         Ok(tx)
     }
-
-    pub async fn submit_tx_cancel(
+    // Check if tx_punish already exists.
```
binarybaron, line 683: Redundant comment

binarybaron:
> > This whole thing with checking the response codes is too unreliable. I know it was specified in the issue but I think we should pivot to using a function like this:
> > https://github.com/comit-network/xmr-btc-swap/blob/49a7f7eed6fbfa0c4b5b09c431e13dd4dc30ecbd/swap/src/protocol/bob/state.rs#L492-L511
> > 
> > We can use it to check directly if the cancel/punish transaction is present on chain (Check when we receive an error, when trying to publish cancel/refund tx). Then we can be sure that we are in the correct state.
> 
> I implemented this.

thank you!

swap/src/cli/cancel_and_refund.rs
patrini32: Sorry, forgot about that part of code.

swap/src/protocol/bob/state.rs
```diff
@@ -347,13 +347,15 @@ impl State2 {
                 punish_timelock: self.punish_timelock,
                 refund_address: self.refund_address,
                 redeem_address: self.redeem_address,
+                punish_address: self.punish_address,
```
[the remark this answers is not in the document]
patrini32: https://github.com/comit-network/xmr-btc-swap/blob/master/swap/src/network/swap_setup/alice.rs#L69

Does it mean that punish_address equals redeem_address?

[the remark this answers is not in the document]
patrini32: > We want to avoid changes to the state machine types because this is a breaking change. It's not backwards compatible with swaps that were started before this change.
> 
> Is there a way to do this differently?

I think that's the easiest way to do this. Can you notify users about the breaking change in release notes?
Or we can read the punish_address from db in the refund function.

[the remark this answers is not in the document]
patrini32: Well, actually, I can implement check_for_tx_punish on state3, then load it from db in fn refund.

binarybaron (changes requested):

swap/src/cli/cancel_and_refund.rs
```diff
@@ -20,16 +19,19 @@ pub async fn cancel_and_refund(
...
     Ok(state)
 }
 
 pub async fn cancel(
     swap_id: Uuid,
     bitcoin_wallet: Arc<Wallet>,
     db: Arc<dyn Database + Send + Sync>,
-) -> Result<(Txid, Subscription, BobState)> {
+) -> Result<(Txid, BobState)> {
+    // Subscription was introduced in commit f39e1de964f7e2f81f9624043834a31154b68a55, but isn't used anywhere. We can't get transaction if Alice already cancelled the swap.
```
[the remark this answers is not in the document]
patrini32: In case tx_cancel already exists, the function can't return a subscription from submit_tx_cancel because Alice has already broadcast tx_cancel. I grepped the code and the subscription is not used anywhere.

swap/src/protocol/bob/state.rs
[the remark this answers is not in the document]
patrini32: I can only change State2 and then load it from the db in the refund function to minimize the changes to the db.

swap/src/cli/cancel_and_refund.rs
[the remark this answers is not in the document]
binarybaron: Ok I understand that makes sense. But please remove the comment. It's not really useful now that this has been removed. It's better to put it in the commit message.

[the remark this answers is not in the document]
patrini32: > Ok I understand that makes sense. But please remove the comment. It's not really useful now that this has been removed. It's better to put it in the commit message.

Done.

patrini32:
@binarybaron, I rewrote comments.

swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,9 +56,22 @@ pub async fn cancel(
...
         Err(err) => {
+            if let Ok(tx) = state6.check_for_tx_cancel(bitcoin_wallet.as_ref()).await {
+                // Alice already cancelled swap, so Bob sets his state to BtcCancelled.
+                let state = BobState::BtcCancelled(state6);
+                db.insert_latest_state(swap_id, state.clone().into())
+                    .await?;
+                tracing::info!("Cancel transaction has already been confirmed on chain. The swap has therefore already been cancelled by Alice.");
+                return Ok((tx.txid(), state));
+            }
             if let Ok(error_code) = parse_rpc_error_code(&err) {
```
binarybaron, line 75: We have only covered this cases here: Cancel timelock has expired but cancel transaction has already been published by Alice

We still need to cover the case where we try to publish the cancel transaction but can't because the timelock has not expired yet. We have to print a fitting log message in that case.

swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,9 +56,22 @@ pub async fn cancel(
...
                 tracing::debug!(%error_code, "parse rpc error");
```
binarybaron, line 76: Choose a more fitting log message or remove entirely

swap/src/cli/cancel_and_refund.rs
```diff
@@ -105,11 +112,32 @@ pub async fn refund(
...
+            let state = BobState::BtcRefunded(state6);
+            db.insert_latest_state(swap_id, state.clone().into())
+                .await?;
 
-    Ok(state)
+            Ok(state)
+        }
+        Err(error) => {
+            if state6
+                .check_for_tx_punish(bitcoin_wallet.as_ref())
```
binarybaron, line 126: I think here we should check for the timelock instead of for the punish transaction.

The punish timelock might have expired which means we cannot refund anymore and have (from Bobs perspective) already been punished. But Alice might not have published the punish transaction yet which means we won't find it on the Blockchain. This'll also revert the breaking database change.

binarybaron (changes requested):
Left some comments.
As a general rule: Since this is the code responsible for the *manual* cancel/refund logic, we should always try to publish the transactions first, before doing anything else. All the logic with checking for the existence of transactions / checking timelocks is only to display fitting information to the user. I think this is already the case but I wanted to clarify in case it wasn't clear.

swap/src/cli/cancel_and_refund.rs
patrini32: I added check and message for this case.

patrini32: Should we still throw an unappealing error if the cancel time has not yet expired or return original state? This will break tests.

patrini32:
@binarybaron, is there anything else keeping this PR from merge?

binarybaron:
I'll have to give it another review and manually test it.

swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,9 +56,27 @@ pub async fn cancel(
...
+                return Ok((tx.txid(), state));
+            }
+            if let ExpiredTimelocks::None { .. } =
+                state6.expired_timelock(bitcoin_wallet.as_ref()).await?
+            {
+                bail!("Cancel timelock hasn't expired yet");
+            }
             if let Ok(error_code) = parse_rpc_error_code(&err) {
                 tracing::debug!(%error_code, "parse rpc error");
                 if error_code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain) {
```
binarybaron, line 82: The case below should never happen but if it does we should transition to BtcCancelled like above. If possible avoid duplicating the code above here but combine the two somehow

swap/src/cli/cancel_and_refund.rs
```diff
@@ -55,9 +56,27 @@ pub async fn cancel(
...
+                tracing::info!("Cancel transaction has already been confirmed on chain");
+                return Ok((tx.txid(), state));
+            }
+            if let ExpiredTimelocks::None { .. } =
+                state6.expired_timelock(bitcoin_wallet.as_ref()).await?
+            {
+                bail!("Cancel timelock hasn't expired yet");
+            }
             if let Ok(error_code) = parse_rpc_error_code(&err) {
                 tracing::debug!(%error_code, "parse rpc error");
```
binarybaron, line 81: Change this here to a more fitting log message here that adds some context.

swap/src/cli/cancel_and_refund.rs
```diff
@@ -68,15 +87,8 @@ pub async fn cancel(
             }
```
binarybaron, line 87: Add an else case here if the error code is not RpcVerifyError nor RpcVerifyAlreadyInChain and add a log message that gives some context

binarybaron (commented):
some comments

swap/src/cli/cancel_and_refund.rs
patrini32: As far as I know, or operation isn't supported with if let.

patrini32:
> @patrini32 I'm going to award you 0.7 XMR for this PR. Can you please give me a Monero address to sent the bounty to. Thank you for your work!

Sent you a message from @patrini:matrix.org on matrix. Thanks.

PR #1669, opened 2024-06-05
feat (Bob): Buffer transfer proof to database when we are running a different swap

binarybaron:


binarybaron:
@delta1 @einliterflasche I haven't been able to get the test to work yet but the rest of the code is ready to be reviewed :)

swap/src/database/sqlite.rs
```diff
@@ -303,6 +303,56 @@ impl Database for SqliteDatabase {
...
+        swap_id: Uuid,
+        proof: TransferProof,
+    ) -> Result<()> {
+        let mut conn = self.pool.acquire().await?;
+        let swap_id = swap_id.to_string();
+        let proof = serde_json::to_string(&proof)?;
+
+        sqlx::query!(
+            r#"
+            insert into buffered_transfer_proofs (
```
Einliterflasche, line 317: Nitpick: I feel like sql keywords should be capitalized

binarybaron: Fair point

swap/tests/happy_path_bob_runs_different_swap_while_alice_sends_transfer_proof.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        // TODO: Fix this
+        let result = select! {
+            alice_state = alice_state => {
+                drop(bob_handle); // Explicitly drop the handle to cancel bob_state2
+                ()
+            },
+            // This should ideally never be reached.
+            _ = &mut bob_handle => {
+                ()
```
Einliterflasche, line 57: if this is unreachable we should put unreachable!() or panic!() there so make sure the test fails

binarybaron: The issue is that this is sometimes reached even after Alice completed. Which shouldn't be possible (I think?). This whole test needs to be reworked.

binarybaron:
@delta1 All tests are passing. I've manually tested it. [concurrent_bobs_before_xmr_lock_proof_sent.rs](https://github.com/comit-network/xmr-btc-swap/pull/1669/commits/048cf024ee03497b9e6345d67a6bc36db9f0fea0) tests this specific edge case that is now fixed by this PR.

If you give it one final review I'll merge this.

Einliterflasche (commented):
lgtm! `let else` could be used to reduce indentation a bit, I think.

binarybaron:
> lgtm! `let else` could be used to reduce indentation a bit, I think. 


thank you. where exactly?

Einliterflasche:
> > lgtm! `let else` could be used to reduce indentation a bit, I think.
> 
> thank you. where exactly?

in the cli event loop both match statements are over `Result`s and have an `Err` arm which just logs and continues. but nvm, I just realized that the error object is needed and we can't access that when using `let else` . LGTM!

Issue #1696, opened 2024-07-01
Allow `--change-address` to be omitted to refund Bitcoin back to internal Bitcoin wallet instead of external wallet


binarybaron:
When Bob starts a swap he has to provide a "change address" which is set as the Bitcoin refund address.

Bob should be able to ommit the `--change-address` and then the Bitcoin refund address should be set to an address of the internal Bitcoin wallet.

Einliterflasche:
This may cause confusion. Perhaps we should allow a different, more explicit flag like `--refund-to-internal-wallet`

binarybaron:
> This may cause confusion. Perhaps we should allow a different, more explicit flag like `--refund-to-internal-wallet` 

You're right that's better. Good idea

PR #1709, opened 2024-07-09
allow --change-address to be omitted and default to internal wallet address

Einliterflasche:
Closes https://github.com/comit-network/xmr-btc-swap/issues/1696

### DIFF
--- a/swap/src/api.rs
+++ b/swap/src/api.rs
@@ -268,11 +268,8 @@ impl Context {
         }
     }
 
-    pub fn bitcoint_wallet(&self) -> Option<Arc<bitcoin::Wallet>> {
-        match &self.bitcoin_wallet {
-            None => None,
-            Some(wallet) => Some(wallet.clone()),
-        }
+    pub fn bitcoin_wallet(&self) -> Option<Arc<bitcoin::Wallet>> {
+        self.bitcoin_wallet.clone()
     }
 }
 
--- a/swap/src/api.rs
+++ b/swap/src/api.rs
@@ -267,6 +267,13 @@ impl Context {
             tasks: Arc::new(PendingTaskList::default()),
         }
     }
+
+    pub fn bitcoint_wallet(&self) -> Option<Arc<bitcoin::Wallet>> {
+        match &self.bitcoin_wallet {
+            None => None,
+            Some(wallet) => Some(wallet.clone()),
+        }
+    }
 }
 
 impl fmt::Debug for Context {
--- a/swap/src/cli/command.rs
+++ b/swap/src/cli/command.rs
@@ -314,6 +314,7 @@ enum CliCommand {
         #[structopt(
             long = "change-address",
             help = "The bitcoin address where any form of change or excess funds should be sent to",
+            required_unless = "refund-to-internal-wallet",
             parse(try_from_str = bitcoin_address::parse)
         )]
         bitcoin_change_address: Option<bitcoin::Address>,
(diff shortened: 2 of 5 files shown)


Einliterflasche:
This change isn't as straight forward as one would think it would be because the the `structopt` proc macro doesn't seem to work as intended.

swap/src/api.rs
```diff
@@ -267,6 +267,13 @@ impl Context {
             tasks: Arc::new(PendingTaskList::default()),
         }
     }
+
+    pub fn bitcoint_wallet(&self) -> Option<Arc<bitcoin::Wallet>> {
```
binarybaron, line 271: Typo

binarybaron (changes requested):

Einliterflasche:
The flag is now added, but the help dialogue for `swap buy-xmr` now shows:

```rust
Error: error: The following required arguments were not provided:
    --change-address <bitcoin-change-address>
    --receive-address <monero-receive-address>
    --refund-to-internal-wallet
    --seller <seller>

USAGE:
    swap buy-xmr --change-address <bitcoin-change-address> --receive-address <monero-receive-address> --refund-to-internal-wallet --seller <seller> --tor-socks5-port <tor-socks5-port>
```

even though the command works as intended (with `--refund-to-internal-wallet`, `--change-address` isn't required). Using both flags correctly gives `Error: error: The argument '--change-address <bitcoin-change-address>' cannot be used with '--refund-to-internal-wallet'`

Einliterflasche:
lmao did you just merge into my branch?

Einliterflasche:
w8 i still have to test 💀

Einliterflasche:
I added a logging message and tested the flag. Do you think we need a unit test? I don't really think so.

delta1:
Does this require a flag? change-address could be optional, and if it's excluded then we can default to using the internal wallet (with a log message and/or confirmation)

delta1:
Yep I think what I've suggested is a better UX and should just be communicated to the user, and documented in the changelog/docs. It's actually always felt painful to have to supply a change address.

Einliterflasche:
Very well. It seems that clippy agrees with you ;) 

```rust
error: field `refund_to_internal_wallet` is never read
   --> swap/src/cli/command.rs:336:9
    |
314 |     BuyXmr {
    |     ------ field in this variant
...
336 |         refund_to_internal_wallet: bool,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`
```

I'll push what you suggested shortly.

delta1:
Thanks @Einliterflasche ! Hopefully @binarybaron agrees :)

binarybaron:
> Yep I think what I've suggested is a better UX and should just be communicated to the user, and documented in the changelog/docs. It's actually always felt painful to have to supply a change address.

I agree 👍

swap/src/cli/command.rs
```diff
@@ -69,6 +69,36 @@ where
...
+                None => {
+                    let internal_wallet = context
+                        .bitcoin_wallet()
+                        .expect("bitcoin wallet should exist")
+                        .new_address()
+                        .await?;
+
+                    tracing::info!(
+                        internal_wallet=%internal_wallet,
+                        "No --change-address supplied. Incase of a refund funds will be sent to internal wallet."
```
delta1, line 95:
```suggestion
                        "No --change-address supplied. Any change will be received to the internal wallet."
```

delta1:
Please update the changelog and any docs that refer to change-address

binarybaron:
LGTM!

PR #1717, opened 2024-07-16
fix(asb): Allow history command to be run while asb is running

binarybaron:


binarybaron:
@Einliterflasche You can review this when you have time

delta1 (approved):
LGTM

I think it's worth adding a changelog entry so it will be listed with the next release

Einliterflasche:
I don't know how you feel about this, but I feel we could use something like `enum AccessMod { ReadOnly, ReadWrite }` instead of a `bool`, just to make things clearer. I could go ahead and implement it if you agree in principal but find it boring :)

binarybaron:
> I don't know how you feel about this, but I feel we could use something like `enum AccessMode { ReadOnly, ReadWrite }` instead of a `bool`, just to make things clearer. I could go ahead and implement it if you agree in principal but find it boring :)

Sounds good!

binarybaron:
Feel free to open a new PR based on my branch

binarybaron:
superseded by https://github.com/comit-network/xmr-btc-swap/pull/1722

PR #1722, opened 2024-07-22
Allow history command to be executed while asb is running by opening database in read-only mode

Einliterflasche:


binarybaron:
LGTM

binarybaron (approved):

binarybaron:
Can you add a changelog entry as well? Just a small one-liner that this is now possible (running history command while asb is running)

PR #1725, opened 2024-07-23
feat: Enhance history command with more swap details

binarybaron:
feat(asb): Enhance history command with more swap details

- Add start datetime, BTC/XMR amounts, exchange rate, and peer ID to history output
- Introduce --only-unfinished flag to filter for incomplete swaps
- Improve JSON output formatting for better parsing


binarybaron:
@Einliterflasche @delta1 This is ready for review

delta1:
Looks good, just an unhappy clippy

swap/src/asb/command.rs
```diff
@@ -33,13 +33,13 @@ where
             env_config: env_config(testnet),
             cmd: Command::Start { resume_only },
         },
-        RawCommand::History => Arguments {
+        RawCommand::History { only_unfinished } => Arguments {
```
delta1, line 36: what about `only_active` or `active_only` ? not to bikeshed the name, feel free to ignore

binarybaron: No I see your point.
"Unfinished" is a term that is used somewhere in the code base I believe. That's why I chose it.

binarybaron:
A function with a signature of `get_state(swap_id: Uuid, state_type: AliceState) -> Option<AliceState>` would be so neat.
Do you if that's possible? @delta1 @Einliterflasche

binarybaron:
> A function with a signature of `get_state(swap_id: Uuid, state_type: AliceState) -> Option<AliceState>` would be so neat. Do you know if that's possible? @delta1 @Einliterflasche

I looked a bit into it and I believe this is not possible without a macro. Something like this might work.

```rust
#[macro_export]
macro_rules! get_state {
    ($swap_id:expr, $state:path) => {{
        let states = get_states($swap_id)?;
        states.into_iter().find_map(|state| {
            if let State::Alice($state { .. }) = state {
                Some($state)
            } else {
                None
            }
        })
    }};
}
```

binarybaron:
While we're at it we should also implement this for Bob

Einliterflasche (commented):
LGTM in general. This part 
```rust
if json {
    tracing::info!(
        swap_id = %swap_id,
        swap_start_date = %swap_start_date,
        latest_state = %latest_state,
        btc_amount = %state3.btc,
        xmr_amount = %state3.xmr,
        exchange_rate = %exchange_rate,
        trading_partner_peer_id = %peer_id,
        completed = is_completed,
        "Found swap in database"
    );
} else {
    table.add_row(vec![
        swap_id.to_string(),
        swap_start_date.to_string(),
        latest_state.to_string(),
        state3.btc.to_string(),
        state3.xmr.to_string(),
        exchange_rate,
        peer_id.to_string(),
        is_completed.to_string(),
    ]);
}
```

feels a little unclean but I'm not sure there is a much better way to do it.

binarybaron:
> LGTM in general. This part 
> ```rust
> if json {
>     tracing::info!(
>         swap_id = %swap_id,
>         swap_start_date = %swap_start_date,
>         latest_state = %latest_state,
>         btc_amount = %state3.btc,
>         xmr_amount = %state3.xmr,
>         exchange_rate = %exchange_rate,
>         trading_partner_peer_id = %peer_id,
>         completed = is_completed,
>         "Found swap in database"
>     );
> } else {
>     table.add_row(vec![
>         swap_id.to_string(),
>         swap_start_date.to_string(),
>         latest_state.to_string(),
>         state3.btc.to_string(),
>         state3.xmr.to_string(),
>         exchange_rate,
>         peer_id.to_string(),
>         is_completed.to_string(),
>     ]);
> }
> ```
> 
> feels a little unclean but I'm not sure there is a much better way to do it.

I don't like it either. But I don't think there's a bit way unless we manually construct a tracing log but that seems unnecessary.

The code for the History command on the asb and cli side is very similar. We could extract some common logic but I don't know if this is really worth the effort.

Einliterflasche:
I agree that there doesn't seem to be an obvious, straight forward solution. I just wanted to voice my concerns incase anyone can think of anything. xD

We can probably go ahead if noone knows how to do it better

PR #1733, opened 2024-07-25
feat(asb + cli): Redact logs + unify tracing infrastructure

Einliterflasche:
The goal is to provide commands like `asb logs --swap-id <swap_id> --redacted` and output the log messages with addresses and id's redacted. This would enhance privacy and hopefully lead to more privacy conscious people to submit their logs too, which in turn would allow us to diagnose problems better. Examplary redacted log:

```
2024-07-22T21:26:36.413628679Z  INFO monero_address=<monero_address_0>
2024-07-22T22:03:38.716628669Z  INFO swap{id=<swap_id_0>}: Waiting for Bitcoin transaction finality txid=<transaction_id_0> seen_confirmations=1 needed_confirmations=1
2024-07-22T22:03:38.731170301Z  INFO swap{id=<swap_id_0>}: Advancing state state=btc is locked rate=0.002384 BTC
2024-07-22T22:03:39.131034802Z  INFO swap{id=<swap_id_0>}: Attempting to lock XMR monero_lock_address=<monero_address_1>
2024-07-22T22:03:39.348100081Z ERROR swap{id=<swap_id_0>}: error=JSON-RPC request failed with code -4: failed to get output distribution
2024-07-22T22:03:39.348196883Z ERROR Swap failed: JSON-RPC request failed with code -4: failed to get output distribution: JSON-RPC request failed with code -4: failed to get output distribution swap_id=<swap_id_0>
...
2024-07-23T20:37:53.044695858Z DEBUG Swap completed swap_id=<swap_id_0> final_state=safely aborted
2024-07-23T20:37:53.045118503Z DEBUG A swap stopped without sending a transfer proof: request channel closed

```


binarybaron:
First step would be to save all tracing logs to disk such that we can read them later.

Einliterflasche:
Yeah. I was under the impressions that we already do this on the cli?

[here](https://github.com/comit-network/xmr-btc-swap/blob/2eda2476ebd4953858f951cf2a401264319f868b/swap/src/cli/tracing.rs#L15)

binarybaron:
> Yeah. I was under the impressions that we already do this on the cli?
> 
> [here](https://github.com/comit-network/xmr-btc-swap/blob/2eda2476ebd4953858f951cf2a401264319f868b/swap/src/cli/tracing.rs#L15)

Yes correct

swap/src/bin/asb.rs
```diff
@@ -242,6 +248,31 @@ async fn main() -> Result<()> {
...
+
+                println!("====reading log from file `{}`", &file_name.display());
+
+                // print each line, redacted if the flag is set
+                while let Some(line) = lines.next_line().await? {
+                    let line = if redact { common::redact(&line) } else { line };
+                    println!("{}", line);
+                }
+
+                println!("====no more log files found");
```
binarybaron, line 273: The print statements probably more annoying than useful. Might be a pain when parsing?

swap/src/bin/asb.rs
```diff
@@ -242,6 +248,31 @@ async fn main() -> Result<()> {
             let config_json = serde_json::to_string_pretty(&config)?;
             println!("{}", config_json);
         }
+        Command::Logs { path, redact } => {
+            // use provided directory of default one
+            let default_dir = system_data_dir()?.join("logs");
+            let logs_dir = path.unwrap_or(default_dir);
+
+            // go through each file and print it
+            let log_files = read_dir(&logs_dir)?;
```
binarybaron, line 257: Filter here for *.log files?

swap/src/bin/asb.rs
Einliterflasche: i removed the println's and added two tracing infos at the beginning instead so they won't interfere with the output. I also added an option for printing to a file directly.

swap/src/bin/asb.rs
Einliterflasche: i did that. we probably need an option to turn of this filtering though for the sake of completeness

Einliterflasche:
Current help output:

```bash
$ ./target/debug/asb logs --help
asb-logs 0.13.3
Prints all logging messages issued in the past.

USAGE:
    asb logs [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
        --redact     Redact swap-ids, Bitcoin and Monero addresses.
    -V, --version    Prints version information

OPTIONS:
    -d <logs-dir>           Print the logs from this directory instead of the default one.
    -o <output-path>        Print the logs into this file instead of the terminal.
```

in the `redact` function we currently recreate the regex automata everytime the function is called, which it is for every single log line. As that is highly expensive i will need to rework that, probably using `Lazy<Regex>` since the patterns are constant.

binarybaron:
> Current help output:
> 
> ```shell
> $ ./target/debug/asb logs --help
> asb-logs 0.13.3
> Prints all logging messages issued in the past.
> 
> USAGE:
>     asb logs [FLAGS] [OPTIONS]
> 
> FLAGS:
>     -h, --help       Prints help information
>         --redact     Redact swap-ids, Bitcoin and Monero addresses.
>     -V, --version    Prints version information
> 
> OPTIONS:
>     -d <logs-dir>           Print the logs from this directory instead of the default one.
>     -o <output-path>        Print the logs into this file instead of the terminal.
> ```
> 
> in the `redact` function we currently recreate the regex automata everytime the function is called, which it is for every single log line. As that is highly expensive i will need to rework that, probably using `Lazy<Regex>` since the patterns are constant.

You could also run the redact function on the content of the entire file

Einliterflasche:
I thought about that but log files could get really big (gigabytes). In that case, loading the entire file into a string would eat a lot of memory. When the ASB is run on a less powerful server it might even run out of memory. Even if the memory is available, it would cause memory spikes. 

Reading the file line by line using BufReader is way more memory efficient and actually recommended, I think. 

I would have already fixed the regex issue already but my language server keeps crashing :(

Einliterflasche:
> Have you thought about how we'll implement the --swap-id flag? We could probably just regex for the id in every line

Yeah, though we can just use `str::contains` for that. I don't know if that will catch all log messages but I don't see how else to do it.

swap/src/asb/tracing.rs
```diff
@@ -1,18 +1,28 @@
...
     }
 
     let is_terminal = atty::is(atty::Stream::Stderr);
 
+    let appender = tracing_appender::rolling::never(dir.as_ref(), "swap-all.log");
+
     let builder = FmtSubscriber::builder()
         .with_env_filter(format!("asb={},swap={}", level, level))
-        .with_writer(std::io::stderr)
+        .with_writer(appender.and(std::io::stderr))
```
binarybaron, line 25: We should make sure the logs always get written to disk in JSON in the same format regardless of whether the --debug or --json flag is enabled.

This already is the case for the cli. I don't know if this also applies here.

Einliterflasche:
I took this opportunity to unify the `init` function for both asb and cli and moved it to `common::tracing_util`. The behaviour is now 100% the same. I have also removed the `--disable-timestamps` option since it was like not used at all and would require significantly worse code.

Einliterflasche:
For the RPC server I could probably make it so that a json array of log messages is returned. I think that'd make sense. What do you think @binarybaron?

binarybaron:
> For the RPC server I could probably make it so that a json array of log messages is returned. I think that'd make sense. What do you think @binarybaron?

Seems reasonable!

swap/src/common/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Print the logs from the specified logs or from the default location 
+/// to the specified path or the terminal.
+/// 
+/// If specified, filter by swap id or redact addresses.
+pub async fn print_or_write_logs(logs_dir: Option<PathBuf>, output_path: Option<PathBuf>, swap_id: Option<Uuid>, redact_addresses: bool) -> anyhow::Result<()> {
+    // use provided directory of default one
+    let default_dir = system_data_dir()?.join("logs");
+    let logs_dir = logs_dir.unwrap_or(default_dir);
+
+    tracing::info!("Reading `*.log` files from `{}`", logs_dir.display());
```
binarybaron, line 80: I think we can remove this. Most of the time we'll be reading from a single log file.

swap/src/cli/command.rs
```diff
@@ -321,6 +332,30 @@ enum CliCommand {
...
     /// Show a list of past, ongoing and completed swaps
     History,
+    /// Output all logging messages that have been issued.
+    Logs {
+        #[structopt(
+            short = "d",
+            help = "Print the logs from this directory instead of the default one."
+        )]
+        logs_dir: Option<PathBuf>,
+        #[structopt(
```
binarybaron, line 342: I think we could remove this for simplicity sake but I'll let you decide. I'd also be okay with leaving it in.

swap/src/common/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+/// helper macro for [`redact`]... eldrich sorcery
+/// the macro does in essence the following:
+/// 1. create a static regex automaton for the pattern
+/// 2. find all matching patterns using regex
+/// 3. create a placeholder for each distinct matching pattern
+/// 4. add the placeholder to the hashmap
+macro_rules! regex_find_placeholders {
+    ($pattern:expr, $create_placeholder:expr, $replacements:expr, $input:expr) => {{
+        // compile the regex pattern
+        static REGEX: once_cell::sync::Lazy<regex::Regex> = once_cell::sync::Lazy::new(|| {
```
binarybaron, line 53: I think this doesnt cache

binarybaron (commented):
left a few comments

Assuming you have tested (I think there were internal logs printed by other crates last time we tested) this this is good to go.

swap/src/cli/command.rs
binarybaron: (the write to file argument)

swap/src/common/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+        let mut counter: usize = 0;
+
+        // for every matched address check whether we already found it
+        // and if we didn't, generate a placeholder for it
+        for address in REGEX.find_iter($input) {
+            if !$replacements.contains_key(address.as_str()) {
+                $replacements.insert(address.as_str().to_owned(), $create_placeholder(counter));
+                counter += 1;
+            }
+        }
```
binarybaron, line 66: This doesn't work unless we run the function on the entire file at once. The counter wont be preserved across logs. Different addresses might get the same placeholder

Einliterflasche: I added a function `redact_with` that takes `& mut HashMap<...>` to hoist the state. currently testing

Einliterflasche:
I just realized that we should also mention the removed flag (`asb --debug`) in the changelog. I'll push a commit later

binarybaron:
> I just realized that we should also mention the removed flag (`asb --debug`) in the changelog. I'll push a commit later

--debug hasn't been removed. Do you mean --disable-timestamp?

binarybaron:
Clippy throws an error. I believe it's a false positive, seems be a known issue

binarybaron:
One of the tests is failing: https://github.com/comit-network/xmr-btc-swap/actions/runs/10488516043/job/29051158230

Einliterflasche:
The regex which checks the tracing env filter fails. strange... i'm looking into it, it may even be an error on `tracing`'s part.

Einliterflasche:
waiting for the test run to finish, but it did run past the point where it failed before. monero-harness was using an old version of `tracing-subscriber` but I could bump it without any refactoring.

binarybaron:
> All tests passed! Ready to merge?

LGTM!

Issue #1743, opened 2024-08-01
Attach monero reserve proof to quote


binarybaron:
We can allow Alice to include a monero [reserve proof](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#get_reserve_proof) in her quote which she sends to Bob. This way Bob can verify if Alice does posses the funds she offers for sale before initiating a swap with her.

This will provide some mitigation against potential spammers. Bob will check the validity of the `reserve proof` using [check_reserve_proof](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#check_reserve_proof) before initating the swap setup. Alice cannot offer to sell more Monero than she can prove she owns.

As long as his monero daemon is fully synced Bob can check the validity of a reserve proof even if his `restore_height` is higher than the height of the outputs Alice is proving to own.

PR #1749, opened 2024-08-07
feat: Include proof of reserves in quote

binarybaron:


binarybaron:
Closes https://github.com/comit-network/xmr-btc-swap/issues/1743

binarybaron:
Closes https://github.com/comit-network/xmr-btc-swap/issues/1743

